PR #133, opened 2026-05-06
feat: post-idle compact layer + aggressive idle distillation

BYK:
## Summary

Follow-up to #132. When the prompt cache goes cold after idle, reduce the cold-cache write cost by distilling aggressively and using a smaller context window on resume.

- **Force-distill ALL pending messages on idle** (`force: true`), even below the normal `minMessages` threshold. The cache is expiring anyway — distilling now means less raw content in the next context.
- **Allow meta-distillation on idle unconditionally** — cache is cold so row ID rewrites don't cause additional busts.
- **Post-idle compact layer**: when `onIdleResume()` fires, skip layer 0 (full-raw passthrough) and use a tighter raw budget (20% of usable instead of 40%) for layer 1. The distilled prefix covers the older history; the raw window only needs the current turn + minimal recent context.

## How it works

1. Session goes idle → `session.idle` fires → `backgroundDistill(sessionID, true)` force-distills everything + meta-distills
2. Cache TTL expires (5m default) → user returns → `onIdleResume()` clears caches + sets `postIdleCompact = true`
3. `transformInner()` sees `postIdleCompact` → skips layer 0, uses `rawBudget = usable * 0.20` instead of `0.40`
4. Result: distilled prefix (compact history) + tight raw window (current turn) = smaller total context = cheaper cold-cache write

## Cost impact

| Metric | Before (cold resume) | After (cold resume) |
|--------|---------------------|---------------------|
| Raw budget | 40% of usable (~58K) | 20% of usable (~29K) |
| Undistilled messages on resume | Up to `minMessages - 1` (4) | 0 (force-distilled on idle) |
| Meta-distill consolidation | Skipped when cache warm | Always on idle |
| Total context written | Full raw window + prefix | Tight raw window + consolidated prefix |

## Files changed

- `packages/core/src/gradient.ts` — `postIdleCompact` flag on SessionState, set by `onIdleResume()`, consumed by `transformInner()` to skip layer 0 and use 20% raw budget
- `packages/opencode/src/index.ts` — idle handler passes `force: true` to `backgroundDistill()`  
- `packages/gateway/src/idle.ts` — idle handler force-distills with `force: true`, no threshold gate
- `.lore.md` — updated knowledge entries reflecting implemented fixes

744 tests pass, 0 failures.


BYK:
Superseded by #134 which includes both post-idle compact layer and batch API integration on a rebased branch.

PR #406, opened 2026-05-19
fix: add deepseek- dash-prefix routing to UPSTREAM_ROUTES

rgutzen:
- Add { prefix: 'deepseek-', url: 'https://api.deepseek.com', protocol: 'openai' } to UPSTREAM_ROUTES in config.ts for direct DeepSeek API access
- Models like deepseek-v4-pro now correctly route to api.deepseek.com instead of falling back to OpenAI with a DeepSeek API key (401 error)
- Add upstream-routes.test.ts with 16 tests covering all route prefixes including DeepSeek dash-prefix (deepseek-v4-pro, deepseek-chat) and slash-prefix (deepseek/deepseek-r1 to Nvidia NIM)


BYK:
Thanks for the contribution @rgutzen! The DeepSeek dash-prefix route is a great catch — this was a real gap in the routing table.

I addressed the Copilot review comments and added a few extra test cases (o3-, o4-, mistralai/, google/ prefixes) while I was at it. Merged!

PR #523, opened 2026-06-03
fix(gateway): never merge unrelated sessions onto the gateway cwd

BYK:
## Problem

A user (Onur) running a **central/remote lore gateway** with Claude Code on a **separate machine** (via `lore run`) reported that **two unrelated projects' sessions were both grouped under a single "lore-config" project**.

### Root cause (confirmed via Sentry + code)

- Claude Code's **haiku side-channel requests** (title generation, quota/"new topic" probes) reach the gateway **without** `X-Lore-Project` and with a minimal system prompt that has no inferable path.
- `getProjectPath()` then falls back to the **gateway's own `process.cwd()`** = `/home/onur/lore-config`.
- Because that cwd is a **single shared path**, every path-less session collapses onto it → unrelated projects merge under one bogus `lore-config` project (`ensureProject` names it `path.split('/').pop()`).
- The gateway's cwd has **no relationship** to a client's project on a remote/central gateway.

The phantom `no such column: project_path` errors seen in Sentry were traced to a *different* user's private fork and are unrelated.

## Fix (provider-agnostic)

All logic operates on the **normalized request in the shared pipeline**, so it protects Anthropic, OpenAI Chat, and OpenAI Responses clients identically — no per-provider code.

1. **Sticky per-session binding** — a confident path (`X-Lore-Project` header or one inferred from the system prompt) binds the session and clears the provisional flag. A cwd fallback **never** overwrites a confident binding; a freshly-seeded cwd path is marked *provisional* so a later confident turn takes over.
2. **Reorder `initIfNeeded()`** to run **after** path correction, so a probe request never creates a project row for the cwd / an unattributed bucket.
3. **Remote-gateway mode** (`LORE_REMOTE_GATEWAY`; implied by `LORE_HOSTED_MODE`) — path-less requests route to a **per-session synthetic bucket** `/__lore_unattributed__/<sessionID>` instead of the gateway cwd, so **unrelated sessions are never merged**. Buckets get a `(unattributed)` provisional name + a dashboard badge.
4. **Self-heal** — when a confident path later arrives for a provisional session, rows already stored under the bucket are **re-pointed** into the real project via `mergeProjectInternal()`.
5. **`lore data consolidate`** — new maintenance command that merges leftover unattributed buckets into matched real projects by git remote.

## Tests

New unit tests cover sticky binding, remote bucketing, **two path-less sessions getting distinct (non-merged) buckets**, and self-heal re-attribution. Full suite: only the pre-existing flaky `quota.test.ts` header test fails (present on clean `main` too); everything else green. Typecheck passes for all 4 packages; npm bundle + bundle-exports smoke tests pass.

## Notes

- `.lore.md` session churn intentionally excluded to keep the diff focused.
- No DB schema change: `projectPath` isn't persisted in `session_state`, so the provisional flag is purely in-memory bootstrap.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1262,6 +1262,26 @@ export function close() {
 // Project management
 
 /**
+ * Path prefix for synthetic "unattributed" project buckets. Created when a
+ * remote/central gateway can't determine a confident project path for a
+ * request (no `X-Lore-Project` header, no inferable path in the system
+ * prompt). Each such session gets its own bucket
+ * (`/__lore_unattributed__/<sessionID>`) so unrelated sessions are never
+ * merged onto the gateway's own cwd. Buckets self-heal (when a confident path
+ * later arrives) or can be consolidated. Defined here (in core) so both the
+ * gateway and DB-layer naming agree on the canonical prefix.
+ */
+export const UNATTRIBUTED_PROJECT_PREFIX = "/__lore_unattributed__";
+
+/** True when a project path is a synthetic unattributed bucket. */
+export function isUnattributedProjectPath(path: string): boolean {
+  return (
+    path === UNATTRIBUTED_PROJECT_PREFIX ||
+    path.startsWith(`${UNATTRIBUTED_PROJECT_PREFIX}/`)
+  );
+}
+
+/**
  * Look up or create a project by filesystem path, with git-remote awareness.
  *
  * Resolution order:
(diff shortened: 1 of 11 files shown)


BYK:
## Addressed critical-review findings

A critical adversarial review surfaced findings; all are now addressed:

- **C1 (critical) — `consolidate` was a no-op (buckets never carried a git remote):** `resolveSessionProjectPath` now records the session's git remote on the bucket/cwd project row (`ensureProject(projectPath, undefined, effectiveRemote)`) when known, so both self-heal and `lore data consolidate` can match a bucket to its real project by remote.
- **M1 (medium) — self-heal failure permanently stranded data:** `reattributeProvisionalProject` now returns a boolean; the session stays `provisional` when re-attribution fails (e.g. transient SQLITE_BUSY), so a later confident turn retries instead of orphaning the bucket.
- **M2 (medium) — no end-to-end coverage of the merge risk:** added `test/remote-attribution.test.ts` driving the **full pipeline** via the HTTP harness — two unrelated path-less sessions get distinct buckets, the gateway cwd never becomes a project, plus a `lore data consolidate` merge test and an orphan dry-run test. Added core db tests for provisional bucket naming + `isUnattributedProjectPath`.
- **M3 (medium) — `consolidate --json` dry-run semantics:** made explicit with a `dryRun` field in JSON output and a `--dry-run` flag; destructive merge only runs with `--yes`.
- **L1/L3 — git_remote backfill narrowed:** now uses `effectiveRemote = result.gitRemote ?? sessionState.gitRemote` for both confident backfill and bucket recording.
- **L2 — docs:** added `consolidate` examples to `lore data` help.
- Also fixed a consolidate matching bug: target search now explicitly excludes buckets (a bucket and its real project can share a remote), instead of relying on a first-match lookup.

Typecheck passes (4/4 packages); full suite green (2154 pass); bundle + bundle-exports smoke pass.

PR #528, opened 2026-06-04
fix(ci): make embedding tests resilient to HuggingFace download flakes

BYK:
## Problem

The CI `test` job runs `bun test` **without vendoring the embedding model first** (only the binary-build jobs vendor it). So the `LocalProvider integration` tests in `packages/core/test/embedding.test.ts` download `nomic-embed-text-v1.5` from **HuggingFace Hub live on every run**. A transient HF **429**/outage fails the whole `test` job.

On `main` this is worse than a flaky PR check: `Build Nightly Binaries` and `Publish Nightly to GHCR` are gated on `test`, so a single HF 429 **blocks the nightly** entirely. (Hit this exact failure publishing the #523 nightly — cleared only on re-run.)

A failed model init also poisons a module-global probe flag, cascading into the `isAvailable()` tests, so one 429 fails ~5 tests.

## Fix (3 parts)

1. **CI** — vendor the model in the `test` job (cached, same key as the binary-build jobs) and point the run at it via a new `LORE_LOCAL_MODEL_PATH` env var. The test run no longer touches HF Hub.
2. **core** — `vendorModelInfo()` now honors `LORE_LOCAL_MODEL_PATH`: an existing dir makes the local provider load from disk with `allowRemoteModels=false`. Useful beyond CI for **air-gapped/offline installs**. Verified to load with the HF endpoint disabled (`HF_ENDPOINT=http://127.0.0.1:1` → still `len=768`).
3. **tests** — model-dependent assertions wrapped in `withLocalModel()`, which **skips (not fails)** on `LocalProviderUnavailableError` so a flaky/offline download never blocks a green run. Implemented as a body wrapper (**no extra embedding-worker spawn**) to avoid perturbing the ONNX worker's fragile NAPI teardown under Bun. Also reset the local-provider probe in the `isAvailable` block so a prior suite's failed probe can't poison it.

## Verification

- `bun run --filter '@loreai/core' typecheck` ✅
- Full core suite: **1124 pass, 0 fail** ✅
- Offline load proven: `LORE_LOCAL_MODEL_PATH=… HF_ENDPOINT=http://127.0.0.1:1 → LOCAL-LOAD-OK len=768` ✅
- With model available, integration tests run; when unavailable, they skip cleanly (matches clean-main behavior, no crash).

No production behavior change outside the opt-in env var.

Follow-up to the nightly-blocking flake observed while shipping #523.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,8 +40,30 @@ jobs:
       - name: Typecheck
         run: bun run typecheck
 
+      # The vendor staging tree (`.vendor-build/.model-cache/`) contains the
+      # nomic-embed-text-v1.5 ONNX model files. Needed by the binary build AND
+      # by the test run (so `LocalProvider integration` tests load the model
+      # from disk instead of downloading from HuggingFace Hub — avoiding
+      # transient 429/outage failures that block the nightly).
+      # Cached aggressively — only rebuilt when core deps or vendor scripts change.
+      - name: Cache vendor staging dirs
+        id: vendor-cache
+        uses: actions/cache@v5
+        with:
+          path: .vendor-build
+          key: vendor-${{ hashFiles('packages/core/package.json', 'packages/gateway/script/vendor-embeddings.ts', 'packages/gateway/script/vendor-paths.ts') }}
+
+      - name: Populate vendor staging (cache miss)
+        if: steps.vendor-cache.outputs.cache-hit != 'true'
+        run: bun run packages/gateway/script/vendor-embeddings.ts
+
       - name: Test
         run: bun test
+        env:
+          # Point the local embedding provider at the vendored model cache root
+          # (transformers.js resolves <root>/<modelId>/...). Keeps the test run
+          # off HuggingFace Hub. See packages/core/src/embedding-vendor.ts.
+          LORE_LOCAL_MODEL_PATH: ${{ github.workspace }}/.vendor-build/.model-cache
 
       # Compute nightly version once, pass as job output to downstream jobs.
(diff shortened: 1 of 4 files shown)


BYK:
## CI status note

The required `pull_request` CI run isn't starting on this PR due to a GitHub Actions scheduling quirk specific to **workflow-modifying PRs on this repo**:

- This PR edits `.github/workflows/ci.yml`. GitHub creates a `push`-event workflow-validation run for the workflow-modifying commit, which **fails instantly with 0 jobs** (startup failure — no logs, same pattern the `main` `push` runs show all day, paired with a `dynamic` success).
- That failed `push` check-suite for `ci.yml` appears to **pre-empt the `pull_request` run** for the same SHA, so the PR's required `test` check never runs and the PR stays `BLOCKED`.
- Confirmed it's specific to touching workflows: the concurrent #-quota PR (no workflow change) gets clean `pull_request` runs; every feature branch that doesn't touch `.github/workflows` does too. Mine is the only one with phantom `push` failures.

Tried (no effect): new SHAs, close/reopen, deleting the failed runs. This needs a maintainer/admin to either re-run from the UI, merge with admin override, or adjust the `push` trigger/permissions so workflow-modifying branches don't produce a failing push run.

### Local verification (the change itself is solid)
- `bun run --filter '@loreai/core' typecheck` ✅
- Full core suite: **1124 pass, 0 fail** ✅
- Offline load proven: `LORE_LOCAL_MODEL_PATH=… HF_ENDPOINT=http://127.0.0.1:1 → len=768` (no HF dependency) ✅
- Integration tests run when the model is available; skip cleanly (no crash) when not — matches clean-main behavior.

PR #569, opened 2026-06-05
fix(opencode): restore well-known provider list for dynamic UI selection

BYK:
## Problem

PR #562 replaced the static `GATEWAY_PROVIDERS` allowlist with `Object.keys(cfg.provider)` iteration. But OpenCode resolves some providers dynamically when the user selects a model in the UI — these providers are NOT in `cfg.provider` at config-hook time. Result: providers like OpenCode Zen (`opencode`) were still not redirected through the gateway despite being in `PROVIDER_ROUTES`.

## Fix

Hybrid approach: redirect both existing `cfg.provider` entries (wildcard) AND a `WELL_KNOWN_PROVIDERS` list (pre-registered for dynamic selection). The list ensures providers selected in the UI have their `baseURL` pre-set to the gateway.

```typescript
const allProviders = new Set([
  ...Object.keys(p),        // user-configured providers
  ...WELL_KNOWN_PROVIDERS,  // dynamically-selectable providers
]);
```

## Verification
- Typecheck: 4/4 pass
- Lint: 0 errors

### DIFF
--- a/packages/gateway/src/config.ts
+++ b/packages/gateway/src/config.ts
@@ -306,7 +306,7 @@ const PROVIDER_ROUTES: Record<string, ProviderRoute> = {
     url: "https://api.fireworks.ai/inference",
     protocol: "anthropic",
   },
-  "github-copilot": { url: null, protocol: "anthropic" },
+  "github-copilot": { url: null, protocol: "openai" },
   minimax: {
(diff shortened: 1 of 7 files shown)


BYK:
Superseded by a more comprehensive fix in a new PR.

PR #688, opened 2026-06-10
feat(recall): search people & entities (aliases + relationships) in recall

BYK:
## Why
The `recall` tool searched knowledge, distillations, temporal messages, and lat.md — but **not the entity registry**. Entities were only used for *query expansion* (aliases added as extra query variants), so identity questions like "who is Seylan?" or "what's our CI service?" never returned the entity itself with its aliases and relationships.

## What
Adds **`entity`** as a first-class recall source.

- New `TaggedResult` variant `{ source: "entity"; item: ScoredEntity }` (key prefix `e:`). Extends every exhaustive `switch` in `recall.ts`: `getTaggedText`, `taggedResultKey`, `SOURCE_WEIGHT`/`SOURCE_ORDER`/`SOURCE_LABELS`, `getFullContentLength`, `renderResultLine`, and `recallById`.
- `searchRecall()` adds two RRF lists (gated to `all`/`project` scopes — entities are cross-session identity info, so excluded from `session`/`knowledge`):
  - **FTS list** via `entities.search()` (matches canonical name *and* aliases).
  - **Vector list** via `embedding.vectorSearchEntities()`.
  Both use the `e:` key so lexical + semantic matches fuse.
- Rendering shows `**Name** (type): aka <aliases> — <role/description> — <relations>` (e.g. *"partner of Burak Yigit Kaya"*). `recallById('e:<id>')` returns full detail via `entities.formatForPrompt`.
- Dashboard `/ui/search` renders entity results (linked to the entity page).
- Recall tool description updated so the agent knows it can recall people/entities.

## Test plan
- `pnpm run typecheck` / `pnpm run lint` — clean
- `pnpm test` — 2525 passed / 6 skipped (94 files), incl. new `recall-entity.test.ts` (FTS + alias match, relation rendering, `recallById`, and scope-gating for `knowledge`/`session`)
- Manual: `lore recall "Seylan"` / dashboard search returns the person with aliases + "partner of …".

## Follow-up (not in this PR)
Investigating this surfaced an entity-dedup gap: first-name vs full-name duplicates (`Seylan` vs `Seylan Çinar Kaya`) aren't merged or even suggested — name-Jaccard is 1/3 (< 0.5), there's no shared alias value, and short-name embedding cosine is < 0.85. A **name-containment signal** (token-subset → strong duplicate, suggest/auto-merge for `person`) would fix it. Happy to do that next.

### DIFF
--- a/packages/core/src/recall.ts
+++ b/packages/core/src/recall.ts
@@ -47,6 +47,10 @@ type Distillation = {
 
 export type ScoredDistillation = Distillation & { rank: number };
 
+/** An entity result. `rank` mirrors the other scored sources for type parity;
+ *  RRF fuses by list position, not by this value. */
+export type ScoredEntity = entities.EntityWithAliases & { rank: number };
+
 export type RecallScope = "all" | "session" | "project" | "knowledge";
 
 export type RecallInput = {
@@ -86,7 +90,8 @@ export type TaggedResult =
     }
   | { source: "distillation"; item: ScoredDistillation }
   | { source: "temporal"; item: temporal.ScoredTemporalMessage }
-  | { source: "lat-section"; item: latReader.ScoredLatSection };
+  | { source: "lat-section"; item: latReader.ScoredLatSection }
+  | { source: "entity"; item: ScoredEntity };
 
 export type ScoredTaggedResult = { item: TaggedResult; score: number };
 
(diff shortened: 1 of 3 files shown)


BYK:
Addressed self-review findings:

- **B1 (blocking, fixed):** the entity *vector* path used `embedding.vectorSearchEntities` which is **not** project-scoped, so a semantic match could leak another project's project-scoped `repo`/`infra` entity. Now post-filtered with the same visibility predicate `entities.search()` uses (`project_id === pid || project_id IS NULL || cross_project === 1`). Added a cross-project no-leak test.
- **I2 (fixed):** `recallById('e:<id>')` dropped relations (`formatForPrompt([ent])` only renders relations relative to a `self` entity in the array). Now appends `formatRelationsForPrompt` explicitly, and the test asserts the relation appears in the drill-down.
- **Nits:** corrected the `ScoredEntity` doc (RRF fuses by list position, not the rank value) and the stale RRF-cap comment.

typecheck/lint clean; `pnpm test` 2526 passed / 6 skipped.

PR #734, opened 2026-06-12
fix(gateway): omit unsupported Codex token cap

### LINKED ISSUE #715: Pi /compact fails with session_not_found when using openai-codex provider
## Summary

When using `@loreai/pi` with Pi configured for the `openai-codex` provider, every Pi `/compact` fails with:

```text
pi: compaction endpoint returned 404: {"error":"session_not_found","message":"No active session found for the given headers. Ensure at least one conversation turn has been routed through the gateway."}
```

This appears to happen because the Pi extension installs the Lore compaction hook globally, but normal `openai-codex` conversation turns are not routed through the Lore gateway. Later, `/compact` calls Lore's explicit compaction endpoint for a session Lore has never seen.

## Environment

Installed package versions:

```json
{
  "@loreai/gateway": "0.27.0",
  "@loreai/pi": "0.27.0"
}
```

Pi settings:

```json
{
  "defaultProvider": "openai-codex",
  "defaultModel": "gpt-5.5",
  "transport": "websocket-cached",
  "packages": [
    "/usr/lib/node_modules/@loreai/pi"
  ]
}
```

Pi session logs show normal model calls using:

```json
{
  "provider": "openai-codex",
  "api": "openai-codex-responses"
}
```

## Why this happens

`@loreai/pi` registers a fixed set of providers that can be routed through the Lore gateway. The current `OPENAI_PROVIDERS` list does not include `openai-codex`.

As a result:

1. Pi uses `openai-codex` / `openai-codex-responses` for normal turns.
2. Those requests bypass Lore because the provider is not registered/routed by `@loreai/pi`.
3. Lore never sees a normal conversation turn for the Pi session.
4. When `/compact` runs, the Pi extension still calls:

   ```http
   POST /v1/compact
   ```

   with `x-lore-session-id`.
5. The gateway has no existing active session for that ID, so `handleCompactEndpoint()` returns `404 session_not_found`.

This matches the gateway behavior in `packages/gateway/src/pipeline.ts`:

```ts
if (isNew) {
  return new Response(
    JSON.stringify({
      error: "session_not_found",
      message:
        "No active session found for the given headers. " +
        "Ensure at least one conversation turn has been routed through the gateway.",
    }),
    { status: 404, headers: { "content-type": "application/json" } },
  );
}
```

## Additional complication: Codex endpoint shape

Adding `openai-codex` to the provider list alone may not be enough.

Pi's `openai-codex-responses` provider uses ChatGPT/Codex-specific endpoints such as:

```text
https://chatgpt.com/backend-api/codex/responses
wss://chatgpt.com/backend-api/codex/responses
```

Lore currently routes/intercepts standard LLM paths such as:

```text
/v1/messages
/v1/chat/completions
/v1/responses
/api/v1/messages
/api/v1/chat/completions
/api/v1/responses
```

It does not currently handle `/backend-api/codex/responses`, and the Pi config may use WebSocket transport first via:

```json
"transport": "websocket-cached"
```

So full `openai-codex` support likely requires more than just adding the provider ID.

## Reproduction

1. Configure Pi with:

   ```json
   {
     "defaultProvider": "openai-codex",
     "defaultModel": "gpt-5.5",
     "transport": "websocket-cached",
     "packages": ["/usr/lib/node_modules/@loreai/pi"]
   }
   ```

2. Start a new Pi session.
3. Send a normal prompt.
4. Run `/compact`.
5. Observe:

   ```text
   pi: compaction endpoint returned 404: {"error":"session_not_found", ...}
   ```

6. Inspect Pi session logs and confirm the turn used:

   ```json
   "provider": "openai-codex",
   "api": "openai-codex-responses"
   ```

## Expected behavior

Either:

- Lore supports routing `openai-codex` sessions and `/compact` succeeds, or
- Lore recognizes that the active provider/session was not routed through Lore and cleanly falls back to Pi's default compaction without a noisy `session_not_found` error.

## Actual behavior

`@loreai/pi` calls Lore's `/v1/compact` endpoint even though the current session has not been routed through Lore, causing a 404 `session_not_found` error.

## Two possible ways to fix this

There seem to be two reasonable fix directions:

### 1. Full feature fix: add real `openai-codex` support

This would make Lore memory and compaction work for Pi's `openai-codex` provider.

Likely requirements:

- Add `openai-codex` to the Pi provider registration path.
- Route or intercept Codex requests using `/backend-api/codex/responses`.
- Decide how to handle `websocket` / `websocket-cached` transport, or force HTTP/SSE when using Lore.
- Preserve Codex-specific auth/headers.
- Verify that the Codex request/response/SSE event format is compatible with Lore's existing OpenAI Responses pipeline, or add translation logic.
- Add gateway route/interceptor support for the Codex-specific path if needed.

### 2. Minimum UX fix: skip Lore compaction for unsupported or unrouted sessions

This would not add Codex memory support, but it would prevent the confusing failure.

Possible behavior:

- If the active Pi provider is not one of Lore's supported gateway providers, do not call `/v1/compact`.
- If `/v1/compact` returns `session_not_found`, log a clearer message and return `undefined` so Pi's default compaction runs.
- Example message:

  ```text
  Lore compaction unavailable: current provider/session has not been routed through Lore. Falling back to Pi compaction.
  ```

This would make the limitation explicit and avoid making `/compact` look broken for unsupported providers.

## Workaround

Use a Lore-supported Pi provider instead of `openai-codex`, start a fresh session, send at least one normal turn through Lore, then run `/compact`.

aminvakil:
Fixes https://github.com/BYK/loreai/issues/715#issuecomment-4690955037

### DIFF
--- a/packages/gateway/src/llm-adapter.ts
+++ b/packages/gateway/src/llm-adapter.ts
@@ -415,7 +415,6 @@ function buildCodexWorkerRequest(
   model: { providerID: string; modelID: string },
   system: string,
   user: string,
-  maxTokens: number,
   sessionID?: string,
   temperature?: number,
 ): { url: string; headers: Record<string, string>; body: string } {
@@ -435,7 +434,6 @@ function buildCodexWorkerRequest(
       // Codex REQUIRES store:false (rejects store:true).
       store: false,
       stream: false,
-      max_output_tokens: maxTokens,
       ...(temperature != null && { temperature }),
       instructions: system,
       input: [
@@ -549,7 +547,6 @@ function buildWorkerRequest(
         model,
         system,
         user,
-        maxTokens,
         sessionID,
         temperature,
       );
--- a/packages/gateway/src/translate/openai-responses.ts
+++ b/packages/gateway/src/translate/openai-responses.ts
@@ -327,7 +327,7 @@ export function buildOpenAIResponsesUpstreamRequest(
     stream: req.stream,
   };
 
-  if (req.maxTokens) {
+  if (req.maxTokens && !req.codex) {
     body.max_output_tokens = req.maxTokens;
   }
 
--- a/packages/gateway/test/llm-adapter.test.ts
+++ b/packages/gateway/test/llm-adapter.test.ts
@@ -502,6 +502,7 @@ describe("createGatewayLLMClient.prompt", () => {
     const body = JSON.parse(init?.body as string) as Record<string, unknown>;
     expect(body.store).toBe(false);
     expect(body.model).toBe("gpt-5.1-codex-mini");
+    expect(body.max_output_tokens).toBeUndefined();
     expect(Array.isArray(body.input)).toBe(true);
   });
 
--- a/packages/gateway/test/openai-codex.test.ts
+++ b/packages/gateway/test/openai-codex.test.ts
@@ -91,6 +91,20 @@ describe("buildOpenAIResponsesUpstreamRequest (codex)", () => {
     expect(body.service_tier).toBe("priority");
   });
 
+  test("does not emit unsupported max_output_tokens for Codex", () => {
+    const req = parseOpenAICodexRequest(
+      { ...codexBody, max_output_tokens: 1234 },
+      {},
+    );
+    const result = buildOpenAIResponsesUpstreamRequest(
+      req,
+      "https://chatgpt.com/backend-api",
+    );
+    const body = result.body as Record<string, unknown>;
+    expect(req.maxTokens).toBe(1234);
+    expect(body.max_output_tokens).toBeUndefined();
+  });
+
   test("forces store:false even when the client omits store", () => {
     const { store: _omit, ...noStore } = codexBody;
     const req = parseOpenAICodexRequest(noStore, {});


aminvakil:
The issue is `openai-codex` uses https://chatgpt.com/backend-api/codex/responses which rejects `max_output_tokens`.

With this PR, when using `openai` , lore still sends `max_output_tokens` and it's valid.

BYK:
Thank you so much for catching this and for the precise diagnosis, @aminvakil — pinning it to `{"detail":"Unsupported parameter: max_output_tokens"}` on both the foreground turns and the worker (`Turn prefix summarization`) made the root cause obvious. 🙏

Your fix is correct. I've folded it into #735, which addresses the same bug but consolidates the suppression into the dedicated Codex path (`applyCodexResponsesDelta` + the worker builder) instead of an inline `!req.codex` guard, so all the Codex-vs-Responses differences live in one place. It also adds the inverse regression test (non-codex Responses must still send `max_output_tokens`).

Closing in favor of #735 — but the credit for finding and fixing this is yours. Thanks again for the great report and PR!

aminvakil:
Great, thanks! I'll test again when next release gets released.

PR #739, opened 2026-06-12
fix(cch): anchor cch signing to billing header to stop self-referential cache busts

BYK:
## Problem

Session `0cYAkvpcYYf5qonx` was busting its ~1.7MB prompt cache on **every turn**. `journalctl -u opencode` pinned the divergence to `system[2].text` at byte 40354, where a `cch=XXXXX` token kept changing (`cch=2d825` → `cch=928b9` → …):

```
cache-analytics: session=0cYAkvpcYYf5qonx turn=129 hit=5% read=35927 create=638460 prefixMatch=2.3% (40354/1744282B) divergence="system[2].text"
WARN: session ...: unsustainable conversation detected (5+ consecutive cache busts)
```

## Root cause (self-referential)

LTM knowledge entry `019ebc55` (rendered into `system[2]`) literally contains the text `` CCH_PLACEHOLDER = `cch=00000` (cch.ts:103); signBody() replaces it… ``.

The signing functions rewrote the *first* `cch=`/`cc_version=` token in the serialized body, which could be a token embedded in LTM/system/message content instead of the billing header. Every turn that token was re-stamped with a fresh hash → `system[2]` bytes changed → full prompt-cache bust (hit 100%→5%).

## Fix

`packages/gateway/src/cch.ts` — anchor all `cch`/`cc_version` rewrites to the **full billing-header sentinel** `x-anthropic-billing-header: cc_version=…; cc_entrypoint=…; cch=…`:
- `signBody`, `resignBody`, `validateSeed` anchor to the full sentinel (not a bare `cch=`/`cc_entrypoint=` fragment).
- `resignBody` rewrites `cc_version` **and** `cch` in one anchored pass; `cc_entrypoint` preserved verbatim.
- `signBody` is a strict no-op when no billing header is present (bare-placeholder fallback removed) — a content `cch=00000` is never signed.
- Correctness relies on the real header always being **system[0]** (worker prepends it via `buildBillingBlock`; Claude Code emits it first, enforced by `BILLING_HEADER_RE`'s `^` anchor), so first-match always targets the real header and no content can serialize before it.
- `BILLING_RESIGN_RE` is permissive on `cc_version` (suffix optional + case-insensitive) so a client header lacking the 3-hex suffix still re-signs instead of silently forwarding a stale cch (→ 401).

`packages/gateway/src/cache-analytics.ts` — broaden `CCH_PATTERN` to normalize backtick/quote/whitespace-terminated `cch=` tokens (previously required a trailing `;`), removing false-positive `messages[N]` divergence noise in the analytics-only comparison.

## Review follow-ups (all addressed)

Round 1 (positional → structural):
- **#1 anchor defeated by content before the header** → anchored to the full sentinel; the real header is always system[0], so content can't precede it. Regression test locks in "real header wins over a full-sentinel doc block after it".
- **#2 unanchored `cc_version=`** → rewritten in the same anchored pass; regression test with content `cc_version=…;` before the header.
- **#3 bare-placeholder fallback** → removed; tests assert no-op with no header and header-still-signed when content `cch=00000` sorts first.

Round 2 (self-review via subagent):
- Suffix-less / uppercase client header would have silently skipped re-signing (→ 401) → `BILLING_RESIGN_RE` made permissive; regression tests added for both.
- Docstrings corrected (removed stale regex reference and overstated "never regardless of order" wording).

## Tests

- Reworked bare-placeholder `signBody` tests to use a real billing header.
- Added regression tests: content `cc_entrypoint`/`cc_version`/`cch` tokens that sort before the header, two-sentinel ordering, suffix-less header, uppercase suffix/cch, no-header no-op, non-default `cc_entrypoint`, backtick/quote analytics normalization.
- Full gateway suite: **1530 passed**; typecheck clean; lint clean.

### DIFF
--- a/packages/gateway/src/cache-analytics.ts
+++ b/packages/gateway/src/cache-analytics.ts
@@ -47,9 +47,21 @@ export function decompressBody(compressed: Uint8Array): string {
  * variable-width replacements are safe (offsets stay aligned across turns).
  */
 
-/** Claude Code content-cache hash: changes every turn. */
-const CCH_PATTERN = /cch=[0-9a-fA-F]+;/g;
-const CCH_REPLACEMENT = "cch=__;";
+/**
+ * Claude Code content-cache hash: changes every turn.
+ *
+ * Matches the billing-header form (`cch=abcde;`) AND content occurrences that
+ * appear as markdown code spans or quoted text (`` `cch=abcde` ``,
+ * `"cch=abcde"`) or bare/whitespace-terminated tokens. These content tokens
+ * are NOT real upstream cache busts (Anthropic strips `cch=` before computing
+ * its cache key), but if left un-normalized they produce false-positive
+ * `messages[N]`/`system[N]` divergence in this analytics comparison.
+ *
+ * The trailing terminator (`;`, backtick, quote, whitespace, or end) is kept
+ * out of the match so it is preserved and byte offsets stay aligned.
+ */
+const CCH_PATTERN = /cch=[0-9a-fA-F]+(?=[;`"'\s\\]|$)/g;
+const CCH_REPLACEMENT = "cch=__";
 
 /**
  * Claude Code version suffix: 3 hex chars derived from
--- a/packages/gateway/src/cch.ts
+++ b/packages/gateway/src/cch.ts
@@ -102,6 +102,17 @@ const WORKER_SALT = "59cf53e54c78";
 
 const CCH_PLACEHOLDER = "cch=00000";
 
+/**
+ * Matches the `cch=00000` placeholder ONLY when it sits inside a billing
+ * header (i.e. immediately preceded by `cc_entrypoint=<value>; `). This
+ * prevents `signBody` from rewriting a `cch=00000` that happens to appear in
+ * arbitrary system / LTM / message content — which would silently mutate the
+ * cached prefix every turn and bust the entire prompt cache.
+ *
+ * Capture group 1 is the `cc_entrypoint=...; ` prefix so it can be preserved.
+ */
+const BILLING_PLACEHOLDER_RE = /(cc_entrypoint=[^;]*;\s*)cch=0{5}/;
+
 // ---------------------------------------------------------------------------
 // Signing
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


BYK (commented):
Reviewed in depth (and empirically reproduced the gaps). The fix correctly resolves the **observed** incident and the analytics noise, but it's a *positional* (first-match) fix where the bug is *structural* — and the exact content class that caused this incident (an LTM entry documenting the billing-header format) can still defeat it.

## SHOULD-FIX (real, verified)

**1. The anchor is defeated by content that precedes the billing header and itself contains `cc_entrypoint=…; cch=…`.**
`BILLING_PLACEHOLDER_RE = /(cc_entrypoint=[^;]*;\s*)cch=0{5}/` and `CCH_IN_BODY_RE` are non-global / first-match. Normally the billing header rides `system[0]` and sorts first — but if a system/LTM/message block that documents the header format serializes *before* it, the content token is matched instead. Reproduced:

```js
const BILLING_PLACEHOLDER_RE = /(cc_entrypoint=[^;]*;\s*)cch=0{5}/;
// LTM doc content serialized before the billing header:
// → matches content `cc_entrypoint=cli; cch=00000` at index 39,
//   not the billing header at index 113.
```

This is the same self-referential class as the original bug. Recommend anchoring to the full `x-anthropic-billing-header:\s*cc_version=…;\s*cc_entrypoint=…;\s*cch=` sentinel (the shape `BILLING_HEADER_RE` already encodes) rather than the bare `cc_entrypoint=…;` fragment.

**2. `cc_version=` in `resignBody` is still fully unanchored** (`CC_VERSION_RE`, non-global, `body.replace(...)`). Content containing `cc_version=2.1.35.abc;` before the header gets rewritten every turn → identical cache bust. The PR fixes the `cch=` half but leaves the adjacent `cc_version=` half with the same hazard. Anchor both in the same pass.

**3. `signBody`'s bare-placeholder fallback re-introduces the original bug** for a body with content `cch=00000` and no billing header. Production is safe today (caller is `billingBlock`-guarded), but the fallback contradicts the new contract and has no test asserting it doesn't fire when a header is present.

## Test gap
All regression tests place the billing header **first**, so the first-match hazard (#1/#2) passes the suite while remaining broken. Add a "content-with-`cc_entrypoint`/`cc_version` BEFORE the billing header" regression test.

## Nit
PR description says a content `cch=` is "**never** rewritten" — given first-match semantics that's overstated; it's "never, *as long as the billing header sorts first*."

The analytics `CCH_PATTERN` change is offset-safe and correct — no concerns there.

Happy to pair on the sentinel-anchoring if useful.

PR #778, opened 2026-06-15
fix(core): stop LTM cache churn + close the cache-bust test blind spot

BYK:
## The bug (third in a series)

After the tier-gate (#771) and raw-window pin-march (#773) fixes, a large opus session **still** busted the prompt cache every turn (cache_write:read ≈ 12.5:1, "unsustainable" warning every turn). Root cause: the context-bound LTM block (system[2]) churned. Two compounding mechanisms:

1. **Budget wobble → LTM set churn.** `getLtmBudget()` derives the LTM token budget from `usable`, which shifts every turn (it's computed from the per-turn calibrated-overhead EMA). That moving budget moved the `ltm.forSession()` greedy packing boundary, so the lowest-ranked pinned entry dropped in/out **every turn** → the selected knowledge set changed every turn.
2. **Delta accumulation.** Each set change **appended a new** durable prompt-delta row (seq 0,1,2,…) at a different `insertAt`. Replaying them inserts a fresh synthetic message into the cached prefix every change → shifts later messages → cache bust.

Fueled by a feedback loop: tool failures in a debugging session caused idle distillation to auto-mint `Recurring <tool> failure` gotcha entries that entered the selected set and amplified the churn.

## Fixes
- **Quantize the LTM budget** (`getLtmBudget`): snap to `LTM_BUDGET_STEP` (8K), round to nearest, never below one step — so normal overhead wobble no longer moves the packing boundary.
- **Coalesce the durable-delta channel** (`upsertSessionPromptDelta` + `appendKnowledgePromptDelta`): keep exactly ONE row (sentinel `seq 0`), replaced in place via `ON CONFLICT`, at a **frozen** `insertAt`. The message prefix is byte-stable until the delta CONTENT genuinely changes.
- **Gate the tool-failure gotcha auto-mint** behind `knowledge.autoToolFailureGotchas` (default **false**).

## Why the suite never caught this — and the structural fix

Post-mortem: the cache-stability suite asserted only on cache **inputs** (byte-stability between two turns) and never on the cache **output**. Replay fixtures report **zero** cache usage, so the gateway's own oracles (`categorizeBust` / `recordCacheUsage` / `calibrate`) ran **blindfolded** in every test — a bust produced no observable difference. All three production bugs lived in exactly the region (`messages[idx>1]`) that `isTailDivergence()` waved through.

- **Simulated-cache oracle** (`test/helpers/simulated-cache.ts` + harness `simulateCache`): computes each turn's `cache_read`/`cache_creation` from the **actual** upstream-body prefix stability and feeds it back, so a real bust is observable/assertable.
- **New e2e** (`cache-bust-oracle.e2e.test.ts`): long, naturally-compressed session + mid-session knowledge growth; asserts the steady-state write:read ratio stays low — the cache-output guardrail that was absent.
- **Unit regression**: `getLtmBudget` stable across overhead wobble (fails pre-fix: 7 distinct budgets; passes quantized).
- **Coalescing regression**: multiple successive knowledge changes → ONE `seq 0` row (fails pre-fix: 3 rows).
- Harness gains first-class `budget` (per-project `.lore.json`) + `cacheTurns()`.

All three new regressions **verified to fail on revert** of their respective fix.

## Verification
- `pnpm test` — **3170 passed**, 6 skipped, 0 failed
- `pnpm run typecheck` — clean (all packages)
- `pnpm run lint` — exit 0 (13 pre-existing warnings, none in changed files)

### DIFF
--- a/packages/core/src/config.ts
+++ b/packages/core/src/config.ts
@@ -258,8 +258,24 @@ export const LoreConfig = z.object({
         .describe(
           "Max entities to inject into the agent system prompt. Set to 0 to disable. Default: 30.",
         ),
+      /** Auto-create `gotcha` knowledge entries from recurring tool failures
+       *  (same tool + error type across multiple sessions). Default false:
+       *  these entries are usually environmental noise (agent/tool flakiness,
+       *  not codebase facts), and minting them mid-session churns the selected
+       *  LTM set, which can bust the prompt cache. Opt in to surface recurring
+       *  tool-failure patterns as knowledge. */
+      autoToolFailureGotchas: z
+        .boolean()
+        .default(false)
+        .describe(
+          "Auto-create gotcha entries from recurring tool failures. Default false (noisy; can churn the LTM cache).",
+        ),
     })
-    .default({ enabled: true, maxEntityInject: 30 })
+    .default({
+      enabled: true,
+      maxEntityInject: 30,
+      autoToolFailureGotchas: false,
+    })
     .describe("Long-term knowledge (curator, entity injection) controls."),
   curator: z
     .object({
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -2294,6 +2294,38 @@ export function appendSessionPromptDelta(input: {
     );
 }
 
+/**
+ * Upsert a SINGLE coalesced prompt delta for a session, pinned to sentinel
+ * `seq = 0`, replacing any prior value in place.
+ *
+ * Unlike `appendSessionPromptDelta` (which appends a new row at MAX(seq)+1 each
+ * call), this keeps exactly one durable-delta row per session. The knowledge
+ * delta describes the CURRENT selected-vs-pinned state, so it must REPLACE the
+ * previous delta, not accumulate alongside it. Accumulating rows (each at a
+ * different insertAt as the conversation grew) inserts a new synthetic message
+ * into the cached prefix every time the knowledge set changes, shifting all
+ * later messages and busting the prompt cache. Coalescing into one row at a
+ * frozen insertAt keeps the message prefix byte-stable until the delta's
+ * CONTENT genuinely changes (one bust per real change, not a growing cascade).
+ */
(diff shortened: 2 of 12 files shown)


BYK:
## Addressed adversarial-review BLOCKER (B1): coalescing no longer drops earlier supersessions

The first review correctly found that coalescing the durable delta into a single replaced row could **drop an earlier turn's supersession**, because the delta was computed against the *rolling* pin baseline (`pinned.entryKeys` was advanced to the current set each turn while the system[2] bytes stayed frozen).

**Fix:** the pin's `entryKeys` is now kept **frozen** at the baseline that matches the rendered system[2] bytes — across all paths that preserve the pinned bytes (Layer-1 material-delta, Layer-1 empty-removal, Layer-4 refresh, Layer-4 no-refresh) and their **persistence** (`ltmPinKeys`) so it survives restarts. The single coalesced row is therefore recomputed as the full **cumulative** delta (frozen system[2] → current selection) every turn — re-upserting the same (frozen, current) pair yields byte-identical content (idempotent, no extra bust).

**New regression** (`cache-stability.e2e.test.ts` "coalesced delta is CUMULATIVE: two entries superseded on different turns…"): pins two entries, supersedes one on turn 2, restarts, supersedes the other on turn 4, and asserts the single row supersedes **both** (counts superseded bullets — UUIDv7 8-char prefixes collide for close-in-time entries, so id-matching can't be used). **Verified to fail on revert** of the frozen-baseline fix (superseded count 1, expected 2).

`pnpm test` 3171 passed / 0 failed; typecheck + lint clean.

BYK:
Also addressed the review's MINOR follow-up: `buildKnowledgeDeltaMessage` now sorts the "Additional Changed Knowledge (truncated)" overflow entries by id before slicing, so that section is byte-stable across turns (it previously inherited volatile forSession ranking order, which could churn the delta when >3 pinned entries changed content at once). The primary rendered section was already order-stable via formatKnowledge.

PR #783, opened 2026-06-15
fix(core): stop curator from minting near-duplicate preferences

BYK:
## Problem

The LLM curator re-observes the same behavioral rule each session and re-phrases it — e.g. "document invariants in code" stated 3 ways, and a cluster of 4 "always run adversarial review before merge" entries. In one day a single project accumulated **19 `preference` entries**, most near-duplicates.

This is the most costly kind of bloat: preferences inject into the **stable `system[1]` LTM block** pinned into *every* session's prompt (~5,300 tokens of duplicated rules became permanent overhead).

**Root cause:** `ltm.create()` dedups on **title only** (exact `LOWER(title)` + fuzzy title word-overlap) — never content, never embeddings. The embedding-based `findSemanticDuplicate` existed but was wired only into pattern-echo. The post-create sweep uses cosine ≥ 0.935 (deliberately strict); paraphrased preferences cluster ~0.85–0.92, just under it. Consolidation never fired (global-count trigger, excludes cross-project, only sends the low-confidence tail).

## Fix (two parts)

**1. Block new near-duplicate preferences at create time.** `findSemanticDuplicate` gains an optional `threshold`; new `PREFERENCE_DEDUP_THRESHOLD = 0.88` (looser than global 0.935, scoped to one category so it can't false-merge distinct architecture/gotcha entries). New async pre-pass `dedupePreferenceCreates` in the curator rewrites a near-duplicate preference `create` into an `update` of the existing entry. Same-batch double-claim guarded.

**2. Merge existing per-category bloat over time.** New per-category consolidation trigger (`PER_CATEGORY_CONSOLIDATION_THRESHOLD = 12`) fires even when global count is under `maxEntries`. `consolidate()` gains a `focusCategory` mode that sends **all project-scoped** entries of that category (not just the low-confidence tail) to the LLM to merge. Cross-project (global) entries are **excluded** from the focus path so one project can't delete a globally-shared preference; global dups are handled at create time.

## Tests
- `ltm.test.ts`: a 0.90 paraphrase is missed by the global 0.935 cutoff but caught at `PREFERENCE_DEDUP_THRESHOLD`.
- `curator-preference-dedup.test.ts`: create→update rewrite; non-preference passthrough; no-dup passthrough; same-batch double-claim guard; embeddings-unavailable no-op. **Verified fail-on-revert.**

## Verification
- `pnpm test` — **3197 passed**, 6 skipped, 0 failed
- `pnpm run typecheck` — clean; `pnpm run lint` — exit 0

### DIFF
--- a/packages/core/src/curator.ts
+++ b/packages/core/src/curator.ts
@@ -15,9 +15,11 @@ import {
   CURATOR_SYSTEM,
   curatorUser,
   CONSOLIDATION_SYSTEM,
+  CONSOLIDATION_MERGE_SYSTEM,
   consolidationUser,
 } from "./prompt";
 import * as toolTrace from "./tool-trace";
+import { tagToTitle } from "./pattern-extract";
 import { detectAndFormat } from "./instruction-detect";
 import { curatorLimiter } from "./session-limiter";
 import type { LLMClient } from "./types";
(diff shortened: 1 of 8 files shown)


BYK:
## Updated: folded in the re-observation root-cause fix + addressed all review findings

The first review found a **BLOCKER** and several MINORs. This revision both fixes them and addresses the deeper root cause (the curator re-observing already-captured patterns every run, which is what *generates* the duplicates).

### Root cause (new in this revision)
The curator appends a "behavioral patterns detected" signal recomputed over ALL project history every run, with no awareness of what's already captured — so it's repeatedly re-prompted to re-create preferences/gotchas it already minted. Fixed:
- `buildActionTagContext` / `buildToolFailureContext` now exclude tags/failures already captured as preference/gotcha entries.
- `instruction-detect.detectAndFormat` drops repeated instructions that semantically match an existing preference.

### Review findings addressed
- **BLOCKER (focusCategory no-op):** `targetMax=length` → "remove at least 0" merged nothing. Fixed with a dedicated merge-oriented prompt (`CONSOLIDATION_MERGE_SYSTEM`, no eviction target, explicitly never merges opposing rules). New test asserts the merge prompt is used and merge ops applied — **verified fail-on-revert** (would have caught the original BLOCKER).
- **M2 (contradictory comment):** corrected to state project-scoped-only.
- **m1 (global mutation from a project):** create-dedup now never rewrites a project-scoped create onto a globally-shared preference (falls through as a create). Test added.
- **m2 (info loss):** never overwrites richer existing content with a terser paraphrase (refreshes confidence only). Test added.
- **m3 (same-batch second paraphrase):** documented as an accepted limitation; the per-category consolidation merges the residual later.
- **(2) 0.88 merging opposites:** the merge prompt explicitly forbids merging opposing rules; create-dedup is scoped to one category.
- **M1 (missing Part-2 tests):** added focusCategory + behavioral-signal-exclusion tests.

### Verification
- `pnpm test` — **3203 passed**, 6 skipped, 0 failed
- `pnpm run typecheck` — clean; `pnpm run lint` — exit 0

PR #785, opened 2026-06-15
fix(gateway): stop background-worker 401s on non-Anthropic providers

BYK:
PR #776 fixed cross-provider **worker** routing/auth, but three background-call paths still 401 on **MiniMax / OpenRouter** sessions — they were never on #776's path. Found via production logs (`cache-warmer: upstream error 401 ... invalid x-api-key`, `lore-curator failed (auth-rejected)`, `lore-curator failed (no-auth)`).

## #1 cache-warmer → 401 (MiniMax) — right key, **wrong host**
`buildAnthropicProfile` fell back to `https://api.anthropic.com` when `resolveUpstreamRoute(model)` was null (no MiniMax prefix) and no `upstreamBase` was passed. `resolveProfile`'s only gate was `protocol !== "anthropic"` — but MiniMax's anthropic-compat endpoint **reports** `protocol:"anthropic"`, so it leaked through and sent the session's MiniMax `sk-cp-…` key to Anthropic → `401 invalid x-api-key`, every warmup.

**Fix:** gate warming on the Anthropic **first-party host**, not the wire protocol. `resolveProfile` resolves the host the warmup would hit (session upstream URL, now threaded from `idle.ts`) and returns null unless it's `api.anthropic.com`. Compat providers don't honor `cache_control` warming anyway.

## #2 curator/distill → auth-rejected (MiniMax) — right host+key, **no user-agent**
The worker reached `api.minimax.io/anthropic` with the correct `x-api-key`, but MiniMax rejected it (*"login fail: carry the API secret key in the 'X-Api-Key' field"*) — its generic auth failure for a request lacking a recognized `user-agent`. The conversation path works only because it forwards the client UA; the worker sent none.

**Fix:** attach a `user-agent` to every anthropic worker request via `workerUserAgent(sessionID)` (replays the sniffed session UA, falls back to a Claude Code-style UA) — for all sessions, not just OAuth.

## #3 curator/pattern-echo → no-auth (OpenRouter) — **fabricated anthropic worker**
`getWorkerModel` defaulted `effectiveProvider` to `"anthropic"` when the session provider was absent, fabricating an Anthropic worker for a non-Anthropic session → lookup of the foreign credential under `"anthropic"` missed → #776 fail-closed `no-auth`.

**Fix:** with no config model AND no session provider, `getWorkerModel` returns `undefined` (skip) instead of inventing an anthropic target.

## Invariants preserved
- 🔴 Never fall back to `api.anthropic.com` for a non-Anthropic key.
- #776's `resolveAuth` fail-closed guard untouched (`auth.ts` unchanged).

## Tests (all fail-on-revert verified)
- `cache-warmer.test.ts`: resolveProfile gate (Anthropic warms; MiniMax/unknown-host skip; non-anthropic protocol skip).
- `worker-cross-provider.test.ts`: anthropic worker request carries a user-agent.
- `worker-model.test.ts`: no-provider / empty-session → undefined.

Full suite: **3210 passed**, typecheck + lint clean.

### DIFF
--- a/packages/gateway/src/cache-warmer.ts
+++ b/packages/gateway/src/cache-warmer.ts
@@ -689,6 +689,11 @@ export function buildAnthropicProfile(
   ttl: "5m" | "1h",
   upstreamBase?: string,
 ): CacheWarmingProfile {
+  const route = resolveUpstreamRoute(model);
+  // || (not ??): an empty-string upstreamBase (header-less Anthropic sessions)
+  // must coalesce to the model route, not produce a base of "".
+  const base = upstreamBase || route?.url || "https://api.anthropic.com";
+
   const entry = getModelEntrySync(model);
   const cacheReadCost =
     entry.cost?.cache_read ?? (entry.cost?.input ?? 3) * 0.1;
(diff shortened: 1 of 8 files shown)


BYK:
## Fixed adversarial-review BLOCKER + MAJOR

**BLOCKER (warming silently disabled for Claude Code):** the host-gate used \`upstreamBase ?? route?.url\`, but header-less Anthropic sessions (Claude Code CLI/Desktop send no \`x-lore-upstream-url\`) have \`lastUpstream.url === ""\`. \`??\` left \`warmupHost = ""\` → \`new URL("")\` throws → skip → **warming never fired for the flagship agent**. The positive test masked it by passing a literal \`api.anthropic.com\` that production never produces.

Fix: use \`||\` (coalesces \`""\` → model route = \`api.anthropic.com\`) in BOTH \`resolveProfile\` and \`buildAnthropicProfile\`. Added discriminating tests with the **real production inputs** (\`upstreamBase=""\` and \`undefined\`) — verified fail-on-revert.

**MAJOR (proxied Anthropic skipped):** gating purely on first-party host would skip legit Anthropic-through-a-proxy (\`LORE_UPSTREAM_ANTHROPIC\` / LiteLLM / Cloudflare AI Gateway). Now the gate accepts a session as Anthropic when **either** \`providerID === "anthropic"\` (preserves proxied Anthropic) **or** the host is \`api.anthropic.com\` (covers the header-less flagship case where providerID is undefined). \`providerID\` is threaded from \`idle.ts\`. Test added.

Full suite: **3213 passed**, typecheck + lint clean.

PR #786, opened 2026-06-16
fix(gateway): re-anchor durable knowledge-delta on compression (stop tool-pair stripping)

BYK:
## Problem
Production logs showed `removed orphaned tool_use/tool_result block(s) from message 497/499` firing **every turn** for ~16 min, correlated with `prompt-delta: upserted knowledge update ... insertAt=498`. A real tool call was being **silently deleted from history every turn** (plus noisy WARNs) — though it did NOT bust the cache (the strip was byte-stable turn-over-turn).

**Root cause:** the durable knowledge-delta's `insertAt` is a frozen **absolute** index into the gradient-transformed message array, which is **non-stationary** — as the conversation grows and the gradient compresses (raw-window eviction / layer changes), the content at each absolute index shifts. An index that was tool-pair-safe when frozen (then the tail, ~msg 499) drifts deep into history (msg 498 of 1777) and lands **between** an `assistant(tool_use)` and its `user(tool_result)`. The old replay spliced anyway and relied on `removeOrphanedToolResults` to **destructively strip** the pair every turn — its "should essentially never fire" assumption was violated.

## Fix
Tie the delta's lifetime to the event that invalidates everything else — **compression**. When the gradient compresses (a cache-busting layer change), the cache is busting anyway, so recompute the delta (position **and** content) **that same turn** instead of replaying a stale frozen index:

- `deleteSessionPromptDelta(sessionID)` (core) — clears the row.
- `shouldResetDeltaOnCompression(prevLayer, curLayer)` — pure predicate (compressed layer `>=1` whose layer differs from the previous turn).
- Pipeline tracks `lastDeltaLayer`; on a compressing turn it deletes the stale row (recompute fresh if a delta is pending) or re-anchors the existing content to a fresh tool-pair-safe near-tail index. Recompute happens **on** the compressing turn so the request stays coherent. Stable turns keep the index frozen → cache stays warm.

**Defense-in-depth:**
- `safeDeltaInsertIndex` now walks back past a **pending/in-flight** `tool_use` (no following result), not just completed pairs.
- `applySessionPromptDeltas` re-validates the persisted index on replay (no-op for safe indices → byte-stable; nudges a drifted index instead of destructive stripping).

## Tests
- `delta-compress-reanchor.test.ts` (**NEW** — written TDD by an independent agent against the spec): predicate table, `safeDeltaInsertIndex` pending/completed/parallel/safe/consecutive cases, delete + no-op, production tool-pair-survival.
- `prompt-delta-tools.test.ts`: existing "splitting a pair" test strengthened to assert the pair **survives**. Verified fail-on-revert.

## Verification
- `pnpm test` — **3231 passed**, 6 skipped, 0 failed
- `pnpm run typecheck` clean; `pnpm run lint` exit 0

## Follow-up (separate)
The cache-analytics early-divergence logs are mostly normal tail-growth; there's a distinct cache-warmer early-prefix divergence (`messages[2]`/`system[0]` busts, ~3×/day) to chase separately.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -2326,6 +2326,23 @@ export function upsertSessionPromptDelta(input: {
     .run(input.sessionID, input.projectID, input.selector, input.content);
 }
 
+/**
+ * Delete all persisted prompt-delta rows for a session.
+ *
+ * Used when the gradient compresses (a cache-busting layer change): the
+ * durable delta's `insertAt` is a frozen absolute index into the
+ * gradient-transformed message array, which is non-stationary — compression
+ * reshuffles what sits at each index, so a once-safe index can drift into a
+ * tool_use/tool_result pair. Rather than tracking/validating the frozen index,
+ * we delete the row on compression so the same turn recomputes the delta
+ * (position + content) fresh against the new array. No-op when no rows exist.
+ */
+export function deleteSessionPromptDelta(sessionID: string): void {
+  db()
+    .query("DELETE FROM session_prompt_deltas WHERE session_id = ?")
+    .run(sessionID);
+}
+
 export function listSessionPromptDeltas(
   sessionID: string,
 ): SessionPromptDelta[] {
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -90,6 +90,7 @@ export {
   loadSessionTracking,
   appendSessionPromptDelta,
   upsertSessionPromptDelta,
+  deleteSessionPromptDelta,
   listSessionPromptDeltas,
   loadHeaderSessionIndex,
   loadParentChildMap,
--- a/packages/gateway/src/pipeline.ts
+++ b/packages/gateway/src/pipeline.ts
@@ -58,6 +58,7 @@ import {
   saveSessionTracking,
   loadSessionTracking,
   upsertSessionPromptDelta,
+  deleteSessionPromptDelta,
   listSessionPromptDeltas,
   loadHeaderSessionIndex,
   isHostedMode,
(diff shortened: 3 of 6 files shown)


BYK:
## Addressed adversarial-review MINOR fixes

- **Lint warnings (7 self-introduced `noNonNullAssertion`):** the test's resilient-import pattern (defensive `| undefined` casts + `!`) is no longer needed now that `deleteSessionPromptDelta` and `shouldResetDeltaOnCompression` are exported — switched to direct imports. Lint back to the 13 pre-existing warnings (exit 0), zero in changed files.
- **Steady-layer-1 disclosure confirmed:** the review verified the actual logged production case (steady layer 1, no layer change) is fixed by the **replay-time `safeDeltaInsertIndex` nudge** (sufficient + cache-stable), while `shouldResetDeltaOnCompression` handles layer *transitions* as defense-in-depth. Commit message already states this.
- **Integration-coverage gap:** acknowledged and accepted per review — the helpers (`shouldResetDeltaOnCompression`, `safeDeltaInsertIndex`, `deleteSessionPromptDelta`) are discriminating-unit-tested, and the steady-state fix (the real prod bug) is covered by the strengthened `prompt-delta-tools.test.ts`. The full-pipeline e2e was removed because reliably establishing a layer-1 delta then forcing a layer change fought the test harness; not worth the flakiness.

Full suite: 3231 passed, typecheck + lint exit 0.

ImJustWeed:
I've gotta say your builds have been a real blessing <3

BYK:
@ImJustWeed would love to hear more specifics!

ImJustWeed:
@BYK I am currently applying all your cache fixes for opencode and will fully run a workflow, but the single proposed fix from "https://github.com/anomalyco/opencode/issues/24840" improved the cache high singficantly. I am looking forward to using loreai as well after I get the cache for anthropic working properly in opencode.

PR #792, opened 2026-06-16
fix(gateway): gate cache-warming on write-efficiency + quiet divergence log noise

BYK:
## Problem (production logs, large session)
The cache-warmer replays the **full** last-turn body (~550K tokens) and writes it all, but only the 1h-TTL stable prefix (system[0]+system[1]+tools ≈ **37K, a fixed floor**) ever survives to the next real turn — the conversation tail grew/compressed since the stored body. Result: `cacheWrite=553122 cacheRead=37030 hit=6% ~$3.50` per warmup, ~10 warmups/day → **~$35/day wasted** on one session. The 37K it "refreshes" was already cached at 1h TTL anyway → the ~516K write is pure waste.

**Why the existing ROI guard misses it:** `MIN_SESSION_HIT_RATE` measures `warmupHits/totalWarmups` = **user-return probability** (~1.0 here — the user always returns), NOT cache-**write** efficiency (6%). So it never trips for a reliably-returning session whose warms are write-inefficient.

## Fix 1 — write-efficiency gate
Record each warmup's `cacheRead/(cacheRead+cacheWrite)` in a bounded rolling window (`WRITE_EFFICIENCY_WINDOW=5`) on `WarmupState`. `shouldWarm` stops warming (both tool-call and normal paths) when the rolling average stays below `MIN_WRITE_EFFICIENCY` (0.20) over ≥3 samples — independent of the return-rate guard. Samples scrubbed on session-identity change (no cross-session bleed). Surfaced in the dashboard "not warming" reason.

## Fix 2 — quiet "early divergence at byte" log noise
~95% of these were normal tail growth (prev body ended where this turn appends new messages; hit ~100%). The multi-line snippet now only logs at INFO on an **actual bust** (`cacheHitRate < 0.5`) or genuine early prefix divergence (`prefixMatch < 5%`); high-hit tail-growth is suppressed (the logger has no debug level).

## Tests (fail-on-revert verified)
- `cache-warmer.test.ts`: `warmupWriteEfficiencyTooLow` predicate; `shouldWarm` blocks on chronically-low efficiency **even with a perfect return rate**; does not over-block healthy sessions.
- `cache-analytics.test.ts`: snippet suppressed at high hit-rate, logged at low hit-rate.

## Verification
- `pnpm test` — **3259 passed**, 6 skipped, 0 failed

## Scope note
Investigation showed "stable-prefix warming" (warm only the 37K head) would be largely redundant — that prefix already lives 1h and survives the sub-5m inter-turn gaps on its own. So this PR targets the actual waste (stop inefficient full-body warms) rather than restructuring the warmup body.

### DIFF
--- a/packages/gateway/src/cache-analytics.ts
+++ b/packages/gateway/src/cache-analytics.ts
@@ -106,6 +106,10 @@ const CACHE_CONTROL_PATTERN =
   /,?"cache_control":\{[^{}]*"type":"ephemeral"[^{}]*\}/g;
 const CACHE_CONTROL_REPLACEMENT = "";
 
+/** API cache hit-rate below which an "early divergence" is logged at INFO (a
+ *  real bust worth surfacing) rather than DEBUG (normal tail-growth noise). */
+const LOW_HIT_FOR_DIVERGENCE_LOG = 0.5;
+
 export function normalizeBodyForComparison(body: string): string {
   return body
     .replace(CCH_PATTERN, CCH_REPLACEMENT)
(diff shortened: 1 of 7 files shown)


BYK:
## Addressed adversarial-review findings

**MAJOR — first-probe lockout (fixed):** The tool-call-path efficiency gate now runs *after* computing `cyclesSpent` and is guarded by `cyclesSpent >= 1`, so the **first warmup of every break (cyclesSpent===0) is always funded** — preserving the 🔴 evidence-gate invariant. Since `writeEfficiencySamples` is lifetime/durable, blocking the first probe would have permanently locked out a session whose prefix later re-stabilized (it could never record a fresh sample). The snapshot's mirror branch (`computeWarmingSnapshot`) was given the same `warmupCount >= 1` guard for consistency.
  - New tests: first probe funded despite chronically-low efficiency; continuation (cyclesSpent>=1) blocked.

**MINOR — stale comment (fixed):** `WarmupState.writeEfficiencySamples` comment corrected from "most recent first" to "FIFO — oldest evicted from the front."

**Coverage (raised >80%):** Added `executeWarmup` recording tests (ratio push, perfect-refresh→1.0, denominator=0 skip, window cap at WRITE_EFFICIENCY_WINDOW) and a snapshot write-efficiency-reason test. cache-warmer.ts patch lines now covered; whole-file 82.65% stmts / 93.47% funcs.

**MINOR (redundant `prefixMatchPercent` term in log gate) & NIT (scrub-path test):** acknowledged. The log-gate term is harmless (subsumed); the pipeline restore scrub mirrors the existing untested `lastWarmupRefreshTokens = 0` scrub at the same site — left consistent with precedent.

**#1 KEY RISK confirmed in PR's favor:** warmer-write-efficiency is the correct, self-validating signal — a session genuinely reusing its warmed prefix records high-efficiency samples on the next (pre-TTL) warmup, lifting the rolling average and releasing the gate within ≤5 warms. Cannot persistently block a valuable-warm session.

Full suite: 3266 passed, typecheck + lint clean.

PR #794, opened 2026-06-16
fix(core): budget-aware distilled-prefix trim to stop spurious Layer 4 front-busts

BYK:
## Problem
The distilled-context prefix (\`[Memory context follows]\`) is injected at \`messages[0]/[1]\` at gradient Layer >= 1. On a **meta-distillation rewrite** turn, \`distilledPrefixCached\` re-renders the full prefix; if it exceeds a compression stage's distilled budget, \`tryFit\` returns null. Stages 1 & 2 have \`distLimit=Infinity\` so they **never trim** — every stage fails → transform falls through to **emergency Layer 4**, despite the session having ample headroom (e.g. tokens=71K ≪ usable=126K, no real overflow).

The next turn has \`lastLayer=4\`, which the sticky-layer guard excludes (\`lastLayer<=3\`) → floor releases → session fits Layer 0 → the prefix **vanishes from \`messages[0]\`** → full prompt-cache front-bust. The reverse re-render re-injects it → bust again. This **0↔4 thrash** recurs for the session's lifetime.

**Log evidence (6 days):** 24 Layer-4 turns vs only **1** genuine context-overflow error → ~96% of Layer-4 escalations were this spurious prefix-overflow path. ~2.6% of all turns front-bust, concentrated in ~20 large sessions.

## Fix
In the compression loop, after the existing \`distLimit\` trim, if the rendered prefix still exceeds the stage's distilled budget, **binary-search the largest distillation count whose rendered prefix fits** and use that. The session stays at a compressed Layer 1-3 with the prefix **present** (\`messages[0]\` byte-stable), instead of escalating to emergency Layer 4. \`selectDistillations\` always preserves meta (gen>=1) entries — only gen-0 segments are dropped.

Eliminating the spurious Layer 4 also removes the 0↔4 thrash entirely: sessions stay at Layer 1-3, where the **existing** sticky guard already prevents the drop-to-0. No sticky-guard change needed (the single genuine Layer-4 overflow in 6 days was isolated — no thrash).

Only the **re-render path** is affected; the warm-cache freeze in \`distilledPrefixCached\` keeps \`messages[0/1]\` byte-identical on steady-state turns (never trimmed here).

## Tests (fail-on-revert verified)
- oversized prefix never escalates to Layer 4 (lands 1-3) — fails on revert (gets 4)
- prefix stays PRESENT across consecutive compressed turns (no 0-drop) — fails on revert (gets 0)
- genuine compaction still RELEASES to Layer 0 — non-regression guard

## Verification
- \`pnpm test\` — **3270 passed**, 6 skipped, 0 failed (all 147 existing gradient tests pass)

## Investigation trail
Initial hypothesis (distilled prefix *grows* and busts a warm cache) was wrong — that path is already prevented by the #752 warm-freeze. A follow-up investigation found the real cause is the meta-distill re-render overflowing the per-stage distilled budget → spurious Layer 4 → 0↔4 thrash. The \`forceMinLayer=4\` hypothesis was also ruled out (max forceMinLayer is 3; zero overflow errors in the oscillation window).

### DIFF
--- a/packages/core/src/gradient.ts
+++ b/packages/core/src/gradient.ts
@@ -2348,12 +2348,54 @@ function transformInner(input: {
     // Otherwise use the cached prefix (Approach C, byte-identical for cache).
     let stagePrefix = cached.messages;
     let stagePrefixTokens = cached.tokens;
+    let stageDistillations = distillations;
     if (
       stage.distLimit !== Infinity &&
       distillations.length > stage.distLimit
     ) {
-      const trimmed = selectDistillations(distillations, stage.distLimit);
-      stagePrefix = distilledPrefix(trimmed);
+      stageDistillations = selectDistillations(distillations, stage.distLimit);
+      stagePrefix = distilledPrefix(stageDistillations);
+      stagePrefixTokens = stagePrefix.reduce(
+        (sum, m) => sum + estimateMessage(m),
+        0,
+      );
+    }
+
+    // Budget-aware prefix trim: if the rendered prefix still exceeds this
+    // stage's distilled budget, drop gen-0 distillations (selectDistillations
+    // always preserves meta/gen>=1) until it fits — instead of handing tryFit
+    // an over-budget prefix, which returns null and escalates the layer. A
+    // meta-distillation rewrite forces a full re-render (cacheValid=false) whose
+    // size can blow past every stage's budget, falling through to emergency
+    // Layer 4 even when the session has ample headroom (the distilled-prefix
+    // front-bust). Trimming keeps the session at a compressed layer (1-3) with
+    // the prefix PRESENT, so messages[0] stays byte-stable across turns.
+    // NOTE: on a STEADY warm turn (unchanged budget) the cached prefix already
+    // fits, so this block does not fire and messages[0/1] stay byte-identical.
+    // It CAN fire on the warm path when stageDistBudget contracts below the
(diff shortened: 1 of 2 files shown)


BYK:
## Addressed adversarial-review findings (verdict: MERGE WITH MINOR FIXES)

Both highest risks were resolved in the fix's favor with code evidence:
- **Binary-search validity:** `selectDistillations` token size is monotonic non-decreasing in `mid` (200K-trial brute force + case proof) → search is correct.
- **Warm-prefix self-bust:** the trim *can* fire on the warm path only when the budget contracts below the frozen prefix — exactly the case where pre-fix code escalated to Layer 4 and thrashed 0↔4. Net-positive; steady warm turns (unchanged budget) stay byte-identical.

Applied both MINOR fixes:
1. **Corrected the misleading comment** (gradient.ts ~2364): now states the trim *can* run on the warm path on budget contraction, why that's still strictly better than the L4 thrash, and the empty-prefix edge case (meta alone over budget → drop rather than escalate).
2. **Added a steady-warm byte-identical test**: asserts that on an unchanged-budget warm turn the trim does NOT fire and `messages[0]/[1]` stay byte-identical (pins the no-regression guarantee).

Full suite: **3271 passed**, typecheck + lint clean. 3 of the 4 prior tests are fail-on-revert discriminating; the new steady-warm test guards the warm path.

PR #798, opened 2026-06-16
fix(core): pin compressed sessions to Layer >= 1 — no Layer-0 re-entry (prefix front-bust)

BYK:
## Problem
Once a session compresses (gradient Layer >= 1), the distilled prefix is injected at `messages[0]/[1]`. If a later turn drops back to **Layer 0**, the prefix **vanishes** from the front → `messages[0]` changes → the entire cached prompt prefix busts; the reverse re-injects it → bust again (0↔N thrash).

The existing sticky-layer guard only pins downward-oscillation for `lastLayer` 1-3 **while calibrated**, leaving two front-bust gaps that recur in production:
1. `lastLayer === 4` (genuine emergency overflow) — excluded by the guard's `<= 3`.
2. the tier-based bust-vs-continue gate, which passes a large session through at Layer 0 (`effectiveMinLayer === 0`) when `shouldCompress()` declines on a huge-context model — dropping the prefix.

**Production evidence:** 86 genuine `1-3 → 0` re-entries across **19 distinct sessions** over 6 days (~14/day), each a full-prefix front-bust. (Distinct from the spurious Layer-4 thrash fixed in #794.)

## Fix
Add a **prefix-present floor**: once a session has compressed (`lastLayer >= 1`), pin `effectiveMinLayer` to at least 1 — keeping the distilled prefix present — **except on a genuine compaction** (the host shrank the conversation below the last-known window). The decision is a pure predicate, `prefixPresentFloorApplies`, placed before all drop-to-0 paths (sticky guard, post-idle, layer-0 fit, tier gate).

🔴 Properties:
- Pins to **>= Layer 1 only** (prefix present) — never higher, so it never traps a session at an emergency layer.
- Covers `lastLayer === 4` (which the strong guard's `<= 3` excludes).
- Releases on genuine compaction (`messages.length < lastKnownMessageCount`, when that count is known `> 0`). A fresh process (count 0) does not spuriously release — it relies on the DB-restored `lastLayer` to hold the floor until the live count is set.
- Never applies to a never-compressed session (`lastLayer 0`) — small sessions stay at 0.

## Behavior change
A session that was force-escalated (`forceMinLayer` / "prompt too long" recovery) and then has a steady conversation now stays at Layer 1 instead of returning to Layer 0. The `forceMinLayer` one-shot semantics are unchanged (the flag is still cleared after one use — proven independently via `loadForceMinLayer`); only the resulting layer differs (2→1 instead of 2→0). Three `forceMinLayer` tests updated to assert the new floor with that rationale.

## Tests
- `prefixPresentFloorApplies` unit truth-table — **discriminating in both directions** (reverting the predicate to always-false fails the "applies" cases; dropping the compaction check fails the "release" cases).
- transform-level guards: genuine compaction still releases to Layer 0; a never-compressed small session stays at Layer 0 (floor does not over-apply).

## Honest limitation
The full production economic confluence (huge-context model + tier-0→1 transition + `shouldCompress` declining) could not be reproduced as a fail-on-revert **end-to-end** test in the synthetic harness — the predicate is tested directly instead, and transform-level guards confirm it's wired in and does not over-apply.

## Verification
- `pnpm test` — **3278 passed**, 6 skipped, 0 failed

### DIFF
--- a/packages/core/src/gradient.ts
+++ b/packages/core/src/gradient.ts
@@ -739,6 +739,35 @@ export function getLastLayer(sessionID?: string): SafetyLayer {
 }
 
 /**
+ * Decide whether the "prefix-present floor" applies — i.e. whether the next
+ * transform must pin to at least Layer 1 (keeping the distilled prefix present
+ * at messages[0]/[1]) because the session has already compressed.
+ *
+ * Returns true when the session has compressed before (`lastLayer >= 1`) AND
+ * this turn is NOT a genuine compaction. A genuine compaction is detected when
+ * the last-known window size is known (> 0) and the incoming conversation has
+ * shrunk below it — that is the one case where dropping back to Layer 0 is
+ * correct (the host shrank the conversation, so the prefix is no longer needed).
+ *
+ * Notes:
+ *  - Covers `lastLayer === 4` (the strong sticky guard's `<= 3` excludes it).
+ *  - When `lastKnownMessageCount === 0` (fresh process, not yet set), this never
+ *    treats the turn as a compaction — it relies on the DB-restored `lastLayer`
+ *    to hold the floor until the live count is established.
+ *  - Pure function of its inputs — unit-testable without driving a full transform.
+ */
+export function prefixPresentFloorApplies(
+  lastLayer: number,
+  messagesLength: number,
+  lastKnownMessageCount: number,
+): boolean {
+  const hasCompressed = lastLayer >= 1;
+  const genuineCompaction =
+    lastKnownMessageCount > 0 && messagesLength < lastKnownMessageCount;
+  return hasCompressed && !genuineCompaction;
+}
+
+/**
  * Force the next transform() call for this session to use at least the given layer.
  * Called when the API returns "prompt is too long" so the next attempt
  * trims the context enough to fit within the model's context window.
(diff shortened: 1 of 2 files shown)


BYK:
## Addressed adversarial-review findings (verdict: MERGE WITH MINOR FIXES)

Both highest-risk questions were answered definitively in the fix's favor:
- **Does the floor wrongly trap a session?** No — `lastKnownMessageCount` is the previous compressed-window count, so a genuine compaction is the only legitimate drop-to-0 and it's correctly detected; a token-shrink-without-count-shrink stays at Layer 1 (one redundant compressed turn, strictly better than a front-bust).
- **Are the forceMinLayer test changes masking a regression?** No — the one-shot property is now proven *more strongly* via a direct `loadForceMinLayer(...).toBe(0)` DB read, decoupled from the resulting layer; the 2→1 assertions correctly reflect the new floor and fail on revert.

### MAJOR (doc accuracy) — fixed
The reviewer correctly found that for **calibrated 1-3** sessions the floor is *redundant* with the sticky guard (same condition), so the prior 'tier-gate passes a large 1-3 session through' framing was misleading. Reconciled both the **code comment** and **commit message** to state the genuinely-new coverage accurately:
- (1) `lastLayer === 4` (sticky excludes via `<= 3`)
- (2) compressed-but-**uncalibrated** (`lastKnownInput == 0`) — where the sticky guard is skipped and the layer-0/tier-gate passthroughs drop the prefix.

Empirically confirmed this is where real busts occur: e.g. session `1JRbQfE8FU5nKxmMG` at Layer 1 (29555 tokens) → Layer 0 (29687 tokens, **grew** — not a compaction) one turn later; the sticky guard didn't fire because the session was uncalibrated. The floor (no `calibrated` requirement) closes exactly this.

### MINOR — fixed
- Softened the test-block header comment: the up-pinning **wiring** is proven by the repurposed forceMinLayer tests (2→1); the two transform-level guards cover only the over-apply direction (release-on-compaction, never-compressed-stays-0).
- NIT (stale review premise): this PR never touched `cache-stability.e2e.test.ts` in the final commit — no e2e coverage was removed.

Full suite: **3278 passed**; typecheck + lint clean.

PR #806, opened 2026-06-18
fix(gateway): keep Claude Code 2.1.181+ emitting cch through the gateway

BYK:
## Summary

Follow-up to #805 / #801. Fixes the **runtime** side of the Claude Code 2.1.181 first-party gate.

### Problem
2.1.181 only emits the `cch` billing field when it believes it is talking to the first-party API:

```js
function qu(){ if (qe._CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL) return true; return uln(); }
function uln(){ let e = process.env.ANTHROPIC_BASE_URL; if (!e) return true; return C7e(e); }
function C7e(e){ try { return ["api.anthropic.com"].includes(new URL(e).host); } catch { return false; } }
```

`lore` launches Claude Code with `ANTHROPIC_BASE_URL` pointed at the local gateway, so `qu()` is false and a 2.1.181+ client sends **no `cch` at all**. `resignBody`'s regex requires a `cch=[0-9a-fA-F]{5};` segment, so it cannot re-sign the billing header it modifies — the prior behavior silently breaks for 2.1.181+ clients.

### Fix
Set `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL=1` in the `claude-code` agent `envVars`. The gateway is a transparent proxy to the first-party API, so forcing the first-party assumption is correct, and the client keeps emitting `cch` for the gateway to re-sign.

Only the `claude-code` agent is affected (it's the only one that produces Claude Code's `cch` header; `pi` sets `ANTHROPIC_BASE_URL` but is a different binary).

### Verification
- Empirically confirmed (during #801): the same override makes 2.1.181 emit `cch` against a localhost server.
- New test in `agents.test.ts` asserts the var is set.
- Full suite: 3351 passed / 23 skipped / 0 failed. Typecheck clean. Lint exit 0.

### DIFF
--- a/packages/gateway/src/cch.ts
+++ b/packages/gateway/src/cch.ts
@@ -34,6 +34,19 @@ import { log } from "@loreai/core";
 import * as Sentry from "@sentry/bun";
 import { xxHash64 } from "./xxhash.ts";
 
+/**
+ * Claude Code (>= 2.1.181) only emits the `cch` billing field when it believes
+ * it is talking to the first-party API — it checks that ANTHROPIC_BASE_URL's
+ * host is exactly `api.anthropic.com` (or unset), otherwise it suppresses `cch`
+ * entirely. Setting this env var to a truthy value forces the first-party
+ * assumption. The Lore gateway is a transparent proxy to the first-party API,
+ * so any process it points at the gateway (Claude Code launched via `lore run`
+ * or configured via `lore setup`, and the seed extractor's capture server)
+ * must set this so `cch` keeps flowing. See quality/CCH.md.
+ */
+export const CLAUDE_CODE_FIRST_PARTY_ENV =
+  "_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL";
+
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


BYK:
Addressed independent review findings in b19a5e08:

- **M1 (major):** extended the first-party override to the `lore setup claude-code` path (`updateClaudeCodeSettings` → `~/.claude/settings.json`), so setup-configured users get the fix too — not just `lore run`.
- **n2:** centralized the env-var name as `CLAUDE_CODE_FIRST_PARTY_ENV` in `cch.ts` (used by both `agents.ts` and `setup.ts`).
- **m1/m2/n1:** expanded the rationale comment (unconditional is safe; traceparent side-effect is benign for a transparent proxy).
- **n3:** strengthened tests (URL/cwd invariance; `settings.json` assertions).
- **M2 (follow-up):** the gateway's lack of a defensive fallback for a `cch`-less billing header is tracked in #807.

Full suite 3353 passed / 23 skipped / 0 failed; typecheck + lint clean.

Issue #822, opened 2026-06-19
A1 — Sync scope seam (author_id/scope_id) + usage counter


BYK:
> **Part of epic #821.** Depends on: none (foundation)
> _This issue body mirrors the design doc `.opencode/plans/sync-A1-scope-seam.md`._

---

# A1 — Scope seam + usage counter (foundation)

Parent: `sync-sharing-epic.md`. Depends on: — (foundation, on shipped tables).
Base: latest `main` (≥ 0.33.0, incl. #816 confidence lifecycle + #820 sweep).

## Goal
Add the two-axis ownership seam to every SHAREABLE table and replace the
COUNT(*) quota with a maintained counter — all while v1 behavior stays
byte-identical to today (`scope_id = auth.uid()`). This is pure groundwork that
makes teams (E) a predicate swap and gives Pro (D) a scalable quota.

## Scope (remote schema + sync-data metas)
- Add to shareable remote tables (knowledge, entities, entity_aliases,
  entity_relations, knowledge_entity_refs):
  - `author_id uuid not null default auth.uid()` — who wrote it.
  - `scope_id  uuid not null default auth.uid()` — who owns/bills/sees it.
  - Repoint PK to `(scope_id, id)` (join: `(scope_id, knowledge_id, entity_id)`).
  - RLS keyed on `scope_id` (v1 `scope_id = auth.uid()` — identical behavior).
  - onConflict targets become `scope_id`-based (mirror PR #782's composite fix).
- Maintained `user_table_usage(scope_id, table_name, row_count bigint,
  byte_count bigint)` PK `(scope_id, table_name)`, written ONLY by SECURITY
  DEFINER triggers (no grant to authenticated). `plan_limits.max_bytes bigint`.
  BEFORE-write guard reads the counter O(1); rejects on row OR byte overflow;
  keep the 0005 per-scope advisory lock around the read+write. Backfill existing
  rows transactionally in the migration.
- Per-row size CHECKs stay (single-row ceiling).
- 🔴 `confidence` STAYS synced (it already is — `sync-data.ts:100`). Decay
  converges; do not move it local. (Local-only reinforcement/access tracking
  stays local.)

## Invariants (🔴)
- v1 `scope_id = author_id = auth.uid()` for every row; the ONLY thing teams (E)
  change is the RLS predicate `scope_id = auth.uid()` → `is_member(scope_id)`.
- `user_table_usage` is trigger-maintained only; client can never write it.
[...]

Issue #821, opened 2026-06-19
Epic: Sync, sharing & encryption foundation


BYK:
# Epic: Sync, sharing & encryption foundation

Umbrella for extending the Basic-tier sync engine (PR #782) toward Pro backup,
end-to-end encryption, and eventually team/org shared memory. This doc is the
overview + dependency map; each piece has its own sub-plan and GitHub issue.

The detailed design exploration + round-1 review findings that produced this
split live in `pro-tier-sync.md` (kept for reference).

## Cross-cutting decisions (apply to every sub-plan)

1. 🔴 **Multi-master, not authoritative.** Supabase is a shared sync backend;
   many clients read/write it; neither local nor remote is authoritative. "In
   sync" = the *logical* dataset converges, NOT byte-identical full replicas.
2. 🔴 **Local prune is sync-invisible eviction.** TTL/size prune drops local
   *residency* only — it clears `sync_state` WITHOUT a tombstone and never
   propagates. Re-pullable cache miss, not a delete. Restore/prune must key off
   *local residency*, never original `created_at`.
3. 🔴 **Uniform append-only model.** All synced tables are append + cursor. A
   "delete" is an immutable `is_deleted` marker record, interpreted by a
   materialized "current" view — not an engine-level tombstone/UPDATE path.
   Append-only tables (temporal/distillations) need ZERO tombstones; mutable
   concepts (knowledge/entities) become versioned logs (A2). Compaction of
   superseded versions = local eviction (same as #2).
4. 🔴 **Two-axis ownership.** Every shareable row carries `author_id` (who wrote
   it) AND `scope_id` (who owns/bills/sees it). v1: `scope_id = auth.uid()`
   (identical to today). Teams later = membership table + RLS predicate swap
   `scope_id = auth.uid()` → `is_member(scope_id)`. No PK change, no re-key.
5. 🔴 **Encryption is the access control for sensitive shared data**, not
   structural exclusion. Distillations/temporal ARE team-shareable (per product
   intent); their blobs are encrypted with a per-scope key wrapped to members.
   RLS gates row visibility; encryption gates content readability.
6. 🔴 **Crypto: libsodium/HPKE envelope now, MLS later.** Per-scope symmetric
   DEK encrypts blobs; DEK wrapped per member (sealed box / HPKE). Versioned
   envelope (`scheme_id`, `key_epoch`) + pluggable key-agreement so MLS
   (RFC 9420) is an additive scheme at org-scale / when forward-secrecy /
[...]

PR #828, opened 2026-06-19
feat(sync): profiles pull-only mirror + plan-tier resolution (#824)

### LINKED ISSUE #824: B — Profiles pull-only mirror + tier resolution
> **Part of epic #821.** Depends on: none
> _This issue body mirrors the design doc `.opencode/plans/sync-B-profiles-tier.md`._

---

# B — Profiles pull-only mirror + tier resolution

Parent: `sync-sharing-epic.md`. Depends on: — (independent; small).

## Goal
Give the client its plan tier without ever risking a write to server-authoritative
account data. `profiles` RLS is `select where id = auth.uid()` → a pull returns
exactly ONE row (the user's own).

## Scope
- Add `pullOnly?: boolean` to `SyncTableMeta`. `pushOnce` SKIPS pull-only tables
  (no capture trigger, never enqueued/pushed); `pullOnce` includes them.
- `profiles` becomes a pull-only synced table (single-row): syncColumns id, tier,
  display_name, email, github_login, created_at, updated_at. Local one-row
  `profiles` mirror table.
- Tier resolution = read `tier` from the local mirror. No team_config cache.

## Invariants (🔴)
- The engine can NEVER push `profiles` (pull-only) — belt-and-suspenders with the
  existing tier lock (0004–0006) which already blocks client tier writes.
- Tier changes propagate via the normal pull: billing flips tier (service_role)
  → server `updated_at` bump → next pull re-fetches the one row → mirror updates
  → pro capture (D) toggles. No bespoke "did I become pro?" path.

## Tasks
- [ ] core: `pullOnly` flag; pushOnce/pullOnce honor it; local `profiles` mirror
  table + apply path.
- [ ] migration: `profiles` sync columns/updated_at trigger already exist (0001);
  confirm pull works under RLS.
- [ ] gateway: expose `currentTier()` reading the mirror (used by D).
- [ ] tests: unit (pull-only never pushed) + integration Tier-2 (profiles pull
  returns exactly 1 row; tier flip via service_role propagates on next pull).

## Open
- Whether to also expose display_name/email locally now (dashboard) or only tier.

BYK:
Part of the sync/sharing epic **#821**, scope **B** (no dependencies; runs in parallel with A1 #822). Closes #824.

## Why
Give the client its plan **tier** without ever risking a write to server-authoritative account data. `profiles` RLS is `select where id = auth.uid()`, so a pull returns exactly the user's one row.

## Changes
- **`SyncTableMeta.pullOnly`** flag. Register remote `profiles` as a pull-only, unversioned synced table (`id, tier, github_login, display_name, email, created_at, updated_at`).
- **Local migration v49**: single-row SQLite `profiles` mirror, populated only by the pull apply path.
- **Never pushed**: `installSyncCapture` hardcodes its capture-table list, so `profiles` structurally gets no outbox trigger; `pushOnce` also skips pull-only tables (belt-and-suspenders).
- **No spurious conflicts**: `applyRemote` forces `"apply"` for pull-only tables. They carry no remote `content_hash`, so `classifyRemoteRow`'s identical-content *skip* can never fire and a re-pulled row (a billing-driven tier flip) would otherwise be misclassified as a `conflict` on every change. A pull-only row can never diverge locally → remote always wins, no `sync_conflicts` noise.
- **`currentTier()`** reads the mirror (defaults `'free'`); explicitly distinct from `SyncTier` (basic/pro/max — *which tables* to sync). Tier changes propagate via the normal pull: billing flips `tier` (service_role) → `updated_at` bump → next pull re-fetches the one row. No bespoke "did I become pro?" path.

**No new Supabase migration**: `profiles` (0001) + `tier` (0003) + RLS/`updated_at` trigger (0001) already exist.

## Tests
- **Unit (core `sync-data.test.ts`)**: profiles registered pull-only/unversioned; local writes never captured to the outbox even when sync is enabled; `currentTier()` mirror read + `'free'` default; pulled row upserts into the mirror. Updated the "capture trigger per synced table" test to exclude pull-only.
- **Unit (gateway `sync.test.ts`)**: `pushOnce` never uploads profiles; `pullOnce` mirrors the row & resolves tier; a tier flip propagates on next pull with **`conflicts === 0`** and zero `sync_conflicts` rows.
- **Tier-2 integration (real Postgres, docker-verified)**: RLS-scoped single-row pull resolves tier; a `service_role` tier flip propagates on the next pull (and a subsequent push is a no-op). Both sync integration files pass (7 + 12).
- Full suite 3466 passed; typecheck, lint (no new warnings), build all green.

## Coordination
Cleanly partitioned from A1 (#822): A1 owns the 5 shareable tables / scope seam / quota counter; B only adds the `profiles` table + the `pullOnly` flag. Only `sync-data.ts`/`sync.ts` are co-touched, in non-overlapping regions.


BYK:
## Security + adversarial review — addressed

Ran an adversarial/pentest pass. **Security posture is sound**: server RLS + column GRANTs + `guard_profile_tier` make `tier` genuinely unwritable by the client; pull isolation holds (RLS `id = auth.uid()`); no injection surface (parameterized, registry-allowlisted columns). But it surfaced two real defects (both masked by test ordering), now fixed in `e062718b`:

1. **Outbox prune-floor DoS (high).** `seedOutbox()`/`reconcile()` had no `pullOnly` guard, so a disable→enable cycle with the mirror already populated enqueued a `profiles` outbox entry that `pushOnce` skips forever → push cursor pinned at 0 → `minCursor=0` → **outbox pruning permanently disabled for all tables**. Fixed: skip `pullOnly` in `seedOutbox`/`reconcile` (root cause) + exclude `pullOnly` from the prune cursor set (defense-in-depth).
2. **Cross-account tier residue (latent).** `currentTier()`'s `LIMIT 1` + no logout/account-switch clear meant a free user could read a prior account's `pro` tier. Fixed: `clearProfileMirror()` (row + sync_state + pull cursor) on `clearSession()` (logout/expired) and on `persistSession()` user_id change (account switch); preserved across same-user token refresh.

**Regression tests** added (all fail without the fix): seed/reconcile never enqueue a pull-only table with a local row; a local `profiles` row doesn't wedge the knowledge prune floor; `clearProfileMirror` resets row/state/cursor; logout + account-switch drop the mirror while a token refresh keeps it.

Verification: full suite 3472 passed; 19/19 real-Postgres integration; typecheck, lint, build green.

Issue #835, opened 2026-06-19
Process: graduate adversarial-review findings into standing regression batteries


BYK:
Motivated by the #828 review. The adversarial/pentest pass caught both bugs the green suite missed — but that net is a per-PR human/agent step with no compounding return. The recurring miss-class across this session's PRs (#816, #820, #828) is **skip/early-return branches** and **lifecycle/state-transition sequences**.

## Goal
Make review findings compound into permanent, generalized coverage.

## Actions
- Every adversarial-review finding MUST land a regression test in the same PR (already practiced).
- Recurring *patterns* graduate into standing, parametrized batteries (see PR #831: registry-contract + identity-lifecycle).
- Add a PR/review heuristic checklist:
  - "If this test still passes with the guard deleted, it isn't testing the guard."
  - "Set up state in the adversarial order (state already present before the operation), not the order that makes the code pass."
  - "Adding a row to a fan-out registry (SYNCED_TABLES, migrations, agent registry) requires the parametrized contract to cover it."
- Default to **two separate review agents** (security/pentest + correctness/adversarial) so verdicts aren't conflated.

## Acceptance
- Heuristics documented (AGENTS.md or a CONTRIBUTING/testing doc); referenced from the review workflow.

Context: PR #828, PR #831.

Issue #834, opened 2026-06-19
Test quality: promote load-bearing invariants from comments to runtime assert helpers


BYK:
Motivated by the #828 review. We have excellent invariant *comments* (e.g. "never wedge the prune floor", "mirror holds ≤1 row", "tool_use/tool_result stay paired") that are asserted once and never continuously checked.

## Goal
Turn the load-bearing invariants into shared assert helpers invoked in test teardown, so any path that violates them fails loudly — turning every existing test into an invariant check for ~free.

## Scope
- Sync: `assertSyncInvariants()` (outbox prunable / floor advances; no orphan `sync_state`; no `pullOnly` outbox entries; `profiles` mirror ≤1 row) wired into `afterEach` across sync test files.
- Consider the same for gradient (tool_use/tool_result pairing, prefix coverage) where cheap.
- Keep helpers fast (pure DB queries) so per-test teardown cost is negligible.

## Acceptance
- `assertSyncInvariants()` exported from a test helper and used in the sync test suites; documented.

Context: PR #828, PR #831 (the registry battery is the structural complement to this runtime layer).

Issue #833, opened 2026-06-19
Test quality: property-based (fast-check) sequence tests for the sync engine state machine


BYK:
Motivated by the #828 review. Both bugs lived in **operation sequences** (disable→enable with residual state; login→logout→switch) that single-shot example tests never exercised. Property-based testing attacks exactly this.

## Goal
Generate random sequences of sync ops and assert standing invariants after every step.

## Scope
- Ops: `{ insert, update, delete, enableSync, disableSync, pushOnce, pullOnce, switchAccount, logout }` (against the faithful mock and/or the real-Postgres harness).
- Invariants asserted after each step:
  - outbox prune floor advances / outbox stays bounded (no permanent wedge)
  - no `pullOnly` table ever has an outbox entry
  - local ⇆ remote converge after a push+pull (no lost writes)
  - `currentTier()` reflects exactly the last-authenticated user
  - push-then-pull of our own row is a no-op (no ping-pong)
- Use `fast-check`; shrink failing sequences to minimal repro.

## Acceptance
- A `sync.property.test.ts` with ≥1 model-based or sequence property; runs in CI within a sane time budget.

Context: PR #828, PR #831.

PR #840, opened 2026-06-20
chore(test): Stryker mutation testing on the sync modules (#832)

### LINKED ISSUE #832: Test quality: mutation testing (Stryker) on stateful modules (sync, gradient, ltm)
Motivated by the #828 review: two bugs passed a green suite because the tests confirmed an author-chosen happy ordering rather than constraining behavior. Mutation testing directly measures *that* — "would our tests catch a bug on this line?"

## Goal
Run a mutation tester (Stryker) on the highest-risk stateful modules and treat surviving mutants as named test gaps.

## Scope
- Allowlist: `packages/core/src/sync-data.ts`, `packages/gateway/src/sync.ts`, `packages/core/src/gradient.ts`, `packages/core/src/ltm.ts`.
- Both #828 bugs would surface as surviving mutants today (e.g. deleting `if (m.pullOnly) continue`, flipping `pruneOutbox`'s `<= 0`, weakening `currentTier`'s read).
- Run as a **nightly / on-demand** job (slow, noisy) — NOT per-PR. Set a baseline mutation score per module; ratchet, don't gate hard initially.

## Acceptance
- Stryker config scoped to the allowlist; documented `pnpm` script.
- Baseline scores recorded; a short report of top surviving mutants with follow-up test tasks.

Context: PR #828 (bugs), PR #831 (registry/identity batteries). Part of the testing-strategy hardening.

BYK:
Closes #832. Stands up mutation testing — the tool that directly answers the question that started this thread: *do our tests actually constrain behavior, or just pass?*

## Why
This session's two #828 bugs (prune-floor wedge, tier residue) and earlier #816 lifecycle edges were caught by **review, not tests** — example-based tests confirmed an author-chosen ordering. Mutation testing exposes those gaps mechanically: a **surviving mutant** is a line no test constrains.

## What's here
- **`stryker.config.mjs`** — Vitest runner, `perTest` coverage, **no hard gate** (`thresholds.break = null` — advisory, ratchet over time). Plugins declared explicitly (pnpm's non-hoisted layout breaks Stryker's auto-discovery). Scoped to the **sync engine** (`sync-data.ts`, `sync.ts`).
- **`vitest.mutation.config.ts`** — narrows the per-mutant run to the sync modules' direct tests (the full 3.5k-test suite per mutant isn't tractable).
- **`pnpm mutation`** script (+ `premutation` gateway bundle) and **`.github/workflows/mutation.yml`** — weekly + on-demand, uploads the HTML/JSON report as an artifact. `reports/` + `.stryker-tmp/` gitignored.
- **Pin `@types/node@24.13.1`** at root — the Stryker install perturbed a previously transitive-only hoist and broke `packages/pi` typecheck; the explicit pin restores deterministic shared Node types.
- **`quality/MUTATION_TESTING.md`** — run instructions, baseline, and the prioritized gap list.

## Baseline — `sync-data.ts`
**68.98%** mutation score (189 killed / **80 survived** / 5 no-cov; 274 mutants; 13m38s). Of the 80 survivors, 52 are low-value registry/SQL string mutants (mostly equivalent); **28 are real logic gaps**, e.g.:

| Line | Mutant survives | Gap |
|---|---|---|
| **`sync-data.ts:378`** | `if (minCursor <= 0)` → `< 0` / removed | **The `pruneOutbox` floor guard — the exact #828 wedge mechanism — is unpinned** |
| `:296` | `v === null \|\| v === undefined` → `&&` | `contentHash` null/undefined coalesce unpinned |
| `:624` | `classifyRemoteRow` skip conditional | identical-content pull "skip" unpinned |
| `:422` | `idColumns.length === 1` → `!== 1` | `rowIdOf` single-vs-composite branch unpinned |

Follow-up test tasks are listed in the doc. **`ltm.ts`/`gradient.ts`** are the next expansion (their coverage spans many test files → need a broader include).

## Validation
Full suite **3502 passed**; typecheck, lint (no new warnings), build all green. The Stryker run itself completed end-to-end locally (report attached in the doc).

Part of the testing-strategy hardening (siblings: #833 property tests, #834 invariant teardown, #835 process). Independent of the A2 work.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -1,10 +1,11 @@
-name: Mutation testing
+name: Mutation testing (experimental)
 
 # Stryker mutation testing on the high-risk stateful sync modules (issue #832).
-# Measures whether the suite CONSTRAINS behavior, not just passes — a surviving
-# mutant is a named test gap. Slow (~25 min) and ADVISORY ONLY (no hard gate:
-# stryker.config.mjs sets thresholds.break = null). Runs weekly + on demand; the
-# HTML/JSON report is uploaded as an artifact.
+# ADVISORY ONLY (no hard gate: stryker.config.mjs sets thresholds.break = null),
+# and currently EXPERIMENTAL: on Stryker 9.6.1 + Vitest 4.1.8 the runner reports
+# FALSE SURVIVORS (upstream stryker-mutator/stryker-js#5928; tracked in #843), so
+# the uploaded report must be hand-verified, not trusted as a verdict. Kept green
+# so it's ready the moment the runner is reliable. Runs weekly + on demand.
 on:
   schedule:
     - cron: "0 5 * * 1" # Monday 05:00 UTC
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -1,0 +1,40 @@
+name: Mutation testing
+
+# Stryker mutation testing on the high-risk stateful sync modules (issue #832).
+# Measures whether the suite CONSTRAINS behavior, not just passes — a surviving
+# mutant is a named test gap. Slow (~25 min) and ADVISORY ONLY (no hard gate:
+# stryker.config.mjs sets thresholds.break = null). Runs weekly + on demand; the
+# HTML/JSON report is uploaded as an artifact.
+on:
+  schedule:
+    - cron: "0 5 * * 1" # Monday 05:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  mutation:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+      - uses: pnpm/action-setup@v4
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "24"
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
+      # The Stryker dry run executes the suite once; some gateway tests need the
+      # bundle (mirrors the `pretest` hook, which Stryker does not invoke).
+      - name: Build gateway bundle
+        run: pnpm --filter @loreai/gateway run bundle
+      - name: Run Stryker (sync modules)
+        run: pnpm exec stryker run
+      - name: Upload mutation report
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: mutation-report
+          path: reports/mutation/
+          retention-days: 30
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -1,11 +1,10 @@
-name: Mutation testing (experimental)
+name: Mutation testing
 
 # Stryker mutation testing on the high-risk stateful sync modules (issue #832).
-# ADVISORY ONLY (no hard gate: stryker.config.mjs sets thresholds.break = null),
-# and currently EXPERIMENTAL: on Stryker 9.6.1 + Vitest 4.1.8 the runner reports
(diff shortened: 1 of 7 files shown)


BYK:
## Update: marking this EXPERIMENTAL — Stryker is unreliable on our Vitest 4.1 stack

While preparing to harden the sync-data survivor list, I found the survivor list itself can't be trusted on our stack.

**Proof:** `sync-data.ts:624` (`classifyRemoteRow` skip guard) is reported `Survived` across **every** runner mode (vitest-runner `perTest` / `coverageAnalysis: all` / `off`, **and** the command runner with real file mutation). But forcing that exact mutant by hand (`if (true) return "skip";`) **fails 4 existing tests** — the suite already kills it. In the same command-runner range, 18 *other* mutants were correctly killed, so the sandbox is running mutated code; Stryker just fails to detect this one's death.

**Root cause is a known upstream bug** — not our config: stryker-mutator/stryker-js#5928 ("No proper coverage with Vitest 4", a breaking Vitest 4.0→4.1 API change). A partial fix shipped in 9.6.1, but it's **still reproducing on 9.6.1 + Vitest 4.1.8** (our exact versions; an independent repro in the issue's comments uses the same hand-verify method). Tracked on our side in **#843**.

### What changed in this PR
- `quality/MUTATION_TESTING.md`: prominent **reliability caveat** + hand-verify workflow; removed the confident 68.98% baseline / per-line gap list (kept only as "indicative, unreliable"). Also corrected an earlier mistake — `:378` (`pruneOutbox` guard) is an **equivalent mutant** (`DELETE WHERE seq<=0` deletes nothing), not "the #828 wedge mechanism".
- `stryker.config.mjs` + `mutation.yml`: marked **EXPERIMENTAL**, link the upstream bug. Still no hard gate; CI artifact is advisory.

### Why keep it rather than drop it
The infra is correct and ready the moment upstream is fixed (or we pin Vitest ≤4.0 for a dedicated run). The methodology is sound; only the runner↔Vitest-4.1 attribution is broken.

**Not** proceeding with "killing survivors" (the list is unreliable). Moving on to #834 (`assertSyncInvariants()` teardown), which is plain Vitest and unaffected.

BYK:
## Correction: the "false survivor" finding was my misdiagnosis — reverting the experimental framing

I incorrectly marked this experimental on the claim that Stryker reports false survivors on Vitest 4. That was wrong.

**What I got wrong:** Stryker's `ConditionalExpression` mutator splits a compound `if (A && B)` into separate mutants (`A → true`, `B → true`, whole → `true`), all reported at the **same column** with the same `"true"` replacement. The L624 survivor was the **`remoteHash !== null` operand** (`if (true && localHash === remoteHash)`), which is a **genuine** gap — no test passes `remoteHash = null`, so all classify tests pass under it. My "proof" applied the *whole-condition* mutant (`if (true)`) — a **different** mutant that Stryker had correctly **killed** — and I read the failures as a false survivor.

**Re-verified properly** by hand-applying the *exact* replacements:
- `:624` `remoteHash !== null → true` → all 5 classify tests **pass** → genuine survivor (correctly reported).
- `:296` `serializeValue` `||→&&` → all 56 sync tests **pass** → genuine survivor.
- Synthetic repros confirm Stryker **kills covered mutants correctly**.

So **the 68.98% baseline is reliable** and the survivor list is genuine — real coverage gaps plus equivalent mutants (e.g. `:378` `pruneOutbox`, which is correctly equivalent). This commit reverts the experimental/false-survivor caveats, restores the baseline + gap list (now correctly characterized and individually hand-verified), and keeps a sharpened "hand-verify the **exact** mutant" guide documenting the compound-condition footgun that tripped me up.

`#843` is closed as invalid (no upstream bug). No upstream issue was filed — good thing we verified the repro first.

Issue #823, opened 2026-06-19
A2 — Immutable/append-only knowledge + entities


BYK:
> **Part of epic #821. Depends on #822 (merged).** Decided design — supersedes the original exploratory body. Mirrors `.opencode/plans/sync-A2-immutable-knowledge.md`.

---

## ⚠️ Pre-implementation corrections (review, post-#828 merge — apply before coding)

1. **Migration is v50, not v49.** #828 (B — profiles mirror) shipped local migration **v49** and bumped `db.test.ts:61` to 49; A1 (#822) added no local migration. So A2's first local migration in Sub-PR 1 is **v50**, and the `db.test.ts` assertion goes **49→50** (not 48→49). Everything else in Sub-PR 1 stands.
2. **Sub-PR 3 needs a new remote migration for `logical_id`.** Remote `knowledge` (0002 + A1's 0007) has `is_deleted`/`revision` but **not** `logical_id`/`version`. Syncing the `(logical_id, version)` content log requires a Supabase `0008+` migration (after A1's `0007`). Resolve "is local `version` the synced `revision`?" in the Sub-PR 3 design, not at impl time.
3. **Preserve #828's pull-only sync invariants when touching `sync-data.ts`.** `seedOutbox`/`reconcile`/`pushOnce`-prune all skip `pullOnly` tables, and `clearProfileMirror()` runs on logout/account-switch. The merged **registry-contract battery** (`sync-registry-contract.test.ts`, #831) will fail CI if a newly-synced table (e.g. if `knowledge_meta` becomes synced) violates the pull-only/capture invariants — register it properly.
4. **Decay-driven deletes need a merge-aware rule (Sub-PR 3, non-blocking for 1–2).** `pruneDead` (confidence ≤ 0.2 → `remove()` → `is_deleted` death certificate) propagates a delete to all replicas, but `last_reinforced_at` is **local** — a device that decayed an entry from non-use would death-certificate it even if another member uses it. Use the convergent-merge rule (max/recency), not bare LWW, for decay-originated deletes.

---

# A2 — Immutable / append-only knowledge (Option A)

Parent epic: #821. Issue: #823. Depends on: A1 (#822, merged — scope seam + physical usage counter).
Mirror this doc into issue #823. `.opencode/plans/*` is gitignored — the issue is the source of truth.

## Decision (locked)

**Option A — single append-only `knowledge` table is the source of truth.** "Current" is a
[...]

PR #853, opened 2026-06-20
feat(core): flip update()/remove() onto append-only versioning (A2 sub-PR 2b-2b)

BYK:
**Part of epic #821 · A2 (#823), sub-PR 2b-2b — THE FLIP.** Builds on 2b-2a (#850). No migration (v51 landed in 2b-2a).

This is the behavior-change PR: knowledge **content is now immutable per version**. `update()`/`remove()` are rewired onto `appendVersion()`.

## Core flip
- **`update(id, {content})`** appends a new immutable version (confidence/metadata copied forward); **mutable-only** updates (`confidence`/`updated_by`/`sensitivity`) still mutate the current version in place. Re-embeds the new current version.
- **`remove(id)`** appends an `is_deleted` **death-certificate** version (no physical DELETE). The death cert *is* the tombstone + resurrection guard — `knowledge_current` excludes it, the FTS triggers drop its posting, `isTombstoned()` detects it. Since nothing is physically deleted, **FK `ON DELETE CASCADE` is dormant**, so `remove()` cleans entity refs / wiki refs / transfers **explicitly** by `logical_id`.
- **`isTombstoned()`** checks the death-cert version OR a legacy `knowledge_tombstones` row (pre-flip deletes). `remove()` no longer writes tombstones (the table is now read-only legacy).

## Identity resolution (the subtle part)
- **`logicalIdOf(id)`** — canonical resolver (base table): resolves a current OR **superseded** version id (or a logical_id) to the stable `logical_id`. `update`/`remove`/`syncRefs`/`resolveRef` and the curator route through it, so a `op.id` that `update()` just superseded stays correct.
- `appendVersion`'s current-row lookup moved **inside** the txn (Seer #850 TOCTOU NIT).
- **dedup** (`createWithStatus`) returns the stable `logical_id`, not the soon-superseded version id. The curator's dedup-merge path + update/delete handlers + ref-sync loop key on `logical_id` — fixes a real **dedup-merge undercount** (`result.updated` went to 0) the `curator-changed-entries` test caught. `_importEntries` passes the resolved current id (**Seer #848**).

## Gateway / data
- `handleShowKnowledge` / `deleteKnowledge` / `reassignKnowledge` resolve via `logicalIdOf`→`getByLogical` and present the **stable `logical_id`** as the external id (a cached UI reference keeps resolving across appends).
- `resolveId("knowledge")` matches the `logical_id` of **current** entries — the base-table prefix is now ambiguous (UUIDv7 versions share the timestamp prefix).
- `reassignKnowledge` moves **all** versions (`WHERE logical_id`) so a multi-version entry isn't split across projects (reviewer NIT).
- Bulk-op "entries affected" counts use `knowledge_current` (logical entries, not physical version rows).

## Tests
New **`ltm-flip.test.ts` (6)**: content-append vs mutable-in-place; death-cert + explicit ref cleanup (FK CASCADE dormant); `remove()` idempotency; multi-version chain with ref survival. Updated **9 existing suites** from in-place to append semantics (`get(id)`→`getByLogical(id)` once the row id moves). Full **core (1680) + gateway (1883) green**; typecheck + lint clean.

## Deferred (flagged)
- cleanDeadRefs content-strip is still an in-place UPDATE on the current version (locally correct; routing it through `appendVersion` matters for sync — sub-PR 3).
- Pre-existing `_importEntries` cross-project overwrite (security follow-up).
- Compaction FK-dangle on a v1-row prune (sub-PR 4 — #823 follow-up).

cc reviewer-session — this is the high-risk flip; the curator superseded-id bugs are the kind of thing worth a close look.

### DIFF
--- a/packages/core/src/agents-file.ts
+++ b/packages/core/src/agents-file.ts
@@ -565,9 +565,11 @@ function _importEntries(entries: ParsedFileEntry[], projectPath: string): void {
       // entry.id is the <!-- lore:UUID --> marker, which is the stable logical_id.
       const existing = ltm.getByLogical(entry.id);
       if (existing) {
-        // Known entry — update only if content changed (manual edit in file)
+        // Known entry — update only if content changed (manual edit in file). Pass
+        // the resolved current row id (Seer #848): update() resolves it to the
+        // logical_id and appends a new version.
         if (existing.content !== entry.content) {
-          ltm.update(entry.id, { content: entry.content });
+          ltm.update(existing.id, { content: entry.content });
         }
       } else if (ltm.isTombstoned(entry.id)) {
       } else {
--- a/packages/core/src/curator.ts
+++ b/packages/core/src/curator.ts
@@ -332,10 +332,12 @@ export function applyOps(
       });
       if (!wasCreated) {
         // Dedup-merged into an existing entry — not a real create, but the
-        // existing row was updated in place and may already be pinned in the
-        // prompt cache. Surface it as an update so session LTM caches are
-        // invalidated and durable prompt deltas can replay the new bytes.
-        const entry = ltm.get(id);
+        // existing entry was updated (a content change appends a new version) and
+        // may already be pinned in the prompt cache. Surface it as an update so
+        // session LTM caches are invalidated and durable prompt deltas can replay
+        // the new bytes. `id` is the stable logical_id from tryCreate → resolve
+        // the current version via getByLogical.
+        const entry = ltm.getByLogical(id);
         if (entry) {
           idsToSync.push(id);
           changedEntries.push({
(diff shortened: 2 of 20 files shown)


BYK:
## Review fixes pushed (`d09048af`)

**Code-quality (SHOULD-FIX):**
- Curator `update`/`delete` handlers resolve `op.id` through `logical_id` (`get ?? getByLogical(logicalIdOf)`) — a second op on the same entry in one `applyOps` batch was silently dropped (incl. update-then-delete leaving the entry alive).
- CLI `list` shows `logical_id`; `show` resolves via `getByLogical`; the list/show API presents `logical_id` as the external id — the `list → show/delete/move` flow was broken for any curator-updated entry.

**Code-quality (CONCERN):** `update()` skips the append (and re-embed) when content is byte-identical — re-observing an unchanged entry no longer grows the table unbounded.

**Security (the conditional blocker):** push is now **version-aware** (`knowledgeSyncOp`): a superseded (`is_current=0`) or soft-deleted (`is_deleted=1`) knowledge row becomes a remote **soft-delete** (propagating deletions + removing stale/redacted content), a current-live row upserts. v1 entries upsert exactly as before (engine tests unchanged). This removes the deletion-doesn't-propagate / redacted-content-egress regression. Full cross-device `id↔logical_id` convergence (so refs resolve the same entry on every device) remains **sub-PR 3**.

New `knowledgeSyncOp` unit test (current→upsert, superseded/death-cert→delete, gone→skip). Core 1681 + gateway 1883 green; typecheck + lint clean.

PR #877, opened 2026-06-21
fix(cache-warmer): drop stale warmup body when idle distillation rewrites the prefix

BYK:
Closes the residual gap flagged in #873's review.

## The gap
`lastRequestBody` is invalidated on `/compact` and model/provider switches, but **not** on lore's **own** idle-time prefix mutations. The idle worker force-distills (creating gen-0 segments that enter the in-context prefix) and meta-consolidates (archiving gen-0, adding gen-1+) — both rewrite the distilled prefix (`messages[1]`) of a **compressed** session. The idle force-distill is *explicitly* preparing a smaller post-idle-compact body. So the body the next real turn sends diverges from the last turn's, and the warmer's stored body is stale — replaying it just refreshes a prefix the next turn busts anyway (the large-session `cacheRead=0` waste #873 reduced but couldn't fully eliminate).

## Fix
After the idle distillation step, null `state.cacheAnalytics.lastRequestBody` when:
- the prefix actually mutated this tick (force-distill ran, or `metaDistill` returned non-null), **and**
- the session is **compressed** (`getLastLayer >= 1`).

Layer-0 (full-passthrough) bodies carry no distilled prefix, so distillation doesn't change them — their warming is **left untouched** (these were the healthy `✓ refresh` sessions). Same rationale/shape as the existing `pipeline.ts` invalidations.

Effect: a compressed session that distills during idle stops being warmed for that idle period — correct, because its next-turn prefix will differ regardless (and the warming would be wasted writes). Aligned with the cache-economics direction (large sessions → cool-bust).

## Tests
The decision is an exported pure predicate `idleDistillInvalidatesWarmupBody(prefixMutated, layer, hasStoredBody)`, unit-tested for all edge cases (layer-0 not invalidated, no-mutation, no-body) — without coupling to the LLM/batch distillation internals the idle-work harness deliberately avoids.

## Verification
typecheck (5 pkgs) clean; lint clean (15 warnings, all pre-existing); full suite **3634 passed**; `pnpm build` clean.

## Note
Combined with #873 (breakpoint preservation), this should take large-session warmups from "every one UNCACHED" to either clean `✓ refresh` (stable prefix) or simply *not warmed* (prefix about to change) — no more paying $1–5 to write a cache the next turn won't read.

### DIFF
--- a/packages/gateway/src/idle.ts
+++ b/packages/gateway/src/idle.ts
@@ -29,6 +29,7 @@ import {
   saveGradientState,
   getConsecutiveBusts,
   effectiveMetaThreshold,
+  getLastLayer,
   evictSession as evictGradientSession,
   distillLimiter,
   curatorLimiter,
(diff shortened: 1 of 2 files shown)


BYK:
Addressed the review's SHOULD-FIX in da49d2d8: `idlePrefixMutated` is now gated on `distillation.run().distilled > 0` (mirrors the `if (meta)` gate). Verified against `distillation.ts` — `distilled` is incremented only when `distillSegment` actually creates a gen-0 segment (line 954); it stays 0 for tiny-segment absorption (line 935), worker failures, parse errors, and expansion-guard discards. So a force-distill that produces no gen-0 no longer nulls a still-valid warmup body (which would have caused the exact cold-return waste this PR prevents). The layer-0→post-idle-compact gap (review 2b) is left as-is — self-correcting via `skipCompact` and governed by the warmer's own stop economics, not idle-distill invalidation.

Issue #881, opened 2026-06-21
cache-economics: refine `cacheSizeCompressed` on layer ≥ 1 paths before PR2b acts on it


BYK:
## Summary

The cache-economics shadow snapshot (`@loreai/core`, `packages/core/src/gradient.ts`) under-reports compaction savings for sessions that are **already compressing** (held at layer ≥ 1). It is **harmless today** because the unified evaluator is shadow-mode/log-only (PR #844 / #846), but it **must be fixed before PR2b** flips consumers to act on the model (cool-bust / opportunistic flush).

This is already flagged in-code as a `KNOWN GAP` and was surfaced again by an adversarial review of #872 (CONCERN-2, "informational — not introduced here").

## Background

Every `transform()` records two numbers on the session state so the shared cache-economics evaluator (and the cache-warmer) read identical sizes from one place:

- `cacheSizeFull` — full, uncompressed body token size.
- `cacheSizeCompressed` — compacted-body token target.

`evaluateCacheStrategy()` compares the two to estimate compaction savings and decide warm-vs-compress.

## The gap

`packages/core/src/gradient.ts`:

- Baseline (~`:2529-2531`) sets `cacheSizeCompressed = cacheSizeFull` (assume no compaction).
- It is only **refined** to a real compressed target inside the **layer-0 tier gate** (~`:2605-2614`, via `compressedEstimate`).
- Sessions held at **layer ≥ 1** by the sticky-layer guard / prefix-present floor / post-idle compact / first-sight-large guard **bypass the tier gate**, so `cacheSizeCompressed` stays `== cacheSizeFull`.

Result: for an already-compressing session, the shared model believes compaction buys ~nothing and **under-reports real savings**.

In-code marker (`gradient.ts:2523-2528`):

> KNOWN GAP (shadow-only, PR2a): the refine ONLY fires on the layer-0 tier gate. Sessions held at layer >= 1 … bypass it, so for an already-compressing session `cacheSizeCompressed` stays == full and the shadow under-reports real compaction savings. Harmless while we only LOG; **PR2b MUST source a real compressed target for the layer >= 1 paths before it acts on cool-bust.**

## Impact

- **Now:** none — shadow/log-only. `evaluateCacheStrategy` does not change behavior.
- **After PR2b:** real — the unified model would wrongly conclude "no compaction benefit" for layer ≥ 1 sessions and could make incorrect warm/compress (cool-bust) decisions, which is exactly the divergence the unification refactor exists to prevent.

[...]

PR #887, opened 2026-06-21
fix(cache-economics): normalize cacheSizeCompressed to cacheSizeFull's input scale

### LINKED ISSUE #886: cache-economics (PR2b prerequisite): normalize cacheSizeFull / cacheSizeCompressed to the same token scale
## Summary

Follow-up surfaced by the adversarial review of #883 (which closed #881). Before PR2b flips any consumer to **act** on `evaluateCacheStrategy`, the two size inputs must be put on the **same token scale**. Today they are not — it is **shadow/log-only, so harmless now**, but it would bias real cool-bust decisions.

## The mismatch (`packages/core/src/gradient.ts`)

- `cacheSizeFull = layer0Input` is **INPUT-scale**: it includes system overhead + tools + LTM, and is multiplied by `UNCALIBRATED_SAFETY` (~×1.5) on uncalibrated first-sight-large turns.
- `cacheSizeCompressed = result.totalTokens` (layer ≥ 1, set at the `transform()` exit) is **BODY-scale**: `distilled + raw` only, **excludes** overhead + LTM, never inflated.

This is the **pre-existing convention** (the old tier-gate refine compared body-scale `compressedEstimate` against input-scale `layer0Input`), which #883 deliberately preserved. But `decideCacheStrategy` (`packages/core/src/cache-economics.ts`) treats `full` and `compressed` as the *same* quantity (tokens written/read to the prompt cache):

- `coolBustCost ∝ compressed`, `coolFullWriteCost ∝ full`, `holdWarmCost ∝ full`.

Since overhead + LTM persist in **both** the full and compacted prompt (compaction does not remove them), a body-only `compressed` **understates `coolBustCost`** → biases the model toward bust/compaction. The `UNCALIBRATED_SAFETY` factor on `full` (but not `compressed`) amplifies this on first-sight-large turns.

## What to do (before PR2b acts)

- Lift `cacheSizeCompressed` to the **cache-input scale**: `compressed_body + overhead + LTM`.
- Reconcile the `UNCALIBRATED_SAFETY` inflation so it is applied consistently (or not at all) across `full` and `compressed`.
- Verify `evaluateCacheStrategy` produces an unbiased full-vs-compressed delta on both calibrated and uncalibrated layer ≥ 1 sessions.

## References
- In-code caveat: `gradient.ts` transform() exit (the `SCALE CAVEAT (PR2b prerequisite ...)` comment).
- Closed by the layer ≥ 1 fix: #881 / PR #883.
- Shared model: #842; shadow wiring: #844 / #846.

BYK:
Closes #886 (the PR2b prerequisite surfaced by the adversarial review of #883).

## Problem
`cacheSizeFull` is **INPUT-scale**: it includes the non-message floor (system/tools overhead + LTM) and is ×`UNCALIBRATED_SAFETY` (1.5) inflated on uncalibrated turns — deliberately, so it tracks the real API token count the warmer cross-checks against (`gradFull` vs `apiActual`). But `cacheSizeCompressed` was set from the raw **body-only** rebuilt window (`result.totalTokens`) — body-scale, no floor, never inflated. So the full-vs-compressed delta **over-reported savings** by:
1. the overhead + LTM **floor** (which compaction never removes — the system prompt + LTM persist in the compacted prompt too), and
2. the **uncalibrated** ×1.5 factor (on full but not compressed).

Both bias `decideCacheStrategy` toward cool-bust. Shadow/log-only today, but wrong once PR2b acts on the model.

## Fix
A compacted prompt still carries the whole floor — only the message body shrinks. So:

```
cacheSizeCompressed = (nonBodyFloor + rebuiltWindowBody) × bodySafety   (clamped to full)
```

- `transformInner` captures the un-inflated **non-body floor** (`overhead + sessLtmTokens`) and the **safety factor** (`1` when calibrated, else `UNCALIBRATED_SAFETY`) alongside `cacheSizeFull`.
- `transform()` computes the compressed size via a new pure helper **`computeCompressedCacheSize`**, inflating floor + body together onto `cacheSizeFull`'s scale.

The full-vs-compressed **delta is now exactly the message tokens compaction removed** (the floor cancels), on the input scale `full` uses. `cacheSizeFull` is unchanged (keeps the `apiActual` cross-check intact).

## No behavior change
Both consumers (`evaluateCacheStrategy` shadow log in the warmer; `getCacheSizeSnapshot`) are still measure-only. This only makes the shadow data faithful.

## Tests
- `computeCompressedCacheSize` unit tests: layer-0 → `== full`; floor retained (calibrated); safety applied to floor+body (uncalibrated); clamp-to-full; safety < 1 treated as 1.
- The #881 e2e test now asserts compressed is **lifted above** the body-only value **and** still `< full` (mutation-guarded both ways).

## Verification
typecheck (5 pkgs) clean; lint clean (15 warnings, pre-existing); full suite **3647 passed**; `pnpm build` clean.

## Follow-on
With this landed, the layer ≥ 1 shadow data is faithful end-to-end — clears the last prerequisite before PR2b flips the warmer/compaction consumers onto the unified `evaluateCacheStrategy`.

### DIFF
--- a/packages/core/src/gradient.ts
+++ b/packages/core/src/gradient.ts
@@ -445,6 +445,15 @@ type SessionState = {
   cacheSizeFull: number;
   /** Compacted-body token target from the most recent transform(); === cacheSizeFull when no compaction is available. */
   cacheSizeCompressed: number;
+  /** Non-message floor (system/tools overhead + LTM) of the most recent
+   *  transform(), in the SAME estimate basis as the message body (un-inflated).
+   *  Compaction never removes this; it is added back to the compressed body so
+   *  cacheSizeCompressed lands on the same scale as cacheSizeFull (issue #886). */
+  cacheNonBodyTokens: number;
+  /** Multiplier that maps the gradient's un-inflated (chars/3) body estimate onto
+   *  cacheSizeFull's scale: 1 when calibrated, UNCALIBRATED_SAFETY when not — the
+   *  same factor layer0Bounds applies to cacheSizeFull. (issue #886) */
+  cacheBodySafety: number;
   /** The single stored strategy decision + when it was computed (null until first evaluated). */
   cacheStrategy: { result: CacheEconomicsResult; decidedAt: number } | null;
 };
@@ -473,6 +482,8 @@ function makeSessionState(): SessionState {
     dedupDecisions: new Map(),
     cacheSizeFull: 0,
     cacheSizeCompressed: 0,
+    cacheNonBodyTokens: 0,
+    cacheBodySafety: 1,
     cacheStrategy: null,
   };
 }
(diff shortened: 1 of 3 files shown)


BYK:
Addressed the adversarial review:

1. **Honest calibrated caveat** (comment over-claim fix):  docstring and  exit now correctly state the delta is EXACT for uncalibrated but APPROXIMATE for calibrated (where full is API-measured vs the gradient's overhead+LTM estimate, which double-counts LTM and absorbs the chars/3 body undercount — same pre-existing entanglement as `usable`). The clamp errs conservative.

2. **M1 and M2 both killed**: two new e2e tests — non-zero LTM floor (exact via `computeCompressedCacheSize` + the real `getOverhead()` + `UNCALIBRATED_SAFETY`) and a calibrated safety=1 path (primes via `calibrate()`). Both assert exact match and guard the body-only / uncalibrated-only regressions.

3. Exported `UNCALIBRATED_SAFETY` from gradient.ts as a test seam.

Full suite 3649 passed; lint + typecheck + build clean.

PR #896, opened 2026-06-21
feat(cache-economics): PR2b — flip warmer + compaction onto unified strategy

BYK:
The behavior flip. The unified cache-economics model (`evaluateCacheStrategy`) now **drives** both the warmer and the compaction decision; legacy heuristics become the fallback for non-confident cases (no snapshot / non-finite inputs). All safety guards are unchanged.

### `shouldWarm` (cache-warmer.ts)
The unified strategy is now the **PRIMARY** warm/no-warm decision:
- **Phase A (initial commitment):** confident `cool-bust`/`cool-full-write` → don't warm (was: `pReturns ≤ threshold`). Non-confident → legacy threshold.
- **Phase B (continuation):** confident strategy flips away from `hold-warm` mid-break → stop (was: rising cumulative threshold). Non-confident → legacy rising threshold.

All safety guards (circuit breaker, subagent, cooldown, ROI, write-efficiency, `maxCycles`, `pFinished>0.95`, margin window) are **unchanged** and still run.

### `onIdleResume` (pipeline.ts)
`skipCompact` now reads `strategyWantsWarming(strategy)` when confident:
- `hold-warm` → skip compaction (keep the warm prefix)
- `cool-bust`/`cool-full-write` → don't skip (let it compact)

Falls back to `isCacheWarm` when non-confident. The shadow compaction log is replaced with a real decision log.

## Why now
Shadow data (post-deploy) confirmed the model produces sane strategies on real sessions:
- `0IJUAT4UePvYD6V4` (859K, pReturn=0.55) → `cool-bust` ($8.05 vs $18.10 hold) — correctly stops warming this 800K session
- `0QsIRMDXDbSDNcOc` (407K, pReturn=0.08) → `cool-full-write` — don't warm, barely returns
- `gradCompressed < gradFull` confirms #883/#887 are working

## Tests
4 new `shouldWarm` tests:
- `hold-warm` overrides a below-threshold `pReturns` → warms
- `cool-bust` blocks warming despite high `pReturns`
- `cool-full-write` blocks warming (low return + large body, no compaction)
- Legacy fallback when no snapshot (non-confident)

## Verification
typecheck (5 pkgs) clean; lint clean (19 warnings, all pre-existing/non-touched); full suite **3742 passed**; `pnpm build` clean.

## Follow-on
PR3: gate deferred LTM/distillation flush on the `cool-bust` event (independent schedule, bust as preferred trigger, urgency override).

### DIFF
--- a/packages/gateway/src/cache-warmer.ts
+++ b/packages/gateway/src/cache-warmer.ts
@@ -33,6 +33,7 @@ import {
   evaluateCacheStrategy,
   strategyWantsWarming,
   getCacheSizeSnapshot,
+  getCacheStrategy,
 } from "@loreai/core";
 import type {
   InterTurnHistogram,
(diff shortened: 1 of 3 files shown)


BYK:
Addressed the adversarial review (NEEDS-REWORK → SHOULD-FIX):

**MUST-FIX (point 6) — FIXED:** `skipCompact` now preserves the `isCacheWarm` liveness floor. Extracted as pure exported `decideSkipCompact(econ, cacheIsLive)`: confident hold-warm → skip ONLY if cache is live (intent + liveness, both required); confident cool-bust/full-write → never skip; non-confident → `cacheIsLive` alone (byte-identical legacy). The stale-hold-warm-with-expired-cache regression is now covered by a unit test.

**Test fixes:** `setCachePricing` arg order corrected (write, read); dead `evaluateCacheStrategy` import removed; 5 new `decideSkipCompact` unit tests covering the MUST-FIX regression and the full matrix.

**NITs deferred:** warmer fallback log doesn't include `confident=` (low priority — the compaction log does); Phase B mid-break flip and stale-strategy tests are covered by the `decideSkipCompact` unit tests at the decision level.

Full suite 3747 passed; typecheck + lint + build clean.

PR #897, opened 2026-06-22
feat(sync): re-key remote knowledge sync on logical_id (A2 sub-PR 3)

BYK:
**Epic #821 · A2 (#823), sub-PR 3.** Completes the append-only knowledge story across the sync boundary. **Client-only — no remote migration** (pre-flip rows were keyed `id == logical_id` (v1) already, and sync is release-gated so the remote has no real data to migrate).

Makes the remote a **CURRENT-only mirror, ONE row per logical entry keyed by `logical_id`** — coalescing all of an entry's versions to a single remote row and converging `id`↔`logical_id` across devices.

## Model (settled)
A remote content change to an existing `logical_id` is "same concern, new value" arriving from another device → **append a new local version** (mirrors a local `update()`). Idempotency comes from `content_hash` (re-pulling your own pushed content is a no-op → no ping-pong). Version *history* stays local; only *current content* converges.

## Push — `knowledgePushPlan` (replaces `knowledgeSyncOp`)
Resolve the outbox version row → `logical_id`; push the **current** version's content under `id = logical_id`, or **soft-delete** `id = logical_id` when there's no live current. `sync_state` keyed by `logical_id`, so an entry's version churn coalesces to one upsert (`content_hash` dedups the rest). `pushEntry` drives upsert/delete by the `logical_id` (`effectiveId`).

## Pull — `applyRemoteKnowledge` / `applyRemoteKnowledgeDelete`
Apply a `logical_id`-keyed remote row into the local **append-only** model (under apply-suppression + a txn for the single-current invariant):

| local state for L | remote live (upsert) | remote deleted |
|---|---|---|
| none | INSERT v1 (id=L) | no-op |
| live, content == remote | converge metadata in place | death-cert |
| live, content != remote | append new current version | death-cert |
| only superseded/death-cert | append new current (revive) | no-op |

`hasPendingKnowledgeChange` resolves pending edits **by `logical_id`** (the outbox holds version ids), so a pull never silently overwrites an unpushed local edit without logging the conflict. `applyRemoteUpsert` no longer special-cases knowledge (entity graph only).

## Why no engine-behavior change
v1 entries (`id == logical_id`) push/apply exactly as before, so the existing engine + property suites keep their semantics; only the live-set comparison in the convergence property moves to `logical_id` over current-live versions.

## Tests
- `applyRemoteKnowledge` matrix: new / append-on-change / idempotent / metadata-only / delete→death-cert / revive.
- **Round-trip**: device A's multi-version entry → push plan → peer-apply on B → **ONE coalesced version** on B (proves convergence).
- `knowledgePushPlan` coalescing (superseded + current → one logical-keyed upsert).
- Property **convergence** now compares the live set by `logical_id` over current-live versions — **8/8 stable**, 60 random op-sequences each.
- Core **1741** + gateway **1996** green; typecheck + lint clean.

## Resolves / unblocks
Remote row accumulation, residual-byte persistence, cross-device `id`↔`logical_id` divergence → **lifts the release gate** on `lore sync enable`. Follow-ups: knowledge_meta convergent confidence (sub-PR 3b), compaction/GC (sub-PR 4).

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1488,12 +1488,21 @@ function installSyncCapture(database: Database) {
     "entity_relations",
   ]) {
     for (const [evt, suffix, ref, op] of ops) {
+      // Knowledge is remote-keyed by logical_id (A2 sub-PR 3, #823). INSERT/UPDATE
+      // capture the version id (op=upsert; the push plan resolves it to the current
+      // logical row while the row still exists). DELETE must capture the logical_id
+      // NOW — the row is gone afterward, so the push can't resolve it, and a delete
+      // keyed by a version id would never match the logical_id-keyed remote row.
+      const rowExpr =
+        t === "knowledge" && op === "delete"
+          ? `COALESCE(${ref}.logical_id, ${ref}.id)`
+          : `${ref}.id`;
       sql += `
         CREATE TEMP TRIGGER IF NOT EXISTS ${t}_outbox_${suffix}
         AFTER ${evt} ON ${t} WHEN (${gate})
         BEGIN
           INSERT INTO sync_outbox (table_name, row_id, op, changed_at)
-          VALUES ('${t}', ${ref}.id, '${op}', ${ts});
+          VALUES ('${t}', ${rowExpr}, '${op}', ${ts});
         END;`;
     }
   }
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -526,6 +526,41 @@ export function seedOutbox(tier: SyncTier = "basic"): void {
       // (via the prune floor) permanently disabling outbox pruning for ALL tables.
       if (m.pullOnly) continue;
       const pushCursor = Number(getKV(`sync.push.${m.table}`) ?? "0");
+
+      // Knowledge is remote-keyed by logical_id (A2 sub-PR 3, #823): seed ONE entry
+      // per CURRENT live logical entry — NOT every physical version — keyed + hashed
+      // by logical_id to match the push plan + sync_state. Iterating physical rows
+      // and checking sync_state by version id would miss every v2+ entry (sync_state
+      // is keyed by logical_id) and re-enqueue it (outbox bloat). Deleted entries
+      // (no live current) are NOT seeded here — reconcile's tombstone pass emits the
+      // delete (its liveness check is also knowledge_current-based).
+      if (m.table === "knowledge") {
+        const latestForLogical = db().query(
+          `SELECT o.op FROM sync_outbox o JOIN knowledge k ON k.id = o.row_id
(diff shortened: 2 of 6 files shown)


BYK:
## Review fixes pushed (`3c251810`)

**BLOCKER (Seer + adversarial review) — fixed.** `classifyRemoteRow` + the conflict snapshot read the local row via `getRowById(id=logical_id)`, which after any update is the **demoted v1** row — so every versioned (v2+) entry's clean fast-forward / self-echo was mis-classified as a **conflict** (inflated `sync_conflicts`, stale `local_content`). Fix: a version-aware `currentKnowledgeRow(logicalId)` (reads `knowledge_current`, re-keyed `id=logical_id` — the exact shape push hashes) used in `classifyRemoteRow` and the `localBefore` snapshot; `knowledgePushPlan` reuses it.

**Testing gap (root cause the BLOCKER shipped) — fixed.** The property op-model did raw in-place `UPDATE` / physical `DELETE`, so it only ever produced **v1** rows and never exercised the append-only paths — the convergence/no-ping-pong invariants were vacuous for versioned entries. Now `update`/`delete` go through `ltm.update`/`ltm.remove` (real append / death-cert), the no-ping-pong test asserts `conflicts==0`, and a targeted engine test does `insert → appendVersion → push → pull ⇒ {pulled:0, conflicts:0}`. **Verified both fail (red) without the classify fix and pass with it.**

**Security SHOULD-FIX #1 — fixed.** The soft-delete push scrubs knowledge `content`/`title` to `''` on the remote tombstone (erasure completeness). Local death-cert preserves content; peers rebuild their death-cert from their own current row, so they're unaffected.

**Security SHOULD-FIX #2 (admin bulk physical-delete)** is pre-existing + out of scope → tracked for sub-PR 4 on #823.

Core 1741 + gateway 1997 green; property suite 5× stable; typecheck + lint clean.

BYK:
Re: Seer `seedOutbox` LOW (v1-anchor reliance under future compaction) — **acknowledged, tracked for sub-PR 4, not a current bug.**

The `latestForLogical`/`hasPendingKnowledgeChange` JOINs (and `knowledgePushPlan`) resolve a logical entry via the v1 anchor row (`id == logical_id`), which is **never physically deleted today** — the only physical knowledge deletes are whole-entity wipes (all versions). The reliance only breaks when sub-PR 4 compaction prunes superseded versions including the anchor. That's a compaction design constraint, recorded on #823 with two resolutions (preserve the anchor, or convert the knowledge outbox to full `logical_id` keying — the recommended end-state). The #897 landmine guard (re-validate liveness on the delete push) already defuses the most dangerous symptom (a superseded-version delete tombstoning a live remote entry). Folding the keying refactor into this PR would violate scope discipline on a finished, reviewed change.

PR #910, opened 2026-06-22
fix(gateway): per-model client usage cap + message_delta usage leak

BYK:
## Problem

Lore spoofs the `usage` it reports to the downstream agent (Claude Code) so the client's context meter / auto-compact trigger never fire prematurely. Two bugs made this misbehave — most visibly, a 1M-context session (`claude-opus-4-8`) that grew to ~960K, showed "2% remaining", and hard-locked.

**How the client actually counts** (verified against `freecodexyz/free-code`, a faithful Claude Code reimplementation): the meter and auto-compact trigger read **API-reported usage** — `input_tokens + cache_creation + cache_read + output_tokens` from the last response (`tokenCountWithEstimation` → `getTokenCountFromUsage`) — over a **per-model** context window. So the spoof targets the right fields; it was just miscalibrated and leaking.

### (a) The cap was hardcoded to a 200K model
`AUTOCOMPACT_THRESHOLD = 167_000` → cap `150,300` for **every** model. On a 1M-context model that's wrong by ~6×; on smaller models it can trigger premature compaction. (Flagged previously for a MiniMax 200K model — "is this threshold correct for ALL models?")

### (b) `message_delta` leaked the real usage
Anthropic's **terminal `message_delta` carries the full cumulative usage** (`input_tokens`, `cache_read_input_tokens`, `cache_creation_input_tokens`), but only `output_tokens` was scaled. Since the client's stored usage is **last-write-wins**, the real ~960K from `message_delta` overwrote the scaled `message_start` — defeating the spoof even on 200K models. Existing fixtures only ever put `{output_tokens}` in `message_delta`, so no test caught it.

## Changes

- **Per-model cap** (`compaction.ts`): pure `maxReportedUsageForModel(contextWindow, maxOutput)` = `floor(0.9 × ((window − min(maxOutput, 20k)) − 13k))`, mirroring Claude Code's `autoCompact.ts`. `scaleUsageForClient(usage, cap)` now takes the cap; `DEFAULT_MAX_REPORTED_USAGE` (= 150,300) preserves 200K behavior. Threaded via `maxReportedUsageForModelID` (models.dev window lookup) into both stream accumulators and `nonStreamHttpResponse`. **200K models are unchanged.**
- **`message_delta` fix** (`stream/anthropic.ts`): shared `scaleMessageDeltaUsage()` scales **every** field the delta carries, in `createStreamAccumulator`, `createRecallAwareAccumulator`, and the recall **continuation** stream (which previously didn't scale at all). The delta's own fields are the authoritative scale basis, so an asymmetric `message_start` (multi-provider translation) can never re-leak a raw value.
- `cost-tracker.ts` keeps the 200K `AUTOCOMPACT_THRESHOLD` constant (dashboard estimate) with a TODO.

## Out of scope (follow-up)
`DISABLE_AUTO_COMPACT`: with auto-compact disabled, the threshold becomes the full effective window and a session can still grow to the blocking limit. A correct meter only delays that — at the true ceiling something must reset the buffer.

## Testing
- New unit tests: per-model cap (200K regression-lock at 150,300; 1M → 870,300), `scaleMessageDeltaUsage` (full-field scaling, output-only delta, **asymmetric no-re-leak**), and recall-accumulator scaling (forward + held-back continuation).
- **Mutation-verified** every new test (revert the fix → test fails): per-model cap, `message_delta` field scaling (both accumulators), and the asymmetric re-leak guard.
- `typecheck` + `lint` clean. Full suite green (**3857 passed**), 3× for flakiness.

🤖 Adversarial-reviewed (READ-ONLY): SHIP-WITH-FOLLOWUPS, no blockers; both SHOULD-FIX items (recall coverage gap, asymmetric re-leak) addressed in this PR.

### DIFF
--- a/packages/gateway/src/compaction.ts
+++ b/packages/gateway/src/compaction.ts
@@ -78,33 +78,75 @@ export function estimateTokens(text: string): number {
 // ---------------------------------------------------------------------------
 
 /**
- * Claude Code's auto-compact threshold for a 200K context model:
+ * Claude Code's auto-compact arithmetic (mirrors `freecodexyz/free-code`
+ * `src/services/compact/autoCompact.ts`, which is a faithful reimplementation):
  *   effectiveContextWindow = contextWindow - min(maxOutputTokens, 20_000)
- *   autoCompactThreshold  = effectiveContextWindow - 13_000
+ *   autoCompactThreshold   = effectiveContextWindow - 13_000
  *
- * For 200K context → 167K.  We report at most TARGET_RATIO × 167K ≈ 150K
- * so the client's "X% until auto-compact" UI grows naturally but never
- * triggers the compaction.
+ * The client's context meter and auto-compact trigger both read API-reported
+ * usage (`input_tokens + cache_creation + cache_read + output_tokens` from the
+ * last response). We report at most `USAGE_REPORT_RATIO × autoCompactThreshold`
+ * so the client's "X% until auto-compact" UI grows naturally but never trips.
+ *
+ * The cap is **per-model**: it must be derived from the model's real context
+ * window, not hardcoded. A hardcoded 200K cap (the old behavior) reports ~150K
+ * for every model, which on a 1M-context model is wrong by ~6× and on smaller
+ * models can trigger premature compaction.
+ */
+export const AUTOCOMPACT_BUFFER_TOKENS = 13_000;
+/** Tokens Claude Code reserves for the compaction summary output. */
+export const MAX_OUTPUT_RESERVE = 20_000;
+/** Report at most this fraction of the auto-compact threshold. */
+export const USAGE_REPORT_RATIO = 0.9;
+
+/**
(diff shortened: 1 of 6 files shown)


BYK:
@onurtemizkan heads up — this lands the fix for the context-meter lockup you hit on `claude-opus-4-8`.

**Root cause (two bugs):**
1. Lore's anti-compaction usage-scaling was hardcoded for a 200K model, so on a 1M-context model the cap was miscalibrated.
2. More importantly, the terminal `message_delta` was leaking the real cumulative usage (only `output_tokens` was being scaled), so your client's context meter ended up showing the true ~960K → "2% remaining" → hard lock at the blocking limit.

**Fix:** the reported-usage cap is now derived per-model from the real context window, and *every* usage field in `message_delta` is scaled (so the spoof actually sticks). 200K models are unchanged.

It'll be in the **next nightly** — once you're on it, a 1M session should track correctly and stop hard-locking. If you still see the meter misbehave after updating, ping me with a fresh session and I'll dig in.

Issue #909, opened 2026-06-22
A2 sub-PR 4: knowledge compaction/GC + remote hard-delete reaper


BYK:
Part of the A2 append-only knowledge epic (#823). Splits into a **client** piece (local compaction, buildable now) and a **server-side** piece (remote reaper, lands with team-sharing infra).

## Prerequisite (separate small PR, in progress)
**Full `logical_id` outbox keying for knowledge.** Today the INSERT/UPDATE capture triggers record the *version id* and only DELETE records `COALESCE(logical_id, id)`; `knowledgePushPlan` / `seedOutbox.latestForLogical` / `hasPendingKnowledgeChange` resolve back to the logical entry via `JOIN knowledge k ON k.id = o.row_id`, which **relies on the v1 anchor row (`id == logical_id`) physically existing**. Compaction that prunes the anchor would silently break those lookups (dropped pushes / missed pending edits). Fix: capture `COALESCE(NEW.logical_id, NEW.id)` for ALL knowledge ops, then those three call sites treat the outbox `row_id` AS the logical_id (no JOIN, no anchor dependency). This retires the tracked Seer/anchor fragility and unblocks compaction to prune any version. (Confirmed direction over the cheaper "never delete the anchor" rule.)

## Client: local version compaction
Bound local storage by pruning superseded (`is_current=0`) version rows, **under sync-suppression** (local-only cleanup, no outbox entries):
- **Always keep the head** — the current live version, or the death-cert for a deleted entry.
- **Keep up to the 2 most-recent superseded versions** per `logical_id` (count cap).
- **Hard-drop any superseded version older than ~60 days** (age cap), even within the count.
- Rationale: the `logical_id + version` history *is* the knowledge-evolution trajectory (subsumes #495), so it has product value — we keep a bounded recent window rather than pruning to current-only, while preventing unbounded growth. The current/death-cert head is never touched, so the single-current invariant and deletion propagation are unaffected. Numbers are sensible defaults; can become config.

## Server-side: remote hard-delete reaper (deferred to team-sharing infra)
Soft-deleted remote tombstones (content already scrubbed to ''/null on delete, #897) must eventually be physically purged. This belongs as a **server-side system task** (Supabase pg_cron or a scheduled edge function) — centralized, independent of any client being online, and a natural fit alongside the team-sharing backend. **Do NOT make this client-driven.**

### Cross-cutting constraint (lock now so the pieces compose)
[...]

PR #913, opened 2026-06-22
feat(sync): key the knowledge outbox by logical_id for all ops (#909 prereq)

BYK:
**Prereq for #909 (A2 sub-PR 4 compaction).** Makes the knowledge change-capture uniformly **`logical_id`-keyed** so compaction can prune any version — including the v1 anchor — without breaking sync.

## Problem
Today INSERT/UPDATE capture the *version id* and only DELETE captures `COALESCE(logical_id, id)`. So `knowledgePushPlan`, `seedOutbox.latestForLogical`, and `hasPendingKnowledgeChange` resolve a logical entry via `JOIN knowledge k ON k.id = o.row_id`, which **relies on the v1 anchor row (`id == logical_id`) physically existing**. sub-PR 4 compaction prunes superseded versions (potentially the anchor) → those lookups would silently return null (dropped pushes / missed pending edits). Flagged by the #897 final review + Seer (LOW, forward-looking).

## Change
- **`db.ts` `installSyncCapture`** — knowledge captures `COALESCE(logical_id, id)` for **INSERT/UPDATE/DELETE** (was the version id for INSERT/UPDATE). Every knowledge outbox `row_id` is now the logical_id.
- **`knowledgePushPlan(rowId)`** — `rowId` IS the logical_id → `currentKnowledgeRow(rowId)` directly, no anchor resolve. Live → upsert (current content); no live current → delete.
- **`hasPendingKnowledgeChange` / `seedOutbox.latestForLogical`** — drop the JOIN; match `row_id = logical_id` directly.

## Why it's safe today
Behavior-equivalent for current code: v1 entries have `COALESCE(logical_id,id) = id`, so the outbox keys are unchanged; the only difference is versioned entries now coalesce at capture time instead of via the JOIN. The win is **compaction-safety** for sub-PR 4.

## Tests
- Rewrote the `knowledgePushPlan` unit test for the logical_id contract (no version ids reach it; unknown id → idempotent delete).
- Added a trigger test: `create + appendVersion + remove` coalesce to a **single** logical_id outbox `row_id` (**verified red** against the version-id trigger).
- Core **1766** + gateway **2057** green; property suite stable; typecheck + lint clean.

> Branched off #904 (last green) to stay clear of the in-flight #905 idle red; will rebase onto green main before merge.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1524,13 +1524,14 @@ function installSyncCapture(database: Database) {
     "entity_relations",
   ]) {
     for (const [evt, suffix, ref, op] of ops) {
-      // Knowledge is remote-keyed by logical_id (A2 sub-PR 3, #823). INSERT/UPDATE
-      // capture the version id (op=upsert; the push plan resolves it to the current
-      // logical row while the row still exists). DELETE must capture the logical_id
-      // NOW — the row is gone afterward, so the push can't resolve it, and a delete
-      // keyed by a version id would never match the logical_id-keyed remote row.
+      // Knowledge is remote-keyed by logical_id (A2, #823), so capture the
+      // logical_id for ALL ops (not the per-version row id). This makes the outbox
+      // uniformly logical_id-keyed: the push plan + pending checks read the row_id
+      // directly with no join back to a physical version row — so they survive
+      // compaction that prunes the v1 anchor (#909). DELETE must capture it anyway
+      // (the row is gone afterward); INSERT/UPDATE capture it for uniformity.
       const rowExpr =
-        t === "knowledge" && op === "delete"
+        t === "knowledge"
           ? `COALESCE(${ref}.logical_id, ${ref}.id)`
           : `${ref}.id`;
       sql += `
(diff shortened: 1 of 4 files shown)


BYK:
Re: Seer `hasPendingKnowledgeChange` LOW (pre-existing version-id-keyed outbox rows) — **not a real issue; resolving by reasoning, no migration.**

The version-id-keyed knowledge outbox **only ever existed in unreleased dev builds**. `lore sync enable` is still pre-GA — the release gate was just lifted in #897 and no release has been cut since — so **no released version ever wrote a version-id-keyed knowledge outbox row**. There is no installed base with pre-existing pending entries to migrate, and the first release that ships sync will be logical_id-keyed from the start (this PR + its predecessors are all pre-GA). So Seer's "users with pending changes created before this deployment" cannot exist.

Defense-in-depth even if it could: `reconcile()` (run on every `lore sync enable`) re-seeds the outbox **by content** keyed by logical_id (`seedOutbox` compares `currentKnowledgeRow` hash vs `sync_state`), so any stale version-id entry is superseded by a correct logical_id one; and push processes the outbox by `seq` before pull, draining stale rows below the cursor before `hasPendingKnowledgeChange` is ever consulted. A version-id→logical_id data migration for non-existent data would be over-engineering. Tracking nothing further.

Issue #870, opened 2026-06-21
Support Claude via AWS Bedrock and Google Vertex AI through the gateway


BYK:
## Motivation

Users with enterprise Claude access via **AWS Bedrock** or **Google Vertex AI** cannot route through the lore gateway today. Worse, the current `setup`/`run` paths set `env.ANTHROPIC_BASE_URL` while the user's `CLAUDE_CODE_USE_BEDROCK=1` (or `CLAUDE_CODE_USE_VERTEX=1`) is still active, producing a confusing broken state (surfaced in user feedback — Claude Code silently stopped connecting).

Goal: **actually support Bedrock and Vertex through the gateway.** Detect-and-refuse is only a fallback if a clean implementation turns out to be infeasible.

## Why it's non-trivial

The **message body** for Claude on Bedrock/Vertex is essentially the Anthropic Messages format (that's why Anthropic ships `AnthropicBedrock` / `AnthropicVertex` SDK clients). What differs is everything around it:

- **Auth.** Bedrock = AWS **SigV4** request signing; Vertex = GCP **OAuth2 bearer** tokens. Lore today forwards the client's auth header as a transparent proxy. SigV4 signs the host+path, so once the client points at `127.0.0.1:lore` the signature is invalid — you cannot MITM a SigV4 request. The proxy must become the signer.
- **Routing/envelope.** Model is in the URL (Bedrock `bedrock-runtime.<region>.amazonaws.com` `InvokeModelWithResponseStream`; Vertex `<region>-aiplatform.googleapis.com` `:streamRawPredict` under a project/publisher path), and `anthropic_version` differs (`bedrock-2023-05-31` / `vertex-2023-10-16`).
- **Streaming.** Bedrock streams **AWS binary event-stream framing**, not SSE; lore must decode it back to SSE for the client.

## Proposed architecture: lore as translator + credential holder

The client speaks **plain Anthropic** to lore (i.e. `ANTHROPIC_BASE_URL=lore`, and **not** the `CLAUDE_CODE_USE_BEDROCK/VERTEX` flags). Lore holds the cloud credentials and translates outbound. This fits the existing pipeline, which already does Anthropic<->OpenAI translation and four-tier upstream routing (`X-Lore-Provider` / `X-Lore-Upstream-URL` / model-prefix / config default, `pipeline.ts:2857`).

- Add `bedrock` and `vertex` as upstream protocols in the routing/translation layer.
[...]

PR #925, opened 2026-06-22
feat(gateway): SigV4-sign Bedrock worker + cache-warmer calls

BYK:
## Summary
Completes AWS Bedrock support (follow-up to #898). Background workers
(distillation, curation, query-expansion) and the cache-warmer now **SigV4-sign**
their Bedrock calls, so a Bedrock-only session actually builds memory.

Before this, those calls hit `bedrock-runtime` **unsigned → 403**, making the
feature effectively useless unless the user pointed `LORE_WORKER_UPSTREAM` /
`LORE_WORKER_API_KEY` at a non-Bedrock Anthropic-compatible endpoint (the
documented limitation from #898).

## Worker path (`llm-adapter.ts`)
- `WorkerProtocol` gains `"bedrock"`; `resolveWorkerProtocol` no longer collapses
  bedrock → anthropic.
- `buildBedrockWorkerRequest`: non-streaming InvokeModel URL + Bedrock body
  (`buildBedrockWorkerBody` — sentinel, **no** `stream`). SigV4 headers
  (Authorization, x-amz-date, host) are added in the **retry loop on every
  attempt** (x-amz-date must be fresh). No cch/billing — auth is entirely SigV4.
  The response is Anthropic-shaped, so the existing parser is reused.
- `createGatewayLLMClient` takes `{ bedrock: { region, profile } }` from config.
- The batch queue already auto-bypasses Bedrock (no batch provider registered →
  falls through to the direct/inner client), so no batching change is needed.

## Cache-warmer (`cache-warmer.ts`)
- `CacheWarmingProfile` gains an `auth` mode. `buildBedrockProfile` +
  `prepareBedrockWarmupBody` (max_tokens:1, drop `stream`) target a
  non-streaming InvokeModel warmup, SigV4-signed. `resolveProfile` handles
  `bedrock`.
- `executeWarmup` branches: Bedrock needs **no client credential** (SigV4 is the
  auth) and skips cch / anthropic-version.
- Region is recovered from the session's `bedrock-runtime` URL (falls back to
  config); AWS profile from config (honors `LORE_BEDROCK_PROFILE`).

## Tests (all signing paths verified non-vacuous via swap-fail-restore)
- `buildBedrockWorkerBody`, `regionFromBedrockUrl`, `resolveWorkerProtocol`
  bedrock.
- End-to-end signed worker InvokeModel call (asserts URL, SigV4 Authorization
  with `bedrock-runtime` + signed host, Bedrock body, no client key).
- `resolveProfile`/`buildBedrockProfile`/`prepareBedrockWarmupBody`.
- Signed `executeWarmup` integration test (`bedrock-warmup.test.ts`).

## Test plan
- [x] `pnpm run typecheck` clean
- [x] `pnpm run lint` clean
- [x] `pnpm test` — 3990 passed
- [ ] Manual E2E against a real Bedrock account (worker distillation + warmup)

Closes the worker-SigV4 follow-up of #870.


BYK:
Closing in favor of the bedrock-mantle approach. Real-endpoint testing revealed that AWS's recommended `bedrock-mantle` endpoint speaks the native Anthropic Messages API with a simple Bedrock API key (x-api-key) — no SigV4, no InvokeModel, no binary event-stream decoding, and no cross-region inference profiles. That lets us route Bedrock through the gateway's existing `anthropic` protocol (translation/SSE/caching/warming for free), making this worker-SigV4 work unnecessary. Pivoting to mantle and removing the runtime/SigV4 code on `feat/bedrock-mantle`. (This work surfaced the critical `service: bedrock-runtime → bedrock` signing bug, now moot.)

PR #926, opened 2026-06-23
fix(core): exempt meta-distillation prefix-rewrites from unsustainable warning

BYK:
## Problem

The `unsustainable conversation` warning fired on a bounded session (opencode `ses_10f932a26ffeFBKO9ZsjycCCna` → lore `0XrHNdlsWwgVkX4MH`, ~130K tokens — well under the 200K layer-0 cap and 800K+ usable). Investigation showed **every** non-idle cache bust in that session had the same cause:

```
divergence="messages[0].content[0].text"
reason="distilled conversation prefix changed (meta-distillation rewrite)"
```

This is the `prefix-rewrite` bust class. Meta-distillation consolidates gen-0 segments and rewrites the synthetic distilled prefix (messages[0/1]) on the next turn — a real prompt-cache bust, but **self-inflicted by Lore's own distillation pipeline**, not user-context growth. `recordCacheUsage` counted it identically to genuine growth, so the counter accumulated `0→1→2→3` and tripped the warning.

A feedback loop made it worse: busts → bust pressure → lowered meta-threshold (20→5) → meta fires during a brief idle → rewrites the prefix → more busts.

## Fixes

**(1) Exempt `prefix-rewrite` busts from `consecutiveBusts`.** Thread the categorized bust cause (already computed in `pipeline.ts` via `categorizeBust`) into `recordCacheUsage` and hold the counter on `prefix-rewrite` the same way idle-resume busts are held. `CacheBustCause` moves to `@loreai/core` (canonical) so `recordCacheUsage` can use it without a circular gateway→core import; `cache-analytics.ts` re-exports it. Scope is `prefix-rewrite` **only** — `system-ltm-change`/`system-host-change` still count (a burst of those usually indicates real upstream churn worth flagging).

**(2) Gate the bust-pressure meta-threshold override.** `effectiveMetaThreshold` now requires a deep-idle signal (5 min) before lowering the threshold under bust pressure, and the floor is raised from 3 to 10. Without the deep-idle gate, a churning session that briefly goes idle would fire meta right before the user returns — busting the cache they just paid to re-warm. The floor prevents the meta cadence from collapsing to "every 5 gen-0 segments" (too eager, since meta itself causes prefix-rewrite busts).

Note: meta-distillation remaining **destructive** (rewriting the prefix) is by design — it shrinks the prefix for longer conversations. The fix targets the *warning misattribution*, not the meta behavior.

## Tests

- 17 new tests (8 prefix-rewrite exemption + 9 `effectiveMetaThreshold` branches, including boundary cases).
- Full suite green (4001 passed), 3 consecutive runs, no flakiness.
- Mutation-tested both fixes (7 mutations, all killed expected tests).
- typecheck + lint clean.


BYK:
## Adversarial review complete — findings addressed

Ran a dedicated read-only adversarial review (mutation testing in an isolated worktree). Both production fixes' core functions were confirmed **non-vacuously tested** (mutations A/B/C each killed the predicted tests). Two SHOULD-FIX items found and addressed:

**1. Flaky test (fixed).** `effectiveMetaThreshold` read `Date.now()` internally, so the '1ms before the deep-idle gate' boundary test raced scheduling jitter (~13% failure in isolation). Made the clock injectable (`nowMs`, defaults to `Date.now()`); all boundary tests now use a fixed clock. Verified deterministic across 5 consecutive runs.

**2. Untested wiring (idle side fixed; pipeline side tracked).**
- **idle.ts** (`getLastTurnAt` threading): added two `idle-work.test.ts` tests — under bust pressure a *recent* turn defers meta-consolidation while a *deep-idle* turn runs it. Mutation-verified: passing a constant `0` makes the 'recent defers meta' test fail.
- **pipeline.ts** (`recordCacheUsage(..., bustCause)` threading): could not be guarded faithfully in this PR — the gateway vitest config aliases `@loreai/core` to source (so a barrel `vi.mock` doesn't intercept the pipeline import), and the simulated cache can't produce the `cacheRead===0` full-bust that `categorizeBust` needs to emit `prefix-rewrite`. Tracked as #928. Residual risk is low: both endpoints of the wire are unit-tested and `bustCause` is also consumed by 3 adjacent telemetry calls in the same block.

NITs (stale `NON_USER_BUST_CAUSES` comment, misleading test title, `db.ts` core-owned doc) also fixed.

Overall reviewer verdict: **SHIP-WITH-FOLLOWUPS** → all ship-blocking items resolved; #928 is the lone follow-up.

Issue #916, opened 2026-06-22
docs/eval: publish refreshed token-reduction + cost benchmarks (mex-style before/after)


BYK:
## Summary

Publish refreshed, easy-to-grok token-reduction benchmarks. [mex](https://github.com/mex-memory/mex) leads with a simple per-scenario "before vs after tokens (~60% reduction)" table. Lore already has a richer eval stack **and** a published Benchmarks section — this is about surfacing the cost-dimension numbers, adding a simple before/after framing, and keeping them fresh.

## Current state (more built-out than it looks)

- **Eval stack exists:** `packages/core/eval/` — 5 dimensions (`context`, `recall`, `preferences`, `cross-project`, `cost`), 6 baselines (`lore`, `lore-context-only`, `lore-memory-only`, `tail-window`, `compaction`, `raw`), full `CostMetrics` (`totalCostWithLore`, `totalCostBaseline`, `loreOverheadPct`, `savingsPct`, `breakdown`, `counterfactual.{avoidedCompactions, avoidedCompactionCost, cacheHitRate}`), COST-1..5 scenarios (short / long / multi-session), and a weekly `.github/workflows/eval.yml` (fixture + live modes).
- **Benchmarks already published** in `README.md` (§ Benchmarks): recall accuracy 85% vs 50%, 2.5× less input, 4.4× cheaper, 7.4× cheaper per correct, 19× distillation compression — on Sonnet 4.

## Gaps (this is NOT "build the eval stack")

1. **Cost-dimension numbers aren't published.** The README covers recall accuracy + cost-per-correct, but not the cost scenarios' headline figures: short-session overhead % (COST-2, target <15%), long-session savings % vs compaction baseline (COST-3), avoided-compaction count/cost, and cache hit rate. Run the live cost eval and surface these.
2. **No simple before/after token hook.** Add a mex-style "tokens per session: before X → after Y (N% reduction)" table alongside the richer numbers — the easy-to-grok headline.
3. **Freshness / coverage.** Numbers are single-model (Sonnet 4) and hand-curated. Wire the existing weekly `eval.yml` to refresh published figures (or at least date+model stamp them), and consider a second model.

## Acceptance

- README/website benchmark section includes per-session **token reduction** and **cost overhead/savings** numbers sourced from the cost dimension.
- Each figure carries a reproducible command + date + model stamp.
- (Stretch) the weekly eval workflow updates or validates the published numbers.

## Notes

[...]

PR #930, opened 2026-06-23
feat: surface budget-overflow knowledge as a recall-by-id ToC (#917)

### LINKED ISSUE #917: explore: surface budget-overflow knowledge tail as a recall-able ToC
## Summary

**Exploration (not a commitment to ship).** Investigate surfacing the budget-**overflow tail** of knowledge entries as a compact, recall-able table of contents, so the agent can discover entries it currently can't see.

## Problem

`forSession()` (`packages/core/src/ltm.ts`) scores entries by relevance and greedily packs them into a **hard token budget**; over-budget entries are silently dropped (Pass 2 `continue` at the packing loops). The knowledge-delta message's "Additional Changed Knowledge (truncated)" overflow only lists *changed* entries, sliced to 3 (`pipeline.ts:1068-1086`).

Net effect: entries that **exist but rank below the budget cut** are invisible to the agent. The `recall` tool *can* find them, but only if the agent already suspects they exist and searches — it can't recall what it doesn't know is there.

## Hypothesis

A compact ToC (title + category + short id, **no content**) of the overflow tail, appended cheaply, makes those entries discoverable for recall-on-demand without paying full-content injection cost — improving recall accuracy on facts that live in overflow entries at acceptable token cost.

## Risks / open questions

- **Budget competition:** the ToC competes for the same token budget — must be tiny (ids+titles only) and capped to a small fraction.
- **Cache stability:** must be either part of the frozen system[1] baseline or the durable delta, with deterministic ordering (by id) to avoid cache busts (cf. the existing overflow-slice ordering gotcha).
- **Agent dependence:** only helps if the agent actually recalls — this is the mex GROW-discipline failure mode in miniature.
- **Interaction** with `PROJECT_SAFETY_NET` and the sticky-set hysteresis (`STICKY_RELEVANCE_BONUS`).

## Validation

Run against the eval harness (`recall` + `cost` dimensions): does `lore + overflow-ToC` improve recall accuracy on overflow-resident facts vs `lore`, and at what token cost? Ship only if the eval shows a net-positive accuracy/cost tradeoff.

## Notes

Filed per the mex competitive analysis (item 3). Our two-stage injection (frozen relevance-ranked baseline + durable knowledge delta + `recall`) already covers the bulk of mex's "routing" idea; this explores the one narrow gap — overflow-tail discoverability.

BYK:
Closes #917.

Knowledge entries that are relevance-scored but don't fit the injection budget are dropped silently (`forSession` Pass 2 `continue`), and the delta's "Additional Changed Knowledge" overflow only covers *changed* entries. So entries that exist but rank below the cut are **invisible** — the agent can't `recall` what it doesn't know exists. This surfaces them as a compact **recall-by-id ToC**, extending the existing entities partial-list pattern (`pipeline.ts` entities block) to knowledge.

Per discussion on #917 — shipping **A + B, no flag** (always on).

## A — turn-1 catalog floor (frozen system[1])

`buildKnowledgeCatalogText()` renders a confidence-ranked, capped (`STABLE_KNOWLEDGE_TOC_MAX=15`) project-knowledge catalog (`forProject(path, false)`, preferences excluded), folded into the frozen system[1] baseline next to the entities block. Present from turn 1 (before system[2]/any delta exists), byte-stable for the session's life, conservative visibility (project-owned only).

## B — relevance overflow (frozen knowledge delta)

- `forSession()` gains an opt-in `overflowSink?: KnowledgeEntry[]` option: it pushes the relevance-scored entries that didn't fit the budget (`allScored − result`), in deterministic score order. **Return type is unchanged** (`KnowledgeEntry[]`), so all ~33 existing call sites are untouched.
- The two system[2] call sites (normal + layer-4) capture the sink and thread it through `pendingKnowledgeDelta.overflow` → `appendKnowledgePromptDelta` → `buildKnowledgeDeltaMessage`, which renders `## Other relevant knowledge (recall by id for detail)` (`OVERFLOW_TOC_MAX=12`).
- **Cache-safe by construction:** B rides the *existing* durable-delta freeze cadence — a delta is only built on material change and replayed byte-identically thereafter (`applySessionPromptDeltas`, #747). The overflow is id-sorted (not relevance-order) and excludes ids already shown/superseded, so it adds **zero new cache-bust events**; the frozen delta is just slightly larger. The early-return guard is unchanged, so overflow **never** triggers a delta on its own.

## Why recall-by-id is actionable

The `recall` tool already accepts an `id` param (`k:`/`xk:` prefixes, `recall.ts:60`); recall.ts:422 uses the same "use recall with id parameter" pattern for truncated results.

## Tests

- **forSession overflow sink** (`ltm.test.ts`): selected/overflow disjoint, overflow ⊆ seeded, deterministic order across runs. (red→green proven)
- **buildKnowledgeDeltaMessage overflow** (`knowledge-delta-overflow.test.ts`, 6): heading + id/category render, id-sort byte-stability, cap + remainder, dedup vs changed/removed, **never-alone cadence** (overflow without material change → `null`), back-compat (no arg → no section). (red→green: 4 failed before render)
- **buildKnowledgeCatalogText** (4): render, empty→`""`, cap + remainder, caller-order preservation. **Mutation-checked** (forced `return ""` → 3 fail → restored byte-identical).
- Typecheck clean (core + gateway). Cache-stability e2e (14) and delta/prompt-delta regression (72) unchanged. Affected unit suite green 3× (no flake).

## Coverage gaps / follow-ups (called out for the reviewer)

1. **No end-to-end test of the wiring.** Both A→system[1] and B→delta wiring are low-logic data-passing, unit-covered at both ends + typechecked + no e2e regression. A dedicated e2e for B needs overflow forcing, which the gateway test harness can't express today (`budget` option exposes only `maxLayer0Tokens`, not `budget.ltm`). Recommend adding a harness `budget.ltm` knob + a B-overflow e2e as a fast-follow.
2. **Value eval deferred to #916.** Whether the agent *actually recalls* a ToC'd title is an empirical question requiring a **live LLM** run (real API spend) and a seeded N≫budget scenario. This belongs with the planned from-scratch eval rerun (#916), not this correctness PR. If the agent ignores the ToC there, we revisit.

## Adversarial review

Requesting a READ-ONLY adversarial correctness review before merge (per project policy), with particular scrutiny on the B wiring across the 4 `pendingKnowledgeDelta` sites and the cache-stability invariants.

### DIFF
--- a/packages/core/src/ltm.ts
+++ b/packages/core/src/ltm.ts
@@ -1235,6 +1235,18 @@ export type ForSessionOptions = {
    * wins, but ties and minor fluctuations don't reshuffle the selected set.
    */
   stickyIds?: Set<string>;
+  /**
+   * Optional sink for the budget-overflow tail (#917). When provided,
+   * `forSession` pushes the entries that were relevance-scored but did NOT fit
+   * the token budget into this array, in the same score-descending order it
+   * uses internally. Lets a caller surface a compact "these also exist — recall
+   * by id for detail" table of contents without re-querying. Invariants:
+   * selected (returned) entries are never included; lat.md synthetics are never
+   * included (they are not knowledge rows and carry no recall id). Ordering is
+   * deterministic (matches the internal `allScored` sort) so the rendered ToC is
+   * byte-stable across turns.
+   */
+  overflowSink?: KnowledgeEntry[];
 };
 
 /**
@@ -1638,6 +1650,20 @@ export async function forSession(
     recordSessionInjections(sessionID, projectPath, result);
   } catch (err) {
     log.warn("forSession: reinforcement failed (non-fatal):", err);
+  }
+
+  // --- 9. Surface the budget-overflow tail (#917) ---
+  // Entries that were relevance-scored (`allScored`) but didn't make the
+  // budget cut. `allScored` is already sorted score-desc (with id tiebreak), so
+  // pushing in iteration order yields deterministic, byte-stable ordering for
+  // the caller's recall-on-demand ToC. `allScored` contains only knowledge rows
+  // (lat.md synthetics are packed separately in step 6), so this never leaks a
+  // lat.md row, which has no recall id.
+  if (options?.overflowSink) {
+    const selectedIds = new Set(result.map((e) => e.id));
+    for (const { entry } of allScored) {
+      if (!selectedIds.has(entry.id)) options.overflowSink.push(entry);
+    }
   }
 
   return result;
--- a/packages/core/test/ltm.test.ts
+++ b/packages/core/test/ltm.test.ts
@@ -571,6 +571,53 @@ describe("ltm.forSession", () => {
     expect(result.length).toBeGreaterThan(0);
   });
 
+  test("overflowSink collects scored entries that did not fit the budget (#917)", async () => {
+    // 10 entries (~100 tokens each); a budget that fits only a couple. The rest
+    // are relevance-scored but over-budget — they must surface in overflowSink
+    // so the caller can advertise them for recall-on-demand without re-querying.
+    const ids: string[] = [];
+    for (let i = 0; i < 10; i++) {
+      ids.push(
+        ltm.create({
+          projectPath: PROJ,
+          category: "pattern",
+          title: `Overflow pattern ${i}`,
+          content: "C ".repeat(150),
+          scope: "project",
+          crossProject: false,
+        }),
+      );
+    }
+
+    const overflow: ltm.KnowledgeEntry[] = [];
+    const selected = await ltm.forSession(PROJ, SESSION, 250, {
+      overflowSink: overflow,
+    });
+
+    // Some fit, some didn't.
+    expect(selected.length).toBeGreaterThan(0);
+    expect(selected.length).toBeLessThan(10);
+    expect(overflow.length).toBeGreaterThan(0);
+
+    // Selected and overflow are disjoint, and every overflow entry is a seeded
+    // entry (never a phantom).
+    const selectedIds = new Set(selected.map((e) => e.id));
+    for (const e of overflow) {
+      expect(selectedIds.has(e.id)).toBe(false);
+      expect(ids).toContain(e.id);
+    }
+
+    // No id appears in both lists.
+    const union = new Set([...selectedIds, ...overflow.map((e) => e.id)]);
+    expect(union.size).toBe(selectedIds.size + overflow.length);
+
+    // Overflow is ranked (score desc), same order forSession uses internally —
+    // pin determinism so the rendered ToC is byte-stable across turns.
+    const rerun: ltm.KnowledgeEntry[] = [];
+    await ltm.forSession(PROJ, SESSION, 250, { overflowSink: rerun });
+    expect(rerun.map((e) => e.id)).toEqual(overflow.map((e) => e.id));
+  });
+
   test("includes relevant cross-project entries when session context matches", async () => {
     // Create a cross-project entry about TypeScript
     ltm.create({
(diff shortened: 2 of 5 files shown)


BYK:
## Adversarial review outcome + fixes

Ran a READ-ONLY adversarial correctness review (per policy). The cache-safety machinery passed all five invariants (id-sorted overflow → byte-stable; overflow never creates a delta alone; A frozen once/session; dedup vs changed/superseded; correct scoping across all 4 `pendingKnowledgeDelta` sites). Three mutation tests confirmed the new tests are non-vacuous.

### BLOCKER (fixed in `1846e5bd`)
**B1 — "recall by id" was unfulfillable.** The ToC rendered an 8-char id slice (`[019ef3fa]`), but `recallById` (recall.ts:1248) resolves `k:`/`xk:` by **exact** `ltm.get(id)` / `getByLogical(logicalIdOf(id))` — no prefix matching. Empirically: `recall(id="k:019ef3fa")` → `No entry found`. The agent would follow a literally-broken instruction.

Fix: render the **full** id with a recall-ready prefix → `[k:<uuid>]`. `k:` and `xk:` resolve identically (both hit `ltm.get`), so `k:` is safe for project-owned and promoted rows alike. Added a **recall round-trip regression test** (seed real entry → render ToC → extract token → assert `recallById` resolves it); mutation-verified that reverting to a slice fails it. (It degraded gracefully before — title was rendered in full, so recall-by-title would still work and nothing busted — but the shipped instruction was wrong.)

### Dispositions
- **S1 (A duplicates system[2] content):** accepted/documented. A is frozen at turn 1 before system[2] exists, so its selection can't be predicted; redundancy is bounded by `STABLE_KNOWLEDGE_TOC_MAX=15`. Cross-project leak check **passed** (`includeCross=false` → project-owned only).
- **S2 (no e2e):** the most important gap (id resolvability) is now closed by the round-trip tests. A full pipeline-wire e2e still wants a harness `budget.ltm` knob to force overflow — fast-follow.
- **N1 (empty-selection paths omit overflow):** intentional — degenerate "nothing selected" turns; overflow is a relevance tail and there's no selection to tail.
- **N2 (`forProject` ORDER BY lacks id tiebreak):** harmless — A's byte-stability rests on the once-per-session freeze, not query determinism.
- **N3 (short-id ambiguity):** resolved by switching to full ids.

### Value eval
Whether the agent actually *uses* the ToC is empirical and needs a **live LLM** run + N≫budget seeding — folding into the planned from-scratch eval rerun (#916), not this correctness PR.

BYK:
## Follow-up: full pipeline-wire e2e added (closes review S2)

Added `budget-overflow knowledge surfaces as a recall-by-id ToC in system[1] (A) and the delta (B)` to `cache-stability.e2e.test.ts`. It drives a real session through the gateway and asserts both halves on the upstream wire:

- **A**: the frozen system[1] catalog renders full recall-ready `[k:<uuid>]` ids, summarizes the remainder as "15 more", and is byte-identical across turns.
- **B**: after a material change, the overflow tail rides the durable knowledge delta (asserted against the persisted `session_prompt_deltas` row) with a full recall-ready id.

Mechanics worth recording (discovered while writing it):
- Entries must stay **under the 1200-char cap** — the pipeline runs `pruneOversized(1200)` every turn, which zeroes oversized confidence so they drop below the `confidence > 0.2` floor and never reach the ToC.
- The overflow tail only competes for the 8000-token budget floor when `forSession` keeps the set **relevant** (shared topic keyword); under sparse context it filters to a handful that all fit (no overflow).
- The diff-pin delta path engages via **idle-resume recompute** (near-zero `idleResumeMinutes` + a small real delay), not a forced layer — Layer 4's emergency budget drops the large set into a pure-removal delta.

Stable across 3 consecutive runs.

## #929 compatibility verified
`origin/main` advanced with #929 (confidence → `knowledge_meta` register), which overlaps my `ltm.ts` change. GitHub reports MERGEABLE; I also tested the merge locally: core+gateway typecheck clean, 93 unit/round-trip tests pass, and the new overflow e2e passes against the merged tree. No rebase needed.

Issue #931, opened 2026-06-23
test: flaky EADDRINUSE in createHarness random-port allocation


BYK:
## Symptom

Rare (~1 in 15-20 full-suite runs) failure:

```
FAIL  packages/gateway/test/pipeline-entity-injection.test.ts > ... > injects current-project entities into system[1] on turn 1
Error: listen EADDRINUSE: address already in use 127.0.0.1:46210
```

The failing test varies — it's whichever harness happens to lose the port race. Observed during the #928 work across ~19 full-suite + 12 gateway-suite runs (1 full failure, 1 gateway failure).

## Root cause

`createHarness` in `packages/gateway/test/helpers/harness.ts` picks a port with:

```ts
const port = opts.configOverrides?.port ?? 20000 + Math.floor(Math.random() * 30000);
process.env.LORE_LISTEN_PORT = String(port);
```

With many harness starts per run (and `pool: "forks"`), two harnesses occasionally draw the same random port, or a just-stopped server hasn't released its port before the next `startServer` binds — yielding `EADDRINUSE`. This is pure test-infra flakiness; it is independent of the code under test.

## Suggested fix

Either:
- Bind to port `0` (OS-assigned ephemeral port) and read the actual bound port back from the server, or
- Wrap `startServer` in a small retry loop that re-draws a random port on `EADDRINUSE` (e.g. up to 5 attempts).

Port 0 is the more robust option (no collision possible) but requires `startServer`/config to surface the bound port to the harness.

## Notes

Not a blocker for #928 (whose new test does not start a server). Filing so the harness flake is tracked rather than re-discovered.

PR #932, opened 2026-06-23
test(gateway): guard pipeline bustCause threading (closes #928)

### LINKED ISSUE #928: test: guard pipeline→recordCacheUsage bustCause threading (PR #926 follow-up)
## Context

Follow-up from the adversarial review of #926 (Finding #2, Mutation D).

PR #926 exempts `prefix-rewrite` cache busts from the consecutive-bust counter. The two halves of the contract are unit-tested:
- `categorizeBust` → `"prefix-rewrite"` for messages[0/1] divergence (`cache-analytics.test.ts`)
- `recordCacheUsage` exempts `"prefix-rewrite"` from `consecutiveBusts` (`gradient.test.ts`, mutation-verified)

The **idle.ts** call site (`effectiveMetaThreshold(..., getLastTurnAt(sessionID))`) is now guarded by two tests in `idle-work.test.ts` (mutation-verified — passing a constant `0` makes the "recent turn defers meta" test fail).

## Gap

The **pipeline.ts** call site (`recordCacheUsage(..., turnWasIdleResume, bustCause)` at ~pipeline.ts:4264) is NOT directly guarded. A mutation that drops the `bustCause` argument (reverting to the legacy `undefined` "count it" default) ships green.

## Why it wasn't fixed in #926

Two faithful approaches are blocked by tooling:
1. **Spy via `vi.mock("@loreai/core")`** — the gateway vitest config aliases `@loreai/core` to `packages/core/src`, so the barrel mock does not intercept the pipeline's aliased import (spy never fires).
2. **e2e via the simulated cache** — the harness computes cache read/write from byte-prefix stability, so it cannot produce the `cacheRead===0` full-bust that `categorizeBust` requires to classify a turn as `prefix-rewrite` (that needs Anthropic's real breakpoint semantics).

## Suggested fix

Either:
- Extract the `recordCacheUsage` call into a tiny testable seam in pipeline.ts and unit-test it, or
- Add a fixture-replay e2e that injects a response with `cache_read_input_tokens=0` and a request body whose distilled prefix (messages[0/1]) diverged, then assert `getConsecutiveBusts(sessionID)` does not increment.

Residual risk is low: both endpoints of the wire are tested and the `bustCause` value is also consumed by 3 adjacent telemetry calls in the same block, so a refactor dropping it solely from `recordCacheUsage` is unlikely.

BYK:
Closes #928 (follow-up to #926, adversarial-review Mutation D).

## Problem

PR #926 exempts `prefix-rewrite` cache busts (caused by Lore's own meta-distillation rewriting the distilled prefix) from the consecutive-bust counter. The two halves of that contract were unit-tested:

- `categorizeBust` → `"prefix-rewrite"` for messages[0/1] divergence (`cache-analytics.test.ts`)
- `recordCacheUsage` exempts `"prefix-rewrite"` from `consecutiveBusts` (`gradient.test.ts`, mutation-verified)

…but the **wire** connecting them inside `postResponse` — `analyzeCacheTurn → categorizeBust → recordCacheUsage(..., bustCause)` — was only reachable through the full request path. A mutation dropping the `bustCause` argument (reverting to the legacy "count it" default) shipped green.

Two faithful approaches were blocked in #926: a `vi.mock("@loreai/core")` spy doesn't intercept the gateway's source-aliased import, and the simulated-cache harness can't produce the `cacheRead===0` full-bust that `categorizeBust` needs to emit `prefix-rewrite`.

## Fix

**Extract a testable seam.** The cache-analytics + bust-tracking block moves out of `postResponse` into an exported `recordCacheTurnUsage()`. Behavior-preserving — the only reordering is that `recordCacheUsage` now runs just before `genAiSpan.end()` instead of just after (it never touched the span).

**Unit-test the whole wire directly.** `pipeline-cache-turn.test.ts` drives the seam with real request bodies + usage so `categorizeBust` runs for real (no mocking). Each test asserts BOTH the returned cause (scenario is set up correctly) AND the `consecutiveBusts` side effect:

- prefix-rewrite (messages[0] divergence, cacheRead=0) → **not counted**
- window-shift (messages[2] divergence) → **counted** (control)
- prefix-rewrite **holds** a genuine prior bust run (neither advances nor resets)
- no-request-body path → undefined cause, legacy count
- rolling cold-cache-window + `lastTurnWasIdle` consumption

**Mutation-verified:** dropping `bustCause` from the seam's `recordCacheUsage` call makes the prefix-rewrite tests fail (counter increments to 1).

## Tests

- 5 new tests; full suite green (4009 passed).
- typecheck + lint clean.
- A pre-existing harness `EADDRINUSE` port-collision flake surfaced during 3× suite runs — unrelated to this change (the new test starts no server) — tracked in #931.


BYK:
## Adversarial review complete — findings addressed

Ran a dedicated read-only adversarial review with mutation testing in an isolated worktree. Verdict: **SHIP-WITH-FOLLOWUPS**. The extraction was confirmed faithful/behavior-preserving statement-by-statement, and the headline bust-cause wire is genuinely guarded (mutations M1 `drop bustCause` and M2 `hardcode window-shift` both killed exactly the two prefix-rewrite tests).

Addressed:

**SHOULD-FIX — vacuous `lastTurnWasIdle` assertion.** The cold-window test seeded `lastTurnWasIdle: false` and asserted `false`, so it passed even with the consumption line deleted (mutation M3 survived). Fixed: `seedSession` now takes a `lastTurnWasIdle` arg; the cold-window test seeds `true` and asserts it's consumed to `false`. Re-verified: removing the consumption line now fails the test (M3 killed).

**NIT-1 — span-end reorder.** Rather than leave `recordCacheUsage` running after `genAiSpan.end()` (a marginal new exception-path exposure), I moved `genAiSpan.end()` *into* the seam right before its `recordCacheUsage` call. The extraction is now **ordering-identical** to the original inlined block — no reorder at all.

**NIT-2 — `inspectSessionState` returns `null`, not `undefined`.** Changed `toBeDefined()` → `not.toBeNull()`.

**NIT-3 — shared `BUST_USAGE` not runtime-frozen.** Now `Object.freeze`d.

Full gateway suite green (2193). Full suite green (4009); a pre-existing harness `EADDRINUSE` port-collision flake (tracked in #931, unrelated — the new test starts no server) appeared in ~1/8 full runs.

BYK:
Addressed the Seer finding on `.lore.md`: corrected the auto-generated knowledge entry that described the intermediate ordering. The final code (after review NIT-1) ends `genAiSpan` inside the seam BEFORE its `recordCacheUsage` call — ordering-identical to the original inlined block, so there's no span-leak reorder. (Also rebased onto current main to resolve a `.lore.md` merge conflict that was silently blocking the GitHub Actions `pull_request` workflow from dispatching.)

Issue #885, opened 2026-06-21
Harness isolation: background writes leak across harnesses via shared db() singleton


BYK:
Surfaced by the adversarial review of PR #879/#884 (#859 flake).

## Defect
The core `db()` connection is a process-global singleton (db.ts ~1371). Gateway test harnesses share it: each `createHarness` sets `LORE_DB_PATH`, calls `closeDB()` + `resetPipelineState()`, and reopens. But reset/teardown does **not drain in-flight background work** before swapping DBs:

- `resetBackgroundLimiter()` (background-limiter.ts:367) only `clearQueue()`s — its own docstring notes in-flight tasks continue to completion.
- Urgent/incremental distillation (pipeline.ts ~4448, ~4489) are untracked fire-and-forget `.catch()` promises — never awaited by `resetPipelineState`.
- `stopIdleScheduler` (idle.ts:466) just `clearInterval`s — it doesn't await an in-flight async tick.

Any of these, if mid-flight when the singleton is repointed to the next harness's DB, calls `saveSessionTracking(priorSid, …)` whose `INSERT OR IGNORE` (db.ts:2613) creates a **phantom header-NULL `session_state` row** in the *new* harness's DB. This is the ~1/3639 flake source behind #859 (an unscoped `SELECT … FROM session_state` count read 2 instead of 1).

## Mitigated for now
#884 made `session-rotation-merge.test.ts` deterministic by scoping its count assertions to `header_name`, so phantom rows are ignored. But the underlying leak can flake **other** exact-count assertions on `session_state` (or any table a background task writes).

## Proposed fix
Make reset/teardown quiesce in-flight background work before `closeDB()`:
- Track fire-and-forget distillation promises in a module-level set; `resetPipelineState` awaits `Promise.allSettled([...])` before returning.
- `resetBackgroundLimiter` awaits in-flight tasks (not just `clearQueue`).
- `stopIdleScheduler` returns an async stopper that `clearInterval`s then awaits the current tick.
- Harness teardown/setup awaits full quiescence before `closeDB()`.

## Confirmation (optional but ideal)
Add a diagnostic dump (`SELECT session_id, header_name, header_session_id FROM session_state`) on the next exact-count assertion failure + a high-iteration CI run to empirically confirm the phantom-row mechanism.

PR #937, opened 2026-06-23
test(gateway): bind test servers to port 0 to kill EADDRINUSE flake (closes #931)

### LINKED ISSUE #931: test: flaky EADDRINUSE in createHarness random-port allocation

BYK:
Closes #931.

## Problem

The test harness and three standalone server tests chose a listen port with `20000 + Math.floor(Math.random() * 30000)`. With many harness starts per run (and `pool: "forks"`), two servers occasionally drew the same port — or a just-stopped server hadn't released its port before the next bind — surfacing as intermittent:

```
Error: listen EADDRINUSE: address already in use 127.0.0.1:46210
```

The failing test varied (whichever harness lost the race); observed ~1 in 9 gateway-suite runs during the #928 work.

## Fix

Bind to **port 0** instead. The OS hands out a guaranteed-free ephemeral port, and `startServer` already resolves the actual bound port (`s.address().port`) and returns it as `server.port` — which every call site already uses for `baseURL`. This makes a collision *impossible* rather than merely unlikely. `server.test.ts` already used this exact pattern.

Fixed sites:
- `packages/gateway/test/helpers/harness.ts` (the main one — backs most gateway tests)
- `packages/gateway/test/api.test.ts`
- `packages/gateway/test/recall-openai-stream.test.ts`
- `packages/gateway/test/bedrock-routing.test.ts`

Tests that explicitly pass `configOverrides.port` still get their fixed port (unchanged).

## Verification

- Full **gateway suite ran 9× with zero EADDRINUSE** (2206 passed each run).
- typecheck + lint clean.
- `parsePort("0", …)` returns 0; `startServer` binds 0 and resolves the OS-assigned port (lines 579–584 of server.ts).


BYK:
## Adversarial review complete — verdict MERGE

Ran a dedicated read-only adversarial review (behavior-preservation audit + completeness sweep + 5× stress).

- **Behavior-preserving (all 4 files):** the pre-bind port value is used *only* to set `LORE_LISTEN_PORT`; every `baseURL` consumes `server.port` (the resolved port). The explicit-port override (`configOverrides.port`) is preserved verbatim.
- **Port-0 resolution verified end-to-end in source:** `parsePort("0")`→0; `startServer` doesn't reset 0 to default; binds; `resolvedPort = s.address().port`; returned as `server.port`. Default `hosts=["127.0.0.1"]` (single loopback) so the multi-host path isn't exercised. `server.test.ts`/`fetch.test.ts`/`start-gateway-reuse.test.ts` already use port 0 (precedent).
- **Completeness:** swept all test dirs — no remaining random-port/`LORE_LISTEN_PORT` binding sites. Hardcoded `3207`/`8000` matches are non-binding string fixtures.
- **Stress:** 5/5 gateway-suite runs clean (2206 passed each), zero EADDRINUSE, and no occurrence of the unrelated #885 phantom-row flake in-sample.

NITs (no action needed): env-leak of `LORE_LISTEN_PORT="0"` is strictly safer than the old leaked random port; `portExplicit` was already truthy pre-PR; `baseURL` hardcoding `127.0.0.1` is pre-existing and correct for the single-loopback default.

Issue #911, opened 2026-06-22
feat: validate cited code symbols in knowledge entries (reference-validity follow-up to #627)


BYK:
## Summary

Follow-up to the **Phase 0 reference-validity validator** in #627 (see [Phase 0 comment](https://github.com/BYK/loreai/issues/627#issuecomment-4770130455)). Phase 0 validates the cheap, low-false-positive reference types inside knowledge-entry content — `file:line` citations and `pnpm/npm/yarn/make` command refs. This issue tracks the third, higher-value-but-riskier reference type: **cited code symbols** (function / const / class / type names) still resolving in the repo.

## Motivation

Many entries cite identifiers, not just `file:line` — e.g. "`evaluateCacheStrategy` called only from `cache-warmer.ts:1252`", "`shouldHoldPrefixWarm` is a pure helper". When a symbol is renamed or removed, the `file:line` may still resolve (file exists, line in bounds) while the *claim* is now wrong. Symbol resolution catches this class of drift that Phase 0's path/line check and #627 Phase 3's file-level `git diff` both miss.

## Why it's a follow-up, not part of Phase 0

- **Parsing cost / accuracy:** reliably extracting "this token is a cited symbol" from prose and confirming it exists requires more than a regex (grep is cheap but noisy; AST/tags is accurate but heavier). Higher false-positive risk than `fs.exists`.
- **Same plumbing:** it slots into the Phase 0 `ReferenceResolver` abstraction (Direct-FS / Synthetic-probe / No-op) and the same clock-non-resetting soft-decay penalty — so it's strictly additive once Phase 0 lands.

## Approach (sketch)

- Extract candidate symbols from entry content (backtick-quoted identifiers adjacent to a cited `file:line`, to bound the search).
- Resolve via the active `ReferenceResolver`:
  - **Direct-FS:** `git grep -n <symbol>` / ripgrep, or a tags index, scoped to the cited file first, then the repo.
  - **Synthetic-probe (remote gateway):** batch into the client-side probe channel (`synthetic-tools.ts`), same as Phase 0.
- **🔴 Inherit the Phase 0 invariant: "cannot verify" ≠ "broken."** Ambiguous/unverifiable → strict no-op. Only a definitive "symbol absent from the cited file (and repo)" decays confidence, by one capped `−REFERENCE_DRIFT_PENALTY` step, never delete.
- Guard against false positives: comments/strings, renamed-but-aliased symbols, symbols defined in dependencies, generic words used as identifiers.

[...]

PR #939, opened 2026-06-23
feat: reference-validity validator (#627 Phase 0)

### LINKED ISSUE #627: feat: Add file citations to knowledge entries
## Summary

Link knowledge entries (and recall results) back to the **code and commits** that produced them. This gives Lore two things at once:

1. **Provenance / "why is this line here"** — answer `lore why <file>:<line>` by mapping a line → the session/distillation/knowledge that produced it (competitive parity with Entire's `blame`/`why` and GitHub Copilot Memory's citations).
2. **Staleness detection** — when referenced code changes, **decay confidence** instead of hard-deleting (better than Copilot Memory, which deletes the memory outright).

> **Key revision (see "Approach" below):** the originally-proposed *structured per-edit capture* (Step 1 — persisting every file path touched by every tool call) turned out to be **expensive** — a lot of metadata to collect, store, and keep in sync per tool invocation. We now prefer a **cheaper git-based approach** that leans on data we *already* collect (and currently throw away) plus query-time `git`. Structured capture is kept only as an optional precision upgrade, not the foundation.

## Motivation

Copilot Memory ties facts to code citations and validates them against the current branch HEAD; when citations don't validate, the memory is **deleted**. That's too aggressive — "we use the repository pattern" shouldn't die because a file was refactored. We can do better with **confidence decay, not deletion**: the principle persists at lower priority, and the curator naturally restores confidence for entries that stay useful.

Separately, the same file↔knowledge↔commit links let us answer "why does this line exist?" — surfacing the prompt and decision behind a change.

## What We Already Have (and what's discarded)

| Data | Location | Persisted? |
|---|---|---|
| **`git rev-parse HEAD` of the client** | Collected by the project probe: `RESOLVE_PROJECT_SCRIPT` at `synthetic-tools.ts:365-370`, parsed + validated into `result.gitHead` at `synthetic-tools.ts:552-567` | **No — parsed, validated, then thrown away.** Pipeline only consumes `result.root` + `result.gitRemote` (`pipeline.ts:1232-1241`); `gitHead` is never read. |
| **Commit SHA at commit boundaries** | `containsGitCommit()` already matches commit output `[main abc1234] …` (`pipeline.ts:363-368`) to trigger curation — it *sees* the SHA but only returns a boolean | **No** — SHA is matched then discarded |
| `knowledge.metadata` column | `db.ts` — JSON TEXT column | Column exists but **always NULL** (`ltm.create()` never writes it) |
| `distillations.source_ids` | JSON array of `temporal_messages.id` — written `distillation.ts:717`, surfaced in recall `recall.ts:507-517` | Yes — the existing distillation→message back-link |
| `knowledge.source_session` | `ltm.ts:174` | Yes — knowledge→session back-link (session_id only) |
| Tool input JSON with file paths | `LoreToolPart.state.input` at runtime | **No** — `tool_calls` stores only name/status/error/duration (`db.ts:770-786`) |
| `_extractFilePath()` parser | `gradient.ts:964-990` (handles `{"path"}`, `{"filePath"}`, `{"file"}`) | N/A (utility) |
| `PATH_RE` path regex | `gradient.ts:893` | N/A (utility) |

The existing **coarse provenance chain** today is:
`knowledge.source_session → session_id → distillations (by session_id) → source_ids → temporal_messages.content`.
What's missing is a **commit anchor** and a **file→entry** link.

### Phase 1 — Persist the commit anchor (cheapest, do first)

The whole feature hinges on knowing *which commit a session was at*. We already collect it and throw it away.

- **Persist `gitHead`:** consume `result.gitHead` in the pipeline where `root`/`gitRemote` are already handled (`pipeline.ts:1232-1241`) and store it on `session_state` (keyed by `session_id`; `project_path` already lives there per `db.ts:1029-1030`), or in a small new `session_commits(session_id, commit, observed_at)` table for multiple observations over a session's life. Use `MIGRATIONS.length` for the schema version (known gotcha — never hardcode).
- **Capture commit-boundary SHAs:** `containsGitCommit()` (`pipeline.ts:349-372`) already regex-matches `[main abc1234] …`. Have it (or a sibling) also *return/record* the SHA, writing a `session_commits` row at the moment of commit. This ties a session to the exact commit(s) it produced — no per-edit bookkeeping.

This alone enables Phase 3 (staleness) and a coarse `lore why` (commit → session → distillations/knowledge), with **near-zero new capture cost**.

### Phase 2 — `lore why <file>:<line>` query surface (cheap, git-at-query-time)

No new stored per-file metadata; resolve provenance lazily:

- Add `gitBlameLine(repoPath, file, line)` to `packages/core/src/git.ts`, mirroring the existing `execSync` pattern (`git.ts:98-148`); run `git blame -L <line>,<line> --porcelain` → commit SHA + author + time.
- Map that SHA → session via the `session_commits`/`session_state` rows from Phase 1; if no exact match (line predates Lore, or commit not recorded), fall back to **time-window correlation** (sessions active around the commit's authored time).
- Walk session_id → distillations (by session_id) → `source_ids` → temporal messages, and → knowledge (`source_session`); render the prompt/decision behind the line.
- New `packages/gateway/src/cli/why-cmd.ts` modeled on `recall-cmd.ts` (remote+local split, `--project`, `--json`); register in `main.ts` — **add flags to the global `OPTIONS` map** (`main.ts:90-129`; undeclared flags silently become booleans under `strict:false` — known gotcha), add `case "why"`, add to `knownCommands`. Reuse `runRecall()`/`recallById()` (`recall.ts:606`, `:1236`) to resolve `d:`/`t:`/`k:` IDs to content.

### Phase 3 — Staleness detection via git diff (medium effort)

- On session start, `git diff --name-only <last recorded session commit>..HEAD` (the "last commit" now exists thanks to Phase 1).
- For knowledge entries that have associated files (see Phase 4), if any referenced file changed: **decay confidence by ~0.1–0.2** (not delete) and surface a "may be stale" indicator in recall.
- Distinguish deleted / modified / renamed and handle each appropriately. The curator naturally restores confidence for entries that stay useful despite code churn.

### Phase 4 — File↔knowledge association (do the cheap variant first)

Two ways to populate "which files does this entry cite", cheapest first:

- **(Preferred, cheap) Query-time / blame-derived:** for staleness we mainly need the *commit* anchor (Phase 1) — the changed-file set comes from the `git diff` in Phase 3, not from stored per-edit metadata. For "files this entry is about", enrich the **curator prompt** with a "files touched in this conversation" block (derived from `git diff` over the session's commit range, or from file paths already present unstructured in `temporal_messages.content` / distillation observations), following the `buildToolFailureContext()` pattern at `curator.ts:817-838`. Store the result in the existing `knowledge.metadata` column as `{"source_files": [...]}` (or a `knowledge_file_refs` join table following the `knowledge_refs` / `knowledge_entity_refs` patterns) for reverse lookups.
- **(Optional precision upgrade — previously "Step 1", now de-scoped) Structured per-edit capture:** add `input_path TEXT` to `tool_calls` and populate it in `recordToolCalls()` (`temporal.ts:148`) for file-op tools using `_extractFilePath()` (`gradient.ts:964-990`). This yields exact, query-cheap file→session→knowledge joins **but is the expensive path** (per-tool-call metadata at scale). Pursue only if blame/diff-derived association proves too imprecise.

## Why This Is Better Than Copilot's Approach

- Copilot validates against HEAD and **deletes** memory when a cited file was refactored — even if the principle still holds.
- Ours: **confidence decay, not deletion**; the knowledge persists at lower priority and recovers naturally via the curator.
- We distinguish file deleted / modified / renamed and handle each appropriately.

## Suggested Sequencing

Phase 1 (persist commit anchor) → Phase 2 (`lore why`) and Phase 3 (staleness) can proceed in parallel on top of it → Phase 4 cheap variant → Phase 4 structured capture only if needed. Phases 1–3 deliver both the provenance query *and* staleness with essentially no new per-edit capture cost.

BYK:
## #627 Phase 0 — reference-validity validator (zero-token, commit-anchor-free)

Decays (never deletes directly) confidence on knowledge entries whose **literal references** (`file:line` citations, `pnpm run X`/`make X` commands) no longer resolve against the current repo. This is the cheap, commit-anchor-free half of #627 — *"does the reference still resolve?"* — complementary to the planned Phase 1/3 git-diff signal (*"did the cited file change?"*). Symbol resolution stays deferred to #911.

### Core (`@loreai/core`)
- **`references.ts`** — `extractReferences` (conservative: a file token needs a `:line` or a `/`, and its first path segment must not look like a DNS host, so prose like `e.g`/version numbers/bare URLs never match; **command refs are extracted only from backtick code spans** so prose like "make decisions"/"yarn about" is never a command; `pnpm install` & friends are excluded via a built-ins denylist). A shared **`RepoView` + `resolveRefAgainstView`** backs *both* `DirectFsResolver` (local FS) and `SyntheticProbeResolver` (remote snapshot) — the resolution logic is identical, verified by a real ref-by-ref parity test. `NoopResolver` = neutral.
- **`ltm.penalizeStaleReferences`** — mirrors `decayProject` on `knowledge_meta`; lowers confidence + bumps the sync clock but **never** resets `last_reinforced_at` (so reference-drift and disuse-decay compound). **`validateProjectReferences`** — rate-gated 24h (`projects.last_refcheck_at`), only `cross_project=0` LIVE entries, exactly **one flat capped penalty** per stale entry. Plus `peekProjectRefs` / `refValidity`.
- **Migration v56** — `projects.last_refcheck_at` + `knowledge_ref_validity(logical_id, broken, total, checked_at)`; added to `recoverMissingObjects` (both the table and the column-presence ALTER).
- **`config.knowledge.referenceValidation`** (default `true`).

### Gateway
- **Local (Direct-FS):** idle pass runs the validator when it can `stat` the repo (native plugin / `lore run`), *before* prune so a floored entry is reaped same pass.
- **Remote:** the reference snapshot **piggybacks the existing project-resolution SHELL probe** — one combined round-trip, **no extra injection**. (A separate probe short-circuited turns and broke the e2e suite; the piggyback only acts when the resolution probe already fires, so it's non-breaking.) Capture splits the combined output, binds the project, then applies penalties via `SyntheticProbeResolver`. The snapshot runs from the **repo root** (`cd $(git rev-parse --show-toplevel)`) so a subdirectory-launched agent doesn't mis-resolve root-relative refs.
- **`lore data show`** prints `References: N/M resolve`.

### 🔴 Load-bearing invariant
**"cannot verify" ≠ "broken".** A null resolver result, an empty/degenerate probe snapshot, or refs that resolve to `unknown` (absolute / out-of-tree path, ambiguous basename, missing `package.json`, unknown line count, truncated walk), **never** penalize. Shared knowledge (`cross_project=1`, including promoted entries) is never touched. Extraction biases hard for **precision over recall**: a missed reference (false negative) is acceptable; penalizing prose (false positive) is not.

### Adversarial review (round 2) — fixed in this PR
A second correctness review (per `quality/REVIEW.md` §6) found and this PR fixes two invariant breaches with trivial repros:
1. **Extraction false positives** — `make <word>`/`yarn <word>` prose and bare/relative URLs were extracted and penalized on common repos. Fixed by backtick-gating command extraction (removing the structurally-leaky stopword denylist) and rejecting host-like file paths.
2. **Remote probe** — `git ls-files` ran in the client CWD (subdir → mass false `missing`), and an empty file list mass-penalized instead of going neutral. Fixed by running the snapshot from the repo root and treating an empty file list as unverifiable (`null`).

Plus a security defense-in-depth charset assertion at the `__refset` interpolation site, a **real** Direct-FS↔SyntheticProbe parity test (the prior one only asserted membership in `{ok,missing,unknown}`), and a v56 recovery test that actually drops the column.

### Tests
- ~140 unit tests: extraction (incl. prose/URL non-extraction regressions), **Direct-FS ↔ SyntheticProbe ref-by-ref parity**, validator, combined-probe builder + splitter, probe CWD + empty-snapshot guards.
- **Mutation-verified guards**: cross_project (incl. the PR #902-style *promoted-entry* trap), `broken>=1` threshold, null-neutral, clock-not-reset, rate-gate, command backtick-gate, host-rejection, empty-snapshot-null, cd-toplevel, charset filter, column-recovery.
- Full core (1917) + gateway (2183) suites green; changed tests stable across 3 runs.

### Deferred follow-ups (documented, non-blocking)
- **Tracked-vs-on-disk divergence** (review CONCERN-4c): remote uses `git ls-files` (tracked) while local Direct-FS walks on-disk minus `WALK_IGNORE`, so a ref to a committed file under an ignored dir (e.g. Go `vendor/`) or an untracked-but-present file can resolve differently between modes. Low-frequency; a unified file source is a Phase-1 refinement.
- **Per-probe nonce markers** (security CONCERN): the section separator/markers are fixed strings; attacker-influenced file content could shift section boundaries (bounded to confidence nudging, no RCE). High-entropy per-probe nonces would close it.
- **`lineCount` file-size guard + client-side script-name extraction** (DoS/disclosure, low).
- **Rate-gate TOCTOU** (low): the validator `await`s between gate read and stamp.

Closes #627 (Phase 0). Follow-ups: #911 (symbol resolution), Phase 1/3 (git-diff change signal).


BYK:
**Post-review fixes applied:**

- **BLOCKER-1**: Added  denylist ("make sure", "make it", etc.) + regression tests. False  penalties on prose stopped.
- **SHOULD-FIX-1**: Removed blanket dot-dir skip in  — .github/ workflows, .vscode/ settings, etc. now resolve correctly in local mode. Relies on explicit  only (added .svn, .hg, .venv, .pytest_cache, .idea, .vim).
- **SHOULD-FIX-2**: Added  flag to  — when the walk hits  (60k), file-not-found becomes  (neutral) instead of .
- **NIT-1**: Added v56 recovery test (drop  + null , re-open, assert restored).

BYK:
Post-review fixes applied:

- **BLOCKER-1**: Added MAKE_PROSE_STOPWORDS denylist ("make sure", "make it", etc.) + regression tests. False `missing` penalties on prose stopped.
- **SHOULD-FIX-1**: Removed blanket dot-dir skip in `walk()` — .github/ workflows, .vscode/ settings, etc. now resolve correctly in local mode. Relies on explicit WALK_IGNORE only (added .svn, .hg, .venv, .pytest_cache, .idea, .vim).
- **SHOULD-FIX-2**: Added `truncated` flag to RepoView — when the walk hits WALK_FILE_CAP, file-not-found becomes "unknown" (neutral) instead of "missing".
- **NIT-1**: Added v56 recovery test (drop knowledge_ref_validity + null last_refcheck_at, re-open, assert restored).

BYK:
## Seer findings — disposition

**Fixed in c37b7ba5 (both HIGH):**
- **Windows path separators** (`references.ts`): `resolveRefAgainstView` now forces forward slashes after `path.normalize()`. The bug was real (the "resolved in 8caff59" tag was a false positive from line-shift) — on Windows `normalize("a/b")` → `a\b`, which never matched the forward-slash `view.files`/`basenames`. Verified by reasoning; not unit-tested because the conversion is a no-op on the POSIX CI runner (a Linux test would be vacuous).
- **Rate-gate retry loop** (`ltm.ts`): `stampGate()` moved out of the penalty `withTransaction` into a `finally`, so a rolled-back penalty write can't leave the gate unstamped and re-run the same failing pass every idle tick. Regression test added via DROP-TABLE failure injection; mutation-verified (reverting to stamp-inside-txn fails it).

**Reviewed, not changing (both MEDIUM):**
- **Remote probe consumes the shared 24h gate → "blocks local validator":** a given project is local **XOR** remote for a gateway — `idle.ts` only runs the Direct-FS validator when `existsSync(projectPath)`, and the synthetic probe only fires in remote mode. So a remote failure's gate stamp can never block a local check for the same project. Stamping on a neutral/failed probe is intentional (it rate-limits re-probing; the alternative re-probes every turn on persistent failure). Accepted trade-off: a transient remote failure delays that project's refcheck by ≤24h, with no wrong penalty.
- **Case-sensitive file resolution** (`DB.ts` vs `db.ts`): real but only on case-sensitive filesystems with a wrong-cased reference (refs are normally distilled from real paths, so case matches). A proper fix changes the `RepoView` shape (case-folded `files` set + `basenames` keys) across **both** the FS walk and the probe parser, and risks matching a ref to a different-cased sibling file. Deferring to a follow-up rather than expanding this PR's surface.

PR #941, opened 2026-06-23
feat(core): sync knowledge_meta as a convergent PN-counter register (A2 3b-2)

BYK:
## Summary

Makes knowledge `confidence` **converge across devices** instead of last-writer-wins, and wires the **local** sync plumbing for it. Part of the A2 immutable-knowledge epic (#823).

> Scope: this is the **local half**. The remote Supabase migration (RLS + quota for the two new tables) and its dedicated **security review** land in a stacked follow-up. Cross-device sync is not armed until then.

## Changes

**3b-2a — local CRDT mechanics**
- `confidence` becomes a per-`(logical_id, replica_id)` **PN-counter**: every lifecycle op (`create`/`reinforce`/`update`/`decay`/`outcome`/`prune`/`penalizeStaleReferences`) records a *signed delta* on **this device's own** grow-only `(pos, neg)` accumulators.
- Materialized `confidence = clamp(base_confidence + Σpos − Σneg, 0, 1)`, clamped **only at read**. `base_confidence` is immutable (set once at create).
- `v63` migration adds `base_confidence` + the `knowledge_meta_crdt` table.

**3b-2b — local sync wiring**
- `knowledge_meta` (immutable base, hash-LWW) and `knowledge_meta_crdt` (grow-only, `versioned:false`) become synced tables.
- Pull merge is **per-key MAX** over the counters (a join-semilattice) — applied unconditionally (idempotent/monotonic, never a conflict). Base is upserted **without** overwriting the local materialized value or decay clock.
- **Single-owner counters**: a device only ever pushes its *own* replica's rows; pulled peer rows arrive under apply-suppression and are never re-pushed.
- The `knowledge_meta` capture trigger is **base-gated** and the `knowledge_meta_crdt` UPDATE trigger is **growth-gated** (`pos>old.pos OR neg>old.neg`), so the frequent per-op re-materialization of `confidence` and no-op MAX-merges never churn the outbox.

## Review

Passed two adversarial correctness reviews (initial + final pre-merge), no blockers. Fixes folded in:
- **base_confidence immutability**: `insertMeta` ON CONFLICT no longer overwrites the immutable base on an explicit-id re-create.
- **`pruneOversized` hysteresis**: drives the **raw** (unclamped) confidence to 0, so a banked overshoot past the ceiling can't survive a prune. Now covered by a **mutation-verified discriminating test**.
- **`applyRemoteMetaCrdt`** uses the remote row's `updated_at`, not `Date.now()`.
- A pull-minted register row no longer pushes its **placeholder default base**. Regression test added.
- Dropped the non-existent `created_at` from `knowledge_meta` `syncColumns`.

**Rebase-integration fixes** (caught by Sentry after rebasing onto `main`, which had landed #627 reference-validity on the old confidence-on-`knowledge_meta` model):
- **`penalizeStaleReferences`** wrote `confidence` directly → would be **wiped by the next re-materialization**. Rewired through `applyConfidenceDelta` (durable + convergent), like `decayProject`. Mutation-verified regression test added.
- **Pull-mint decay clock**: a freshly-pulled register seeds `last_reinforced_at` to *now* (a local first-appearance touch), not `NULL` — otherwise `decayProject`'s `COALESCE(last_reinforced_at, k.updated_at)` fell back to the author's ancient content clock and made the entry instantly decay-eligible (which, now that decay is a CRDT counter, would sync back and spuriously lower it for every replica). Regression test added.
- **`pruneOversized` INNER JOIN** documented as deliberate (a LEFT JOIN would inflate the prune count with meta-less rows it can't actually zero).

## Verification

- typecheck 0 (all packages); Biome lint 0 errors
- full monorepo suite green after rebase onto `main` (v62): **253 files, 5143 passed / 32 skipped**
- `sync.property` stable across repeated runs; format clean

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1409,7 +1409,6 @@ const MIGRATIONS: string[] = [
   -- migrate() by KNOWLEDGE_META_MIGRATION_INDEX — the same pattern VACUUM uses.
   -- This string is intentionally a no-op marker so MIGRATIONS.length still counts.
   `,
-
   `
 
   -- Version 56: reference-validity validator (#627 Phase 0). Two additions:
@@ -1531,7 +1530,6 @@ const MIGRATIONS: string[] = [
   -- dropped/empty rollup) and reuse the canonical rebuild query in tests/recovery.
   -- This string is intentionally a no-op marker so MIGRATIONS.length still counts.
   `,
-
   `
   -- Version 61: persist the per-session cache-warmup COST bucket separately.
   -- session_state already stores warmup_savings / warmup_hits, but the warmup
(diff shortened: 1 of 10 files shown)


BYK:
Rebased onto `origin/main` (was 56 commits behind + conflicting → now MERGEABLE).

**Migration renumber:** `main` has since taken Version **57** (costs-page metadata index) and added **58/59/60**, so the CRDT register migration moved **57 → 61** — in both the forward `MIGRATIONS` array and the `recoverMissingObjects` self-heal block. Schema-version assertions updated to `61` (`db.test.ts` ×2, `ltm-versioning.test.ts`).

Conflicts resolved: `db.ts` (migration array + self-heal), `db.test.ts`, `ltm-versioning.test.ts`. `.lore.md` taken from `main` (auto-managed). `ltm.ts` / `sync-data.ts` / `sync.ts` auto-merged cleanly.

**Verification (local, rebased branch):**
- typecheck: clean (all 5 packages)
- full core suite: **2229 passed**, 6 skipped, 0 failures
- full gateway suite: **2287 passed**, 29 skipped, 0 failures
- biome on the 10 touched files: clean (the 31 `noNonNullAssertion` warnings are pre-existing in `synthetic-tools.test.ts`, untouched here)

**Follow-up filed:** the deferred remote half (Supabase migration 0009 — RLS + quota for `knowledge_meta` + `knowledge_meta_crdt` + dedicated security review) is now tracked in #1025. Cross-device confidence sync is not armed until that lands.

PR #951, opened 2026-06-24
feat(core): route bust-spiral detection to host hook for Sentry alerting (#797)

### LINKED ISSUE #797: fix(gradient): suppress the unsustainable-conversation warning during cold-start stabilization of a newly-tracked large session
## Problem
The "unsustainable conversation detected (N consecutive cache busts)" warning fires during the **cold-start stabilization** of a large resumed session — a false alarm that self-corrects within a few turns.

When Lore sees a large conversation for the first time in-process (`new=true`, `source=inferred`), the first few turns inherently rewrite the body (no warm cache exists, LTM gets injected, the layer settles). These count as consecutive cache busts and trip the warning, even though the session stabilizes immediately afterward.

## Production evidence
Session `ses_156668952ffeDIrhFHaU8Frjdq` (internal `12na8hXrpthm52Hl`, witty-garden, opus, 761 msgs), 2026-06-16:
- 22:02:33 — `WARN: unsustainable conversation detected (2 consecutive cache busts)`
- 22:03:42 — `WARN: unsustainable conversation detected (3 consecutive cache busts)`
- 22:04:39 — recovered to **99% hit** (busts 3→0)

The warning was prepended to the response (user-visible) but the session was healthy one turn later — the busts were turn-1 cold write + turn-2 LTM injection + turn-3 layer transition, all expected for a freshly-tracked large conversation.

## Proposed fix
Gate the unsustainable warning so it does not fire during the first N turns after a session is newly tracked in-process (`new=true` / freshly created gradient state), especially when the session starts already-large. Only surface the warning once the session has had a chance to reach steady state and is *still* busting (genuine unbounded growth).

## Notes
- Largely subsumed if #796 (inject LTM on turn 1 for already-large sessions) lands — that collapses the cold-start busts into a single unavoidable write, so the consecutive-bust counter wouldn't reach the warning threshold. Worth doing the warning-gate independently as defense-in-depth, since a genuinely huge single resumed turn can still cold-write more than once.
- The `SUSTAINED_BUST_THRESHOLD` repricing already exists; this is specifically about not alarming the user during a known-transient cold start.

BYK:
Fixes #797

## Problem
The "unsustainable conversation detected (N consecutive cache busts)" warning was injected into the response (user-visible) on any session whose gradient transform had `consecutiveBusts >= 2` past a cold-start grace. The warning is misleading (the user has no actionable response), and a sustained cache-bust spiral is almost always a real upstream bug (prefix drift, prompt-delta position drift, LTM pin mismatch, etc.) we want to investigate — not user-correctable behavior.

The original plan was to suppress the warning during the cold-start grace window. That hides the canary. The pivot: drop the user-visible text entirely and route the signal to Sentry where someone can act on it.

## Approach

- **Remove the user-visible warning** — `contextWarningText()`, the `unsustainable` field on `TransformResult`, the 3 unsustainable gates in `transformInner`, the cap-fit passthrough's `unsustainable: false` hardcode, the `injectContextWarning` call sites in `pipeline.ts`, and `context-warning.test.ts`.
- **Add a host-registered hook** (mirrors the existing `embedding.setEmbeddingFailureHook` pattern) — `setBustSpiralHook({ onColdStart?, onSpiral?, onRecovered? })` in `core/src/gradient.ts`. Core stays Sentry-free; gateway wires `setupBustSpiralCapture()` at startup.
- **Cold-start grace window** preserved as a *gating* signal — first 3 transforms of a freshly-tracked session: info-level Sentry breadcrumb (per episode, debounced). Past the grace: high-severity Sentry error (per spiral episode, debounced, cleared on recovery). Recovery: info-level breadcrumb.
- **Cap-fit passthrough preserved** — sub-cap sessions (layer 0 + tier 0) never alert. Their busts come from structural causes (e.g. prompt-delta position drift after post-idle recompression), not genuine context growth — the existing cap-fit hardcode `unsustainable: false` was the right design, just expressed differently in the new system.

## Why this is better than suppression

Suppression hides a canary. Two structurally different bugs (drift vs. genuinely unbounded growth) currently surface identically to the user; Sentry's structured context lets us distinguish them in aggregate. The issue's own production evidence (`ses_156668952ffeDIrhFHaU8Frjdq` recovering to 99% hit one turn after 3 consecutive busts) is exactly the kind of upstream bug we want to see and fix.

## What gets removed

- `packages/gateway/src/pipeline.ts`: `contextWarningText()`, the 7d block at 6740-6774, the `injectContextWarning` call sites at 7203/7252/7287/7338. `CONTEXT_WARNING_MARKER` + `injectContextWarning` + `stripContextWarnings` retained for the worker-degradation warning (still user-actionable).
- `packages/core/src/gradient.ts`: the `unsustainable?: boolean` field on `TransformResult`, the 3 gates in paths b/c/d, the cap-fit passthrough `unsustainable: false` hardcode.
- `packages/gateway/test/context-warning.test.ts`: deleted entirely.

## What gets added

- `packages/core/src/gradient.ts`: `COLD_START_GRACE_TURNS = 3`, `transformCount: number` field on `SessionState` (incremented at top of `transform()`, NOT restored from DB), `bustSpiralAlerted` / `bustSpiralColdStartLogged` debouncing flags, `BustSpiralHook` interface, `BustSpiralInfo` snapshot type, `setBustSpiralHook()` module-level setter, `maybeDetectBustSpiral()` detection function called from `transform()` post-inner.
- `packages/gateway/src/sentry.ts`: `setupBustSpiralCapture()` — wires the hook to Sentry (`onColdStart` → addBreadcrumb info, `onSpiral` → captureMessage error with fingerprint `["bust-spiral-past-grace"]`, `onRecovered` → addBreadcrumb info).
- `packages/gateway/src/server.ts`: calls `setupBustSpiralCapture()` at startup next to `setupEmbeddingFailureCapture()`.
- `packages/core/src/index.ts`: re-exports `setBustSpiralHook`, `BustSpiralInfo`, `BustSpiralHook`.
- `packages/gateway/test/sentry-bust-spiral.test.ts`: 3 tests for the Sentry wrapper.

## Tests

- **11 new bust-spiral tests** in `gradient.test.ts`:
  - Within-grace spiral fires `onColdStart` once, no `onSpiral` (debounced per episode across 2 transforms).
  - Past-grace spiral fires `onSpiral` (debounced — 3 turns at same level = 1 call).
  - Recovery (busts → 0) fires `onRecovered` and clears the alert flag.
  - Post-recovery spiral fires `onSpiral` again.
  - Past-grace with no hook registered: no crash, state still updates.
  - Grace boundary is inclusive: `transformCount <= 3` in grace, `>3` past it.
  - Coverage of compression-stage path (layer >= 1) and emergency layer-4 path.
  - `transformCount` increments exactly once per `transform()` call.
- **Updated "over-cap session that bypasses compression" test** — now asserts `onSpiral` fires past grace (not the removed `result.unsustainable` field).
- **Updated "tier-0 layer-0 passthrough never alerts" test** — asserts the hook does NOT fire for sub-cap sessions even past grace (preserves the original cap-fit hardcode behavior).
- **3 new gateway tests** for the Sentry wrapper: safe to call when Sentry not initialized, idempotent, hook shape verification.

## Verification

- Typecheck: clean across all 5 workspace packages.
- Lint: clean (the 1 pre-existing `data.ts:213` error is on `origin/main` and not in this diff).
- Full suite: 3963 passed, 32 skipped, 171 files.
- Flakiness: `gradient.test.ts` 3× consecutive runs all clean (201/201 each).
- **Mutation testing**: 6 mutations (M1–M6) all killed by the new tests, files restored byte-identical to baseline.
  - M1: coldStart guard removal → kills within-grace test (2nd transform re-fires).
  - M2: spiral guard removal → kills tier-0 test.
  - M3: `isCapFit` removal → kills tier-0 test.
  - M4: recovery branch removal → kills recovery test.
  - M5: `!bustSpiralAlerted` debounce removal → kills debounced test.
  - M6: `!bustSpiralColdStartLogged` debounce removal → kills within-grace test.

## Open follow-ups

- **Plan file** at `.opencode/plans/1797000000000-bust-spiral-sentry.md` documents the pivot rationale and the final design for future maintainers.
- **Rename candidate**: the `CONTEXT_WARNING_MARKER` / `injectContextWarning` / `stripContextWarnings` names are slightly inaccurate now (they're only used for the worker-degradation warning). Kept as-is to minimize churn; a future cleanup pass could rename to `RESPONSE_WARNING_MARKER` / `injectResponseWarning` / `stripResponseWarnings`.


BYK:
## Adversarial review complete — verdict SHIP-WITH-FOLLOWUPS

Dedicated read-only adversarial review ran 8 mutations on the new detection logic; 7 were killed by load-bearing tests, 1 survived (covered below). All checks green on CI.

**Findings:**

- **SHOULD-FIX (Finding #1) — `isCapFit` is structurally looser than the original cap-fit hardcode.** The new check `layer === 0 && getTier(totalTokens) === 0` collapses both path a (genuine cap-fit) and path b (tier-gate bypass) into the exception. A tier-gate bypass session with large LTM cache (over l0cap on LTM side, under l0cap on message-only side) would be silently suppressed. Filed as [#952](https://github.com/BYK/loreai/issues/952) with full repro and proposed fix (`capFit: boolean` on `TransformResult`).
- **NIT-A — Sentry wiring test is vacuous.** `packages/gateway/test/sentry-bust-spiral.test.ts` registers a hook with empty no-op functions and asserts vacuous conditions. Add a load-bearing test mocking `@sentry/bun`.
- **NIT-B — `BustSpiralInfo.cacheWriteTokens` / `cacheReadTokens` declared but never populated.** Either populate from real sources or remove the dead fields.
- **NIT-C — `bustSpiralColdStartLogged` clearing on recovery has no test coverage (M6 mutation survived).** Add an assertion in the post-recovery test that the flag resets to `false`.

All three NITs and the SHOULD-FIX are folded into [#952](https://github.com/BYK/loreai/issues/952) per the standing directive ("file deferred sub-PR reasoning as GitHub issues — never let it get lost").

**Mutation testing (8 mutations, 7 killed):**
- M1 `isCapFit = false` → tier-0 test FAILS ✓
- M2 Remove `bustSpiralHook?.` optional chain → no-hook test FAILS ✓
- M3 `busts === 0` → `false` → recovery + post-recovery tests FAIL ✓
- M4 Remove `!state.bustSpiralAlerted` debounce → debounced test FAILS ✓
- M5 Remove `state.bustSpiralAlerted = true` → 5 debounce-dependent tests FAIL ✓
- M6 Remove `bustSpiralColdStartLogged = false` in recovery → 0 tests fail (covered by NIT-C)
- M7 `inGrace` `<=` → `<` → boundary test FAILS ✓
- M8 Remove `transformCount++` → 3 tests FAIL ✓

**Other verification:**
- Typecheck clean across all 5 workspace packages
- Lint clean (the 1 pre-existing `data.ts:213` error is on `origin/main` and not in this PR's diff)
- Full suite: 3963 passed, 32 skipped, 171 files
- 3× flakiness check on `gradient.test.ts`: all clean (201/201 each run)
- 6 original muts + 2 follow-up muts from the reviewer all clean

**Stale-reference sweep:**
- `grep -rn "result\.unsustainable" packages/` → 0 hits
- `grep -rn "contextWarningText" packages/` → 0 hits in source
- `grep -rn "inColdStartGrace" packages/` → 0 hits (helper removed)
- All remaining "unsustainable" mentions in `pipeline.ts` are comments explaining the removal

Merging per SHIP-WITH-FOLLOWUPS verdict. [#952](https://github.com/BYK/loreai/issues/952) tracks the deferred work.

BYK:
## Missed Seer comment — follow-up filed in #952

Caught a Seer inline review comment at [gradient.ts:1289](https://github.com/BYK/loreai/pull/951#discussion_r4785028920) that I did NOT address before merge:

> **Bug:** The `bustSpiralColdStartLogged` flag is cleared on recovery, contradicting documentation and allowing multiple `onColdStart` events to fire within the grace period.

This is a real bug — the JSDoc says "never cleared within a process run" but the code clears on recovery. The original design intent (per the plan) was "once per cold-start episode" = "once per process run". Implementation diverged.

My own adversarial review caught the related **NIT-C** (no test coverage for the flag-clearing behavior) but did not surface the underlying doc/code contradiction. **Process gap on my side:** the review explicitly listed NIT-C as a NIT, but a doc/code contradiction is a SHOULD-FIX, not a NIT. I should have escalated the severity when I saw the missing test coverage on a flag whose declared contract was "never cleared".

Updated [#952](https://github.com/BYK/loreai/issues/952) with the full finding and the proposed fix. The NIT-C is now re-classified as a SHOULD-FIX, giving #952 two SHOULD-FIXes + two NITs in scope. Net follow-up work: ~3 small edits + 1 regression test in the bust-spiral alerting describe block.

PR #958, opened 2026-06-24
fix(gateway): make durable knowledge-delta append-only (cache-stable by construction)

BYK:
## Problem

#954 narrowed the durable knowledge-delta trigger to *genuine DB mutations*, but it still wrote **one coalesced `seq=0` row IN PLACE at a frozen deep `insertAt`**, and the surfaced baseline **never advanced**. So a session whose pinned set has a **persistent** mutation re-detected the same change every turn and re-rewrote the same deep message every turn.

Production logs after the #954 deploy (session `1LYkXZ7jkiHHnqPl`, Opus, Layer 4):

```
read=40973 (pinned)  write=262k–272k  cause=incremental  divergence="messages[407]"   — EVERY turn
prompt-delta: upserted knowledge update (4 changed, 66 removed, insertAt=407)
  prev: "...* 116 more — use recall with an id for detail."
  curr: "...* 115 more — use recall with an id for detail."   ← #917 overflow ToC churn
```

`66 removed` because 66 pinned entries are genuinely gone from the DB and the surfaced baseline never advanced past them → the trigger fires legitimately every turn; the volatile #917 overflow ToC then makes the row's *bytes* differ each turn even when the mutation set is stable → in-place rewrite at deep position 407 → busts the whole post-system segment. ~$25–30 / 20 min across affected sessions.

## Fix — append-only, cache-stable by construction

`appendKnowledgePromptDelta`:

1. **APPEND a fresh immutable block at the current tail** (`appendSessionPromptDelta`, `seq=MAX+1`) instead of upserting one row in place. Extending the tail never invalidates the cached prefix; an appended block is byte-immutable forever.
2. **ADVANCE the surfaced set per block.** Each block stashes its mutation signature `mut:{changed:[{id,h}],removed:[id]}` in its **selector JSON** (`parseMessageInsertSelector` ignores unknown fields). `advanceSurfacedKeys(pinBaseline, blocks)` reconstructs `pin ∪ every block's increment` **purely from the persisted blocks** — durable across restart, **no new column/state**. Once a mutation is surfaced it leaves the surfaced set, so a persistent removal fires **exactly once**, not every turn.

`reanchorExistingDelta` now re-anchors **all** blocks (not just `seq 0`) to one fresh tail index on reshuffle, preserving content + `mut` + chronological order (`applySessionPromptDeltas` sorts insertAt-DESC then seq-DESC). #917 overflow still only rides a real-mutation block and is then frozen in it (no per-turn overflow churn).

### What this means for the contract
The delta is **no longer one coalesced row** — supersessions live **across the sequence of immutable blocks**. Block count is bounded by *genuine mutations since the last compaction* (rare); reshuffles coalesce.

## Tests (red → green, mutation-checked)

- **`prompt-delta-append-only.test.ts`** (new): distinct mutations → multiple blocks (not one upsert); **persistent removal → fires once** (the 1LYkXZ fix); block **immutability** across later appends; re-edit same entry → new block.
- **`cache-stability.e2e.test.ts`**: the two `"ONE coalesced row"` / `"cumulative single row"` assertions are **rewritten** to the append-only invariant — per-block immutability, monotonic `insertAt`, and **all supersessions preserved across the block sequence incl. across a restart** (the B1 guarantee, now sequence-based).
- Mutation: disabling `advanceSurfacedKeys` re-fires persistent mutations → kills the no-re-fire tests.
- Full gateway suite (**2179**) green ×3; `tsc` (gateway+core) exit 0; biome clean.

## Scope / non-goals
- Builds on #954 (already merged). The 4 `pendingKnowledgeDelta` orchestration sites keep their selection-based **attempt** gate; `detectSurfacedMutations` over the advancing surfaced set remains the real filter (genuine edits/deletes change the selection keys, so the gate fires — same as #954, not a regression).
- Mechanism C "re-pin system[2] to current selection on compaction" is intentionally **not** included — the pin stays frozen for the session and blocks coalesce on reshuffle (compaction-with-churn deletes blocks then re-appends one cumulative-from-pin block). Lower blast radius; supersession correctness preserved.

Design doc: `.opencode/plans/1782293059815-delta-append-only.md`.


BYK:
### Adversarial correctness review — addressed in `171958ff`

An independent adversarial review (verdict: **no blocker**, design sound + cache-stable) verified the core invariants — hash-basis consistency across version bumps, removal detection, no supersession loss on coalesce, multi-block replay ordering + tool-pair safety, old-format `seq=0` self-heal. It surfaced one real bug + two gaps, now fixed:

- **[HIGH] lat.md synthetics falsely "Superseded"** (pre-existing from #954, but this PR froze it into an immutable block). lat.md entries (`category:"lat.md"`, id `file#Heading`, stored in `lat_sections` not `knowledge`) resolve null in `detectSurfacedMutations` → were classified as removed. Fix: only mark a null-resolving id removed when `ltm.isTombstoned(logicalIdOf(id))`. New red→green test.
- **[MEDIUM] Unbounded block growth** in a pathological no-idle churning session. Fix: `MAX_DELTA_BLOCKS=8` cap → next append clears + re-derives one cumulative block from the frozen pin. New deterministic test.
- **[LOW/MED] Reanchor test gap**: new test asserts multi-block reanchor keeps order + `mut` (mutation-checked: dropping `mut` re-fires and fails).

Deferred (self-healing / cosmetic, tracked): reanchor delete+re-append atomicity (`withTransaction`); changed-vs-overflow id-space duplicate; warm-cache selection-gate latency.

Full gateway suite (2182) green; typecheck + biome clean.

PR #980, opened 2026-06-24
perf(core): cover costs-page token-sum & recent-session aggregates (v58)

BYK:
## Problem

The `/ui/costs` page was still slow after v57. v57 only indexed the `role='assistant'` metadata subquery (~360ms→42ms); the two **largest** aggregates still did a per-row heap lookup for the aggregated/projected column:

| Query | Before |
|---|---|
| `aggregateTokensBySessionAll` token-sum | `SUM(tokens) GROUP BY session_id` streamed off `idx_temporal_session` (session_id only) but read `tokens` from the heap for **every** scanned row (~200K) |
| `listAllRecentSessions` | `COUNT/MIN/MAX(created_at) GROUP BY (project_id, session_id)` read `created_at` from the heap per row |

My own v57 comment flagged that these projections "aren't covered anyway" but never acted on it — that was the lever.

## Fix — migration v58: covering indexes

Widen both indexes to **cover** the columns each query reads, making them index-only scans (EXPLAIN flips `SCAN … USING INDEX` → `USING COVERING INDEX`, zero heap reads):

- `idx_temporal_session_created_tokens (session_id, created_at, tokens)`
- `idx_temporal_project_session_created (project_id, session_id, created_at)`

The wider indexes have the old narrow `idx_temporal_session` / `idx_temporal_project_session` as **exact left-prefixes**, so v58 **drops the narrow ones** — net index count on `temporal_messages` is unchanged, no extra write amplification on the hot table (same prefix-cleanup pattern as v6). Every former session-scoped access pattern (`WHERE session_id=? ORDER BY created_at`, `… AND distilled=0`) stays index-backed (verified empirically + pinned in tests).

## Measured (~200K messages)

| Query | Before | After | Speedup |
|---|---|---|---|
| token-sum | 101ms | 29ms | **3.5x** |
| full tokens fn | 113ms | 48ms | 2.4x |
| recent-sessions | 186ms | 79ms | **2.4x** |

The win grows as the table outgrows the page cache (real on-disk DBs).

## Tests

- New plan/correctness/**subsumption** pins in `cost-bulk-queries.test.ts` (drive verbatim production SQL).
- `recoverMissingObjects` self-heal block added (parity with sibling index blocks).
- **Red-phase verified**: neutralizing the v58 migration *and* its recover block fails exactly the 4 new v58 tests; restored byte-identical (checksum match).
- `pnpm test` full suite: **4221 passed / 0 failures**. Typecheck clean. Lint passes (warnings pre-existing, untouched files).

## Follow-up

Covering indexes are still O(N) index scans. Next-tier win is a **materialized per-session rollup** (maintain token-sum / message-count / first-assistant-metadata incrementally in `session_state`) so the page reads hundreds of rows instead of scanning ~200K. Filing as a separate issue.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1450,6 +1450,42 @@ const MIGRATIONS: string[] = [
   CREATE INDEX IF NOT EXISTS idx_temporal_role_session_created
     ON temporal_messages (role, session_id, created_at);
   `,
+
+  `
+  -- Version 58: COVER the remaining /ui/costs aggregates so they run index-only.
+  --
+  -- v57 (above) indexed the role='assistant' metadata subquery. The other two hot
+  -- aggregates still streamed their GROUP BY off a session-ordered index but did a
+  -- heap lookup for every scanned row to read the aggregated/projected column:
+  --   • aggregateTokensBySessionAll token-sum — SUM(tokens) ... WHERE created_at>=?
+  --     GROUP BY session_id — grouped off idx_temporal_session(session_id) but read
+  --     the tokens column from the table for each of ~200K rows.
+  --   • listAllRecentSessions — COUNT(*) / MIN(created_at) / MAX(created_at)
+  --     GROUP BY (project_id, session_id) — grouped off idx_temporal_project_session
+  --     but read created_at from the table per row.
+  --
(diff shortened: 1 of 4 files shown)


BYK:
Follow-up for the materialized per-session rollup (next-tier optimization) filed as #981.

PR #987, opened 2026-06-25
feat(gateway): support Claude via Google Vertex AI (#870 part 2)

BYK:
feat(gateway): support Claude via Google Vertex AI (#870 part 2)

Adds **Google Vertex AI (Claude)** support to the gateway — part 2 of #870 (part 1, AWS Bedrock, already merged).

## How it works
Claude on Vertex is the native Anthropic Messages API with three differences ([docs](https://docs.claude.com/en/api/claude-on-vertex-ai)):
1. the model id is in the **URL path** (not the body);
2. `anthropic_version: "vertex-2023-10-16"` goes **in the body** (not a header);
3. streaming is selected by the URL **verb** (`:streamRawPredict` vs `:rawPredict`), not a body `stream` field.

Streaming is plain Anthropic SSE and the response is native Anthropic JSON, so both reuse the existing parsers. Auth is a **GCP OAuth2 bearer** via Application Default Credentials (lore holds the GCP creds and mints/refreshes tokens with `google-auth-library`; the client speaks plain Anthropic to lore).

Reachable via `X-Lore-Provider: vertex` / `google-vertex` / `google-vertex-anthropic`. Config uses the standard Google env vars (`GOOGLE_APPLICATION_CREDENTIALS`, `GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_REGION`/`LOCATION`); region defaults to the recommended **global** endpoint (no regional premium).

## What's included (all-in-one)
- **Conversation path** (`pipeline.ts` + the pure `buildVertexUpstream` helper): reuses `buildAnthropicRequest` (incl. `cache_control`), strips `model`/`stream`, injects `anthropic_version`, builds the `:rawPredict`/`:streamRawPredict` URL, swaps `x-api-key` → GCP bearer, and strips the anthropic-only headers (`x-api-key`, `anthropic-version`, `anthropic-beta`) that Vertex does not accept. 5m cache-TTL downgrade for the non-native host (mantle precedent). cch billing re-sign never fires (gated on the anthropic protocol).
- **Worker path** (`llm-adapter.ts`): a distinct `"vertex"` worker protocol + `buildVertexWorkerRequest` (Anthropic body → rawPredict URL + OAuth bearer) so background distillation/curation run on Vertex. `buildWorkerRequest` is now async. Caching is driven by a `cache_control` block (GA on Vertex), not the `anthropic-beta` header.
- **Cache warming** (`cache-warmer.ts`): an `authMode: "vertex"` discriminator + `buildVertexProfile`; `executeWarmup` branches to use the GCP bearer + per-model rawPredict URL, no cch. The warmup body is routed through `toVertexBody` so `stream` is never sent and no `anthropic-beta` is forwarded.
- **`vertex-auth.ts`** (ADC token mint/refresh + project resolution, with a test seam) and **`translate/vertex.ts`** (URL/body/model-id/host helpers; `global` and regional hosts normalized consistently).
- Tests: `vertex.test.ts` (helpers, routes, worker protocol, warmer profile + body, auth seam, header-strip + region-override regressions) and `vertex-routing.test.ts` (`X-Lore-Provider: vertex` E2E).

## New dependency
`google-auth-library` (gateway) — the standard GCP ADC token library.

## Testing
- Full suite: **4295 passed**, 32 skipped, 0 failures. Typecheck + Biome lint clean.
- No live GCP credentials yet, so the request path is exercised via an interceptor + a token test seam (same approach as the Bedrock PR). A live smoke test is the remaining validation once a GCP project with Claude enabled is available.
- Two adversarial read-only reviews were run. The pre-open pass caught a warmup-body `stream` re-injection bug and a project-threading gap (both fixed + regression-tested). The Seer review pass led to stripping `anthropic-beta` before forwarding to Vertex and honoring an explicit `X-Lore-Upstream-URL` regional override; both fixes ship with regression tests proven non-vacuous via mutation (mutants killed on apply-then-revert).

## Follow-ups (non-blocking)
- Live-validate against an entitled Vertex project: model-id `@`-date encoding in the URL and the `anthropic-beta` strip (confirm Vertex behaves as the docs imply once Claude access is granted).
- `X-Lore-Upstream-URL` for Vertex is **region-only**: the region is extracted from the override host and the URL is rebuilt against the Google endpoint. A non-Vertex override host (e.g. a corporate proxy) is therefore discarded rather than used as the base — unlike the anthropic/openai paths. Document or extend if proxying Vertex becomes a requirement.
- Model-id alias map (`toVertexModelId`) is verified against the published table but untested live; `-latest` ids are not mapped.

### DIFF
--- a/packages/gateway/package.json
+++ b/packages/gateway/package.json
@@ -28,6 +28,7 @@
   },
   "dependencies": {
     "@supabase/supabase-js": "^2.58.0",
+    "google-auth-library": "^10.9.0",
     "p-limit": "7",
     "qrcode-terminal": "^0.12.0",
     "semver": "^7.7.3",
--- a/packages/gateway/src/cache-warmer.ts
+++ b/packages/gateway/src/cache-warmer.ts
@@ -1897,7 +1897,9 @@ export async function executeWarmup(
   //    (Vertex is not a Claude Code billing endpoint). The session credential
   //    is irrelevant — Vertex auth is the GCP token lore mints.
   //  - Anthropic (default): the session credential + cch re-sign to /v1/messages.
-  const headers: Record<string, string> = { "content-type": "application/json" };
+  const headers: Record<string, string> = {
+    "content-type": "application/json",
+  };
   let requestUrl: string;
   let signedBody: string;
 
--- a/packages/gateway/src/cache-warmer.ts
+++ b/packages/gateway/src/cache-warmer.ts
@@ -50,6 +50,7 @@ import { resolveUpstreamRoute } from "./config";
 import { isBedrockMantleHost } from "./translate/bedrock";
 import {
   isVertexHost,
+  toVertexBody,
   toVertexModelId,
   vertexRawPredictUrl,
   vertexRegionFromUrl,
@@ -976,10 +977,26 @@ export function buildVertexProfile(
   upstreamBase: string,
 ): CacheWarmingProfile {
   const region = vertexRegionFromUrl(upstreamBase) ?? "global";
+  const base = buildAnthropicProfile(model, ttl);
   return {
-    ...buildAnthropicProfile(model, ttl),
+    ...base,
     upstreamUrl: `https://${region}-aiplatform.googleapis.com`,
     authMode: "vertex",
+    // prepareAnthropicWarmupBody re-adds `stream:false` and (for billing
+    // sessions) leaves the body shaped for Anthropic. Vertex rejects a body
+    // `stream` field (the URL verb selects streaming) and needs `model` absent
+    // + `anthropic_version` present — so run the prepared body back through the
+    // canonical Vertex transform. The stored body is already the Vertex body,
+    // so this only re-strips the `stream` that prepareWarmupBody re-injected.
+    prepareWarmupBody: (storedBody: string) =>
+      JSON.stringify(
+        toVertexBody(
+          JSON.parse(base.prepareWarmupBody(storedBody)) as Record<
+            string,
+            unknown
+          >,
+        ),
+      ),
   };
 }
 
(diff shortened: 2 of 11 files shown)


BYK:
## Seer review addressed (commit `c52f5686`)

Resolved the four findings. Note that **14872465 and 14881922/0 directly contradict each other** (one says *add* `anthropic-beta` on the worker path, the other says *strip* it on the conversation path). The official [Claude-on-Vertex docs](https://docs.claude.com/en/api/claude-on-vertex-ai) settle it: the documented Vertex `:rawPredict` header set is **exactly `Authorization` + `content-type`** (no `anthropic-beta`), and prompt caching is a supported Vertex feature driven by `cache_control` **body** blocks (GA), not a beta header.

| Finding | Severity | Resolution |
|---|---|---|
| 14870494/0 — `global-` host 404 | CRITICAL | Already fixed in `9dc3e402` via `vertexHost()` (bare `aiplatform.googleapis.com` for `global`). |
| 14881922/0 — `anthropic-beta` forwarded to Vertex | MEDIUM | **Fixed.** Now stripped alongside `x-api-key`/`anthropic-version` before forwarding. |
| 14881922/1 — `X-Lore-Upstream-URL` region ignored | MEDIUM | **Fixed.** Region is now parsed from `effectiveUpstreamBase` (honoring the override), falling back to `config.vertexRegion`. |
| 14872465/0 — worker omits `anthropic-beta` → caching fails | MEDIUM | **Rejected (won't fix), documented inline.** The worker enables caching via its `cache_control` block; adding the header would diverge from the documented Vertex contract and risk a 400. |

### Implementation
The conversation-path transport rewrite (region resolution + `toVertexBody` + header strip + bearer) was extracted into a pure, exported `buildVertexUpstream` helper so the header-strip and region logic is unit-testable in isolation. Added **7 regression tests**; both the `anthropic-beta` strip and the region-override behaviors were **mutation-verified** (each test confirmed to fail against the pre-fix behavior, then restored byte-identical).

Full gateway suite green (2249 passed), typecheck + Biome clean.

> Note: this PR is **held** pending live Vertex E2E — Anthropic-MaaS quota is currently blocked by Google (new-billing-account 48h gate). The fixes above are static-verified.

PR #988, opened 2026-06-25
feat(references): validate cited code symbols (#911)

### LINKED ISSUE #911: feat: validate cited code symbols in knowledge entries (reference-validity follow-up to #627)

BYK:
Closes #911. Follow-up to the #627 Phase 0 reference-validity validator. Phase 0 checks `file:line` and `pnpm/npm/yarn/make` command refs; this adds the third type: **cited code symbols** (function / const / class / type names). When a symbol is renamed/removed, the `file:line` often still resolves while the *claim* is wrong — symbol validation is the only check that catches that drift.

## Design — presence-history drift signal

A cited symbol only **decays** confidence when it was **previously confirmed present** in this repo and is **now absent** (genuine rename/removal). A symbol that was never present here (external lib, historical/renamed-away mention, rejected alternative) never gets a presence row, so it can never penalize. This is what keeps symbol validation on the safe side of the 🔴 "cannot verify ≠ broken" invariant — and makes the check a real drift detector rather than a false-positive generator.

- **`knowledge_symbol_presence(logical_id, symbol, last_present_at)`** (migration v59). `validateProjectReferences` records a symbol on a confirmed `ok`; a later `missing` is drift only if a prior row exists.
- **Extraction (precision-first):** `{kind:"symbol"}` only when (a) inside a backtick span, (b) the entry also has a file ref (adjacency bound), (c) the token is *codey* (PascalCase/camelCase/`_`/call-like).
- **Resolution (presence by `git grep -wF`, no AST):** present anywhere → `ok`; definitively searched-and-absent → `missing`; grep unavailable/errored → `unknown`. `RepoView.searchedSymbols` distinguishes "absent" from "couldn't search" so a grep error is never a false `missing`.
- **Direct-FS** greps from the **git toplevel** with a **scrubbed `GIT_*` env**. Non-git repo → neutral.
- **Synthetic-probe** emits per-symbol `name\t0|1` bounded by `OK`…`DONE` markers; a truncated probe → neutral (never a partial false-missing set). Subshell forces `exit 0`.

A symbol drift folds into `knowledge_ref_validity(broken,total)` and the same one-shot `−REFERENCE_DRIFT_PENALTY`, floored, never deleting.

## History
First cut resolved absent→`missing` directly; adversarial review found that false-penalizes valid external/historical/rejected mentions (🔴 violation). Reworked to presence-history; all 4 SHOULD-FIX hardenings (grep exit-128, `GIT_*` env, subdir/probe divergence, truncation) landed too. See the review-response comment for the point-by-point.

## Verification
- Core 2061 / 6 skipped; gateway 2218 / 26 skipped. Typecheck + biome clean.
- Mutation battery on both cuts: extraction gates, codey gate, resolution, both `-w` sites, probe `exit 0`, presence-gate, presence-record, searched-distinction, OK/DONE marker, toplevel-grep, env-scrub — all killed.

## Deferred (per #911)
AST/definition-shape precision, symbol hit-location ("moved" vs "removed"), per-kind observability split.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1486,6 +1486,24 @@ const MIGRATIONS: string[] = [
   DROP INDEX IF EXISTS idx_temporal_session;
   DROP INDEX IF EXISTS idx_temporal_project_session;
   `,
+
+  `
+  -- Version 59: knowledge_symbol_presence — per-entry record that a cited code
+  -- symbol was once confirmed present in the repo (#911). The symbol
+  -- reference-validity check only DECAYS confidence when a symbol that was
+  -- previously present goes absent (genuine rename/removal drift). An
+  -- external/historical/rejected-alternative mention that was never present here
+  -- never gets a row, so it can never be penalized — this is what keeps symbol
+  -- validation on the safe side of the "cannot verify ≠ broken" invariant.
+  -- Keyed by the stable logical_id (matches knowledge_ref_validity /
+  -- knowledge_session_injections / knowledge_meta) so it survives version edits.
+  CREATE TABLE IF NOT EXISTS knowledge_symbol_presence (
+    logical_id      TEXT NOT NULL,
+    symbol          TEXT NOT NULL,
+    last_present_at INTEGER NOT NULL DEFAULT 0,
+    PRIMARY KEY (logical_id, symbol)
+  );
+  `,
 ];
 
 // Index of the migration whose work is performed by a column-presence-aware JS
@@ -1956,6 +1974,12 @@ function recoverMissingObjects(database: Database) {
       broken     INTEGER NOT NULL DEFAULT 0,
       total      INTEGER NOT NULL DEFAULT 0,
       checked_at INTEGER NOT NULL DEFAULT 0
+    );
+    CREATE TABLE IF NOT EXISTS knowledge_symbol_presence (
+      logical_id      TEXT NOT NULL,
+      symbol          TEXT NOT NULL,
+      last_present_at INTEGER NOT NULL DEFAULT 0,
+      PRIMARY KEY (logical_id, symbol)
     );
     CREATE TABLE IF NOT EXISTS knowledge_transfers (
       knowledge_id           TEXT NOT NULL,
--- a/packages/core/src/ltm.ts
+++ b/packages/core/src/ltm.ts
@@ -1124,6 +1124,20 @@ export async function validateProjectReferences(
 
   let checked = 0;
   let penalized = 0;
+  // Symbol drift uses a presence HISTORY (#911): a symbol only counts as broken
+  // when it was previously confirmed present in this repo and is now absent (a
+  // genuine rename/removal). A symbol that was NEVER present here (external lib,
+  // historical/renamed-away mention, rejected alternative) never gets a row, so
+  // it can never penalize — this is what keeps symbol validation on the safe
+  // side of "cannot verify ≠ broken".
+  const recordSymbolPresent = db().query(
+    `INSERT INTO knowledge_symbol_presence (logical_id, symbol, last_present_at)
+       VALUES (?, ?, ?)
+     ON CONFLICT(logical_id, symbol) DO UPDATE SET last_present_at = excluded.last_present_at`,
+  );
+  const wasSymbolPresent = db().query(
+    `SELECT 1 FROM knowledge_symbol_presence WHERE logical_id = ? AND symbol = ?`,
+  );
   // Stamp the 24h gate in a `finally`, OUTSIDE the penalty transaction. If a
   // write inside the transaction throws, `withTransaction` rolls back the whole
   // batch — but the gate MUST still advance, otherwise the next idle tick re-runs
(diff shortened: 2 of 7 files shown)


BYK:
## Reworked after adversarial review — BLOCKER + all 4 SHOULD-FIX addressed

The first cut resolved an absent cited symbol to `missing` (penalty). The review correctly showed that false-penalizes a common, valid class — external libs (`useState`), renamed-away mentions, rejected alternatives — a direct violation of the 🔴 "cannot verify ≠ broken" invariant. I reproduced it independently. Fixed by switching to a **presence-history drift signal**:

**BLOCKER → fixed.** New `knowledge_symbol_presence(logical_id, symbol, last_present_at)` table (migration **v59**). A symbol is recorded on a confirmed `ok`; a later `missing` only counts as drift (penalty) when a prior presence row exists. A symbol that was **never** present here never gets a row → can never penalize. This kills the false-positive class *and* turns the check into a genuine rename/removal detector (its actual purpose). Regression test: `never-present mention (external/historical) absent → neutral`.

**SHOULD-FIX 1 (exit 128 ≠ absent).** `RepoView` now carries `searchedSymbols`; a symbol is `missing` only when definitively searched-and-absent (grep exit `1`). Any other exit (128/signal) leaves it unsearched → `unknown` (neutral). Test: `not in the searched set (grep errored) → unknown`.

**SHOULD-FIX 2 (`GIT_*` env).** `DirectFsResolver` spawns git with a scrubbed env (`GIT_DIR`/`GIT_INDEX_FILE`/`GIT_WORK_TREE`/…removed). Test: `inherited bogus GIT_DIR does not cause a false missing`.

**SHOULD-FIX 3 (subdir divergence).** Direct-FS now greps from the **git toplevel** (`rev-parse --show-toplevel`), matching the probe's `cd $(git rev-parse --show-toplevel)`. Test: `subdirectory root greps from the toplevel → sibling-package symbol → ok` (verified `git -C pkgA grep <pkgB-symbol>` exits 1, so it's non-vacuous).

**SHOULD-FIX 4 (truncation).** The probe emits per-symbol `name\t0|1` bounded by `OK`…`DONE`; a truncated probe (no `DONE`) → null/neutral, never a partial false-missing set. Test: `OK but NO DONE marker (truncated) → null`.

**NIT (file-stem token, e.g. `` `RepoView.ts:42` `` → symbol `RepoView`):** now harmless under presence-history (only decays if it was present and disappeared). Left as-is.

### Verification
- Full suites green: **core 2061 / 6 skipped, gateway 2218 / 26 skipped**; typecheck + biome clean (no new diagnostics).
- Mutation battery on the rework: **MA presence-gate, MB presence-record, MC searched-distinction, MD DONE-marker, MF toplevel-grep, MG env-scrub, MH probe-DONE — all KILLED.** (The `status===1`-vs-`else` exit-code branch is defense-in-depth not deterministically inducible by a unit test; the resolver-side guard it protects is killed by MC.)

PR #1002, opened 2026-06-26
fix(gateway): scale 5m warmup margin by prefix size to avoid TTL-race partials

BYK:
## Problem

A partial warmup was observed in production (session `1F4OWImUH0Nhn4x3`):
input=164,807 tokens, cacheRead=44,153, cacheWrite=120,652 → **27% hit, \$0.78 cost**, with `ageSinceCacheTouch=281s` against a 300s TTL (only 19s headroom).

A 164K prefix spans 3 cache blocks, each with its own effective TTL boundary. The flat **45s** warmup margin is insufficient for large multi-block prefixes: a big conversation block sitting against the 5-minute wall crosses its per-block boundary inside the headroom window, so the warmup lands as a partial (head block refreshed, large block rewritten).

## Fix

The headroom a prefix needs to clear its largest block's boundary is **absolute** (a function of prefix size), **not a fraction of the TTL** (a fraction would balloon the 1h margin to 24 min = thrash).

New pure function `computeWarmupMargin(ttlMs, prefixTokens)`:
- **5m TTL**: `45s + prefixTokens × 0.27ms`, capped at `+90s` → tops out at **135s** (cap bites past ~333K tokens), always well under the 300s TTL.
- **1h TTL**: stays **flat at 300s** (already dwarfs any realistic prefix's headroom).
- `prefixTokens=0` (callers without a live session) reproduces the original flat base — **byte-identical back-compat**.

`state.lastInputTokens` is threaded through `resolveProfile` at every Anthropic-cache warming path:
- live warming loop (`idle.ts`)
- dashboard snapshot (`computeWarmingSnapshot`)
- `buildVertexProfile` — Vertex Claude also uses Anthropic prompt caching, so it gets the same fix (it would otherwise stay on the flat margin).

## Tests

- 8 `computeWarmupMargin` unit tests: 5m base/back-compat, 1h flat (incl. huge prefix), linear scaling (literal-pinned constant), the 164K incident value, cap behavior, monotonicity, margin-never-reaches-TTL invariant, negative-guard.
- `buildAnthropicProfile` + `resolveProfile` + Vertex `resolveProfile` threading tests.
- End-to-end **WIRING** test through `computeWarmingSnapshot`: a 300K prefix opens the warmup window at idle=200s while a 60K prefix is still "too early" — proving `state.lastInputTokens` flows into the internally-built margin.

### Mutation testing
- **M1** (snapshot call site → `0`): killed by the WIRING test.
- **M2** (`MARGIN_PER_TOKEN_MS` 0.27→0.30): killed by 4 tests.
- **M3** (`MAX_SIZE_MARGIN_MS` cap 90k→100k): killed by the cap test.
- **M4** (Vertex branch drops `prefixTokens`): killed by the Vertex threading test.
- All baselines restored byte-identical (sha256-verified).

## Verification
- Full gateway suite: 2263 passed / 29 skipped.
- Typecheck clean (all 5 packages); Biome clean on changed files.
- New tests stable across 3 consecutive runs (no flakiness).

### DIFF
--- a/packages/gateway/src/cache-warmer.ts
+++ b/packages/gateway/src/cache-warmer.ts
@@ -170,6 +170,21 @@ export const MIN_INPUT_TOKENS_FOR_WARMING = 50_000;
  *  ensures at least 30% return probability before the first warmup. */
 export const MIN_RETURN_PROBABILITY_FLOOR = 0.3;
 
+/** Size-proportional warmup-margin scaling (5m TTL only). The headroom a prefix
+ *  needs to clear its largest block's per-block TTL boundary is ABSOLUTE (a
+ *  function of prefix size), not a fraction of the TTL — so the 45s base gets a
+ *  per-token term. ~0.27ms/token puts a 164K prefix at ~44s of added headroom
+ *  (→ ~89s total margin), matching observed warmup round-trip + boundary jitter
+ *  for large multi-block prefixes. See computeWarmupMargin. */
+export const MARGIN_PER_TOKEN_MS = 0.27;
+
+/** Absolute cap on the size-driven margin term. Bounds the margin with a fixed
+ *  tail instead of a fraction of the TTL: a fraction would balloon the 1h margin
+ *  to 24min (warm with 36min of life left = thrash), and even on 5m would warm
+ *  too early. With this cap the 5m margin tops out at 45s + 90s = 135s (the cap
+ *  bites only past ~333K-token prefixes). */
+export const MAX_SIZE_MARGIN_MS = 90_000;
+
 // ---------------------------------------------------------------------------
 // Per-bucket circuit breaker
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


BYK:
## Adversarial review resolution

Ran an independent adversarial correctness review (second reviewer). Verdict was **SHOULD-FIX, no blocker** — runtime logic correct, headline tests kill their mutants. Addressed all findings:

- **SHOULD-FIX 1 (live-warming path uncovered)** — `idle.ts` (the path that actually pays for warmups) threaded `state.lastInputTokens` but no test guarded it; mutating it left all tests green. Closed *structurally*: extracted the byte-identical `state → resolveProfile` call shared by `idle.ts` and the dashboard snapshot into one seam `resolveProfileForSession(state)`. Both now route through it; a unit test on the seam guards the threading for both (mutation now kills the seam test **and** the WIRING test).
- **SHOULD-FIX 2 (NaN not actually guarded)** — `Math.max(0, NaN) = NaN` would propagate to a NaN margin and silently bypass the cooldown/window gates (warm every tick). Added a real `Number.isFinite` guard + real NaN/Infinity test cases (the prior test name over-claimed "NaN-guard" with only negative cases).
- **NIT 3** — pinned the thin no-double-warm invariant: `45s + MAX_SIZE_MARGIN_MS < ttlMs/2` (a future cap bump past ~105K ms would otherwise silently enable two warmups per 5m window).
- **NIT 5** — added a bedrock-mantle margin-threading test for symmetry with first-party/vertex.

### Mutation testing (all killed, baselines restored byte-identical)
M1 snapshot threading · M2 `MARGIN_PER_TOKEN_MS` · M3 `MAX_SIZE_MARGIN_MS` cap · M4 vertex threading · M5 `resolveProfileForSession` seam (kills seam + WIRING) · M6 NaN guard.

Full gateway suite: **2267 passed / 29 skipped**. Typecheck clean (5 packages). Biome clean.

BYK:
## Second adversarial review — verdict: MERGE-WITH-NITS (zero blockers)

Independent adversarial review of the final pushed state. All 6 invariants verified with real numbers; 3 of 4 mutations killed.

**Invariant #1 (no-double-warm), computed not assumed:** worst-case 5m margin `45_000 + 90_000 = 135_000 < ttlMs/2 = 150_000` ✓. Cooldown `165_000 ≥ window-width 135_000` → two warmups per window impossible ✓. Max margin `< 300_000` → never starved ✓.

**Mutations killed:** NaN-guard removal · `MAX_SIZE_MARGIN_MS` cap bump (165000<150000 fails) · seam threading null-out (kills seam **and** WIRING). **Survived:** a *synthetic* injected line overriding `profile.warmupMarginMs` in the idle loop — flagged as SHOULD-FIX (live-loop margin had no direct test; structurally mitigated by the shared seam).

**Closed the SHOULD-FIX:** added `shouldWarm — size-scaled margin warms EARLIER in the live gate than a flat margin would`. Same `baseState()`, `intoWindow=200s`: the large-prefix profile (capped 135s margin → 165s threshold) **warms**, the flat-45s profile (255s threshold) **does not** — the only difference is the margin, proving it's load-bearing in the real warm/no-warm decision. Mutation-verified non-vacuous (reverting the per-token term → `expected 45000 to be 135000`).

Other findings were NITs only (magic-number `3_600_000` TTL discriminator in two spots; scaling keyed on `!== 1h` rather than an allow-list) — neither is a live bug since only `300_000`/`3_600_000` exist.

Full gateway suite: **2268 passed / 29 skipped**. Typecheck + Biome clean. Worktree clean, all mutation baselines restored byte-identical.

Issue #999, opened 2026-06-25
Investigate pathological awaited latency on embed/vector read path (p50 ~20s, p95 ~98–128s)


BYK:
## Summary

The measure-first read-path telemetry from #993 surfaced an unexpected, systemic latency signal that is **not** main-thread sync-blocking: the **awaited** portion of `forSession`/`recall` (embed + pool vector search) is pathologically high — ~20s p50 and ~98–128s p95 — across a healthy sample size. This is off-thread, so it doesn't block the event loop, but a 23s *median* suspension across 217 `forSession` calls is systemic, not tail outliers. It points at the embedding/vector worker subsystem being serialized/backlogged (and/or event-loop contention delaying continuation resumption).

This issue tracks investigating and fixing that latency. It is distinct from the original #966 scope-B question (which was about *main-thread sync-blocking*).

## Data (`byk/loreai-gateway`, `tracemetrics` dataset, since #993 deploy ~7d)

| op | n | sync_blocking p50 | sync_blocking p95 | awaited p50 | awaited p95 |
|---|---|---|---|---|---|
| `forSession` | 217 | 60 ms | 191 ms | ~23 s | ~128 s |
| `recall` | 6 | 3,540 ms | 29,129 ms | ~20 s | ~97 s |

`recall` sync_blocking by scope: `scope:all` p50 7,500 ms; `scope:knowledge` p50 1,925 ms.

Reproduce (note: the `-m`/`--json` metric resolver is flaky; the `-F "agg(value,<metric>,distribution,millisecond)"` form is reliable):

```
sentry explore byk/loreai-gateway -F op \
  -F "count(value,lore.readpath.total_ms,distribution,millisecond)" \
  -F "p50(value,lore.readpath.sync_blocking_ms,distribution,millisecond)" \
  -F "p95(value,lore.readpath.awaited_ms,distribution,millisecond)" \
  --dataset metrics --period 30d
```

## What's wrapped in `timer.await`

`forSession` (`packages/core/src/ltm.ts`) only awaits two things (lines ~1890–1894):

- `embedding.embed([sessionContext], "query")`
- `embedding.vectorSearch(contextVec, 50, excludeFilter)`

So the ~23s median awaited is dominated by the local ONNX embed of the session context + the pool vector search. With `forSession` sync-blocking at only 60 ms p50, the cost is almost entirely in the embed/vector subsystem, not in `forSession`'s own CPU.

## Hypotheses to test

[...]

PR #1051, opened 2026-06-30
feat(core): cut vector storage over to FLAT vec0 with partition keys

BYK:
## What

PR4 of the vec0 cutover. Switches the embedding store from per-row `embedding` BLOB columns to sqlite-vec **`vec0` virtual tables**, eliminating the O(n) brute-force cosine scan behind #999 (awaited p50 ~23s / p95 ~128s) and removing the `MAX_*_VECTOR_ROWS` recency-cap stopgaps.

Builds on PR #1030 (native sqlite-vec in the SEA binary), #1038 (vendored 0.1.10-alpha.4), and #1044 (the capability/read-mode seam). Plan: `.opencode/plans/1782503960000-diskann-ann-switchover.md`.

## Index choice — FLAT vec0, not DiskANN (empirically overturns the locked plan)

A spike against the vendored `v0.1.10-alpha.4` binary disqualified DiskANN for this workload:

| | DiskANN (int8) | **FLAT vec0 + partition keys** |
|---|---|---|
| `PARTITION KEY` / metadata filter pushdown | ❌ *"not supported with DiskANN"* | ✅ |
| insert cost | ~48 ms/vec → **~85 min** / 106K backfill | 0.12 ms/vec → **~13 s** |
| recall@10 | ~0.82 (approximate) | **1.0 (exact)** |
| session-scoped read (partition pushdown) | n/a (can't filter) | **~0.9 ms** |
| project-wide read (106K, off-thread) | sub-linear but can't scope project | **~308 ms** exact |

DiskANN in this build supports only a PK + the vector (no project/session filtering at all); `rescore` also rejects partition keys. **FLAT** is exact, supports partition pushdown, has cheap inserts, and still lets us drop the recency caps. The user confirmed this change before implementation.

## How

- **`db/vec-store.ts`** — vec0 DDL for the 4 tables at the configured dim (`temporal_vec` chunk-keyed `chunk_id`+`+message_id` ahead of multi-vector chunking); `storeEmbedding`/`deleteEmbeddings`/`clearAllEmbeddings` vec0 branches; `ensureVec0Store` (create + drop/recreate on a dimension change); `copyBlobsToVec0` + `dropEmbeddingColumn` cutover helpers; mode-aware `missing/hasEmbeddingSql`; `embeddingByIdSource`; `gcVec0DanglingRows`.
- **`vector-query.ts`** — a `vec0` KNN branch per helper: partition pushdown for temporal/allDistillations, CTE-KNN + JOIN post-filter + over-fetch for knowledge (confidence/category) and distillations (archived). Recency cap removed **on the vec0 path only** (blob paths keep it).
- **`embedding.ts`** — one-time blob→vec0 cutover (backfill via `INSERT…SELECT` → `DROP COLUMN` → flip `vec.storage_mode` last → `incremental_vacuum`), idempotent/resumable and column-presence-aware (the v55 boot-loop lesson); mode-aware backfill detection; dimension-change handling; **degraded short-circuit**; a resilient in-process read fallback; a startup GC sweep.
- **`db.ts`** — ensure the vec0 tables exist on open for an already-vec0 DB (reads work before the cutover runs).
- **`ltm.ts` / `entities.ts` / `pattern-echo.ts`** — read stored vectors by id from the vec0 table in vec0 mode (vec0 returns the same float32 bytes; `distillation_vec` even carries `session_id` as an aux column). `appendVersion` omits the dropped `embedding` column and demotes the old vec0 row so `knowledge_vec` holds only current versions.

## Degrade contract (preserved)

A vec0-store DB opened on a runtime that can't load sqlite-vec (Bun-on-macOS, Node without `allowExtension`, or `LORE_DISABLE_VEC=1`) → vector recall returns `[]` (FTS/keyword recall still answers), **never crashes**, and re-converges when next opened on a capable runtime. Storage stays net-neutral (blob columns dropped at cutover; no doubling).

## Tests

- New `packages/core/test/vec0-cutover.test.ts` (13 tests, real vec0-capable connection): write/read round-trip, vec0↔blob **exact parity**, end-to-end cutover (relocate + DROP COLUMN + flip + idempotent re-run), `appendVersion` after the column is dropped + demote, **recency-cap removal** (4010 rows; the oldest perfect-match still surfaces), dimension change, delete maintenance + GC, mode-aware predicates.
- Updated the obsolete PR3 placeholder (`vec0 → throws`) to the implemented behavior.
- Full core suite green (97 files / 2242 tests); typecheck + Biome clean. Key guards mutation-tested non-vacuous (confidence post-filter, knowledge demote-delete).

## Follow-ups (not blocking)

- Eager bulk-delete maintenance at project/session/prune sites (today dangling rows are harmless — recall hydration drops a hit whose base row is gone, partitions isolate deleted projects — and the startup GC sweep reclaims them).
- PR5: rollout smoke guards, telemetry (storage_mode / coverage / KNN p50-p95), kill-switch hardening.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1,5 +1,10 @@
 import { Database } from "#db/driver";
-import { loadVecExtension, resetVecState } from "./db/vec";
+import { isVecAvailable, loadVecExtension, resetVecState } from "./db/vec";
+import {
+  ensureVec0Store,
+  readStorageMode,
+  readVecDimension,
+} from "./db/vec-store";
 import { join, dirname } from "node:path";
 import { chmodSync, mkdirSync } from "node:fs";
 import { getGitRemote } from "./git";
@@ -2062,6 +2067,16 @@ export function db(): Database {
   // before the tracing Proxy wrap. Never throws — when unavailable, vector
   // search falls back to the pure-JS brute-force path (see db/vec.ts).
   loadVecExtension(database);
+  // For an already-vec0 DB on a capable runtime, ensure the vec0 tables exist
+  // BEFORE any read can reach them — the blob→vec0 cutover only runs later, in
+  // runStartupBackfill. Recovery uses the persisted dimension (set at cutover);
+  // a fresh DB stays blob until that cutover, and an incapable runtime (vec
+  // unavailable → reads degrade to []) skips this. vec0 tables are local,
+  // non-synced derived data, so creating them after installSyncCapture is safe.
+  if (isVecAvailable() && readStorageMode(database) === "vec0") {
+    const dim = readVecDimension(database);
+    if (dim !== null) ensureVec0Store(database, dim);
+  }
   // Wrap the connection in the query-tracing Proxy AFTER migrate() and sync
   // change-capture are installed on the RAW connection, so (a) migration and
   // TEMP-trigger setup queries are never traced / re-entrant, and (b) the
--- a/packages/core/src/db/vec-store.ts
+++ b/packages/core/src/db/vec-store.ts
@@ -51,6 +51,14 @@ export type EmbeddingTable =
 /** `kv_meta` key recording this DB file's {@link VecStorageMode}. */
 export const VEC_STORAGE_MODE_KEY = "vec.storage_mode";
 
+/**
+ * `kv_meta` key recording the vector dimension the `vec0` tables were created
+ * with. vec0 fixes the dimension at DDL time (`float[N]`), but the embedding
+ * dimension is configurable (768 local / 1024 voyage / 1536 openai). On a
+ * dimension change {@link ensureVec0Store} drops + recreates the tables.
+ */
+export const VEC_DIMENSION_KEY = "vec.dimension";
+
 /** Logical table → physical base table name. */
 const BASE_TABLE: Record<EmbeddingTable, string> = {
   knowledge: "knowledge",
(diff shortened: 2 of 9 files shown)


BYK:
Addressed a pre-merge adversarial review (commit `858f12dc`):

- **BLOCKER** — cutover crash-safety: flip `vec.storage_mode` to `vec0` **before** dropping the base `embedding` columns, with drops moved to a separate vec0-mode-only reclaim pass. This makes "mode=blob ⟹ columns intact" an invariant, so `checkConfigChange`'s blob branch can never read a dropped column (the persistent silent-recall-death / v55 boot-loop hazard). Reclaim is presence-aware + idempotent → resumable mid-drop.
- **SHOULD-FIX** — post-filtered vec0 reads now widen the KNN window when attrition leaves `< limit` survivors (blob parity), and every vec0 `k` is clamped to `SQLITE_VEC_VEC0_K_MAX` (4096) — exceeding it errors. This also fixed a latent crash for `limit`/over-fetch > 4096.
- **SHOULD-FIX** — added coverage for the by-id vector point-reads (ltm/entities/pattern-echo, incl. the `distillation_vec` `session_id` aux) in vec0 mode.

Tests: exported `maybeCutoverToVec0` and added real-orchestration coverage (full cutover, mid-reclaim crash resume, idempotency) + the widening + by-id cases. Full core suite green (2246); new guards mutation-verified non-vacuous.

PR #1152, opened 2026-07-02
feat(core): show cumulative temporal re-chunk progress in logs

BYK:
## Why
The only per-run progress log — `re-chunking temporal messages: N re-chunked, N scanned…` — is **per-process**: it resets to zero on every gateway restart. On the current multi-day, restart-spanning re-chunk walk that means the logs never show how far through the whole corpus we are. Answering "are we making progress?" required cross-referencing the startup backlog line across restarts and querying the DB by hand.

## What
The heartbeat and the final summary now report a **cumulative** percentage that keeps climbing across restarts:

```
temporal re-chunk: 13.5% complete (30688/226527 messages) · +18 re-chunked this run
```

And the startup line now shows the baseline too:

```
temporal re-chunk: 2 messages to scan (2/4 already done, 50%), resuming
```

## How (the math)
- At startup we already `COUNT` the remaining `backlog` (embeddable rows above the resume cursor). One extra `COUNT` gives the embeddable `total`, so `baseDone = total - backlog` is what prior runs already covered.
- The walk's `SELECT` is itself filtered to `length(content) >= 50`, so `scanned` counts embeddable rows, and **`baseDone + scanned` reaches exactly `total` on a clean pass → 100%**.
- Both `COUNT`s run once per process, and only when work remains (the vec0-mode / done-flag guards already return early otherwise).

Percentage formatting is extracted into a pure, exported `formatTemporalRechunkProgress(done, total, thisRun)` — clamps to 100, one decimal, `total === 0` ⇒ 100% (never NaN/Infinity) — so it's unit-testable without waiting on the 30s heartbeat.

No behavior change to the walk itself; logs only.

## Tests
- Unit tests for the formatter: one-decimal rounding, 100% clamp when `done > total`, `0/0` ⇒ 100%, fresh `0%`.
- Integration: startup baseline line + exactly `100% complete` on a clean completion.
- **Resuming** test (cursor pre-seeded to `m2` of `m1..m4`) that a fresh run can't cover — proves the metric is cumulative (`4/4`), not per-process (`2/4`).
- **Mutation-verified**: swapping `baseDone + scanned` for `processed` fails the resuming test. Full `packages/core` suite: **2495 passed, 0 failed**.

### DIFF
--- a/packages/core/src/embedding.ts
+++ b/packages/core/src/embedding.ts
@@ -2738,6 +2738,25 @@ export function resetTemporalRechunkProgress(): void {
 }
 
 /**
+ * Cumulative-progress line for the temporal re-chunk walk. `done`/`total` count
+ * embeddable (>=50 char) messages across ALL runs, so the percentage keeps
+ * climbing over restarts — unlike the per-process counters, which reset to zero
+ * on every boot. `thisRun` is what the current invocation re-chunked. A zero
+ * `total` reads as 100% (nothing to do). The percentage is clamped to 100 and
+ * rendered to one decimal. Exported for testing. See {@link
+ * backfillTemporalEmbeddings} for how `done = baseDone + scanned` is derived.
+ */
+export function formatTemporalRechunkProgress(
+  done: number,
+  total: number,
+  thisRun: number,
+): string {
+  const pct =
+    total > 0 ? Math.min(100, Math.round((done / total) * 1000) / 10) : 100;
+  return `temporal re-chunk: ${pct}% complete (${done}/${total} messages) · +${thisRun} re-chunked this run`;
+}
+
+/**
  * Re-chunk existing temporal messages into the multi-vector vec0 layout.
  *
  * New messages are embedded part-aware at write time, but messages written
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial pre-merge review: **MERGE** — all six correctness areas hold (predicate parity for backlog/total/walk, `baseDone + scanned == total` on a clean pass, overshoot/late-insert clamped, retryFrom rewind recomputes cleanly with no >100%, formatter NaN/clamp safety). Mutation-verified non-vacuous: swapping `baseDone + scanned`→`processed` fails the resuming test but not the fresh-run test, confirming only the resuming case distinguishes cumulative from per-process.

Took the one worthwhile NIT: hoisted the `total` COUNT inside `if (backlog > 0)` so the rare `backlog===0` path doesn't pay for a wasted full-table COUNT (safe — the loop latches done without iterating, so `total`/`baseDone` are never read there). Left the two cosmetic NITs ("complete" wording on a transient-failure pass; no integration test for overshoot) — low value for a logs-only change.

Issue #825, opened 2026-06-19
C — Encryption + key management foundation


BYK:
> **Part of epic #821.** Depends on: #822 · prerequisite for #826
> _This issue body mirrors the design doc `.opencode/plans/sync-C-encryption-keymgmt.md`._

---

# C — Encryption + key management foundation

Parent: `sync-sharing-epic.md`. Depends on: A1 (scope seam — scope keys key on
`scope_id`). Prerequisite for D (pro tables must be encrypted from day one).

## Goal
Client-side encryption of sensitive blobs so Supabase stores ciphertext only
(near-zero-knowledge), with a key model that extends from personal (1 user, N
devices) to team (selected members) without a redesign. We do NOT need
server-side search on these blobs, which is exactly what makes this viable.

## Crypto (epic decision #6) — libsodium/HPKE now, MLS later
- Per-scope symmetric **DEK** (XChaCha20-Poly1305) encrypts blob columns only
  (content / narrative / facts / observations). 🔴 ids, timestamps, scope_id,
  author_id, content_hash, cursor fields stay CLEARTEXT (sync mechanics need them).
- DEK wrapped per member via libsodium sealed box / HPKE (RFC 9180); stored in
  `scope_keys(scope_id, member_user_id, wrapped_dek, key_epoch)`.
- 🔴 Versioned envelope on every blob: `scheme_id` + `key_epoch` header, so MLS
  (RFC 9420) can later be an ADDITIVE scheme — old blobs stay readable, no
  re-encryption. Key-agreement sits behind an interface (`getScopeKey(scope)` /
  `wrapForMember`).
- 🔴 Document the libsodium→MLS rationale + threshold (org-scale / churn /
  forward-secrecy / PCS) INLINE at the seam AND in `docs/` architecture. Never
  hand-roll crypto.
- v1: personal only — one DEK per user, wrapped to the user's own device keys.
  Group (multi-member wrapping) is E.

## Key management (epic decision #7) — dual model
- Per-device **identity keypair** (X25519). Behind one interface:
  - **Escrow (recoverable):** root key wrapped by Argon2id(password) + optional
    recovery key, stored server-side (Bitwarden/BitLocker style).
  - **Client-only (zero-knowledge):** root key never leaves devices; recovery via
    another paired device or printed recovery code.
- **Add device:** QR / short-code PAKE (SPAKE2) pairing — existing device wraps
  the root key to the new device's public key (Signal-style linking).
[...]

PR #1167, opened 2026-07-04
feat(core): local encryption key store + passphrase escrow (C-2, #825)

BYK:
## Summary

Second slice of the **encryption + key-management epic (#825)** — the local, db-backed **key store**, built on the C-1 primitives (#1162). No remote/sync yet (that is C-3); this establishes the on-device key hierarchy and the passphrase-escrow flow that makes multi-device work day one.

## Key model (v1 personal)
- **One per-user account identity keypair** (X25519). Its secret is stored in the clear in the **local** SQLite — which is already the plaintext store for conversations/knowledge, guarded by the db file's `0600` mode. Encryption protects only what is pushed to the **remote**. The account key is recovered on other devices via passphrase escrow (a per-device keypair / QR-PAKE pairing is the client-only alternative deferred to C-5).
- **Per-scope DEK**, HPKE-wrapped to the account public key. The DEK plaintext is **never persisted** — unwrapped on demand, cached in memory.
- **Escrow**: the account secret wrapped by an `Argon2id(passphrase)` KEK, so a fresh device recovers the **same** account key with the passphrase. An optional **recovery code** is an independent unlock path.

## What lands
- **v64 migration** (`db.ts`) — `account_identity` (single-row, local-only), `account_escrow` (single-row: wrapped secret + KDF params/salt + optional recovery wrapping with its **own** KDF params), `scope_keys` (PK `scope_id`+`member_user_id`: wrapped DEK). Plus a `recoverMissingObjects` block; schema_version → 64.
- **`crypto/keystore.ts`** — `getAccountIdentity` (create-on-first-use, **`KeystoreLockedError`** when escrow exists but no local identity → a fresh device must unlock, never mints a divergent key), `getScopeKey` (generate+wrap / unwrap, in-memory DEK cache), `setPassphrase` (+ optional recovery code), `unlockWithPassphrase`/`unlockWithRecoveryCode`, `putWrappedScopeKey`, `lock`, `hasAccountIdentity`, `hasEscrow`.
- Barrel: `export * as keystore`.

## 🔴 Invariants
- A fresh device (escrow present, no local identity) is **LOCKED** — it never mints a new identity (that key could not unwrap DEKs wrapped to the real account key). Mutation-verified.
- `account_identity` (holds the plaintext secret) is **never** a synced table — asserted by a test that reads `SYNCED_TABLES`.
- The plaintext DEK is never written to disk; `scope_keys` holds only the wrapped blob.

## Review
Two adversarial subagent reviews — **security** (fresh-device guard mutation-proven) and **correctness** — both **SHIP, no blockers**. Fixes applied in-PR:
- A passphrase change no longer silently **wipes a preserved recovery code** (`COALESCE` preserve; the identity secret is unchanged so the old wrapping stays valid).
- Recovery carries its **own** KDF params, so a params-changing passphrase change cannot silently invalidate it (surfaced by the correctness reviewer; fixed now since the schema is unshipped, avoiding a C-3 migration).

## Tests (18 keystore + schema bumps)
Fresh-device lock guard (identity + scope-key paths), wrong-passphrase rejection (no partial install), DEK-never-in-plaintext, DEK stable across `lock()`, distinct-scope DEKs, recovery path + negative, **recovery preservation across passphrase change**, **recovery param decoupling**, `putWrappedScopeKey` apply, and the `account_identity`-never-synced invariant. All mutation-verified discriminating.

## Verification
Full suite green (5199 passed); typecheck 0 (all packages); Biome lint 0; core esbuild bundle builds.

## Follow-ups (tracked, non-blocking)
- **C-3 gate**: must pull `account_escrow` before any first encryption use, so a genuinely fresh device can't mint a divergent key before sync.
- `getScopeKey`'s `memberUserId` param + scope-keyed cache is v1-only-correct; tighten/document before teams (E).

## Next
- **C-3**: `scope_keys` + `account_escrow` remote mirror (migration + RLS: a member pulls only what is wrapped for them) + fresh-device unlock flow.
- **C-4**: wire encrypt/decrypt into push/pull for `knowledge.content`.


BYK:
Fixed the `getScopeKey` first-use race (Sentry CRITICAL). Two independent guards now prevent a lost/clobbered DEK:
1. **In-flight dedup** — concurrent `getScopeKey(scope)` calls share one `Promise` (`pendingScopeKey` map, keyed like `dekCache`), so exactly one DEK is generated per scope.
2. **Convergent write** — the insert is now `ON CONFLICT(scope_id, member_user_id) DO NOTHING`, after which we re-read the persisted row and unwrap **that**. So even if two writers race, everyone converges on the single stored DEK; a lost insert can never leave earlier ciphertext unopenable.

Added a regression test (`concurrent first-use calls for one scope converge on a single DEK`) asserting three concurrent first-use calls return the same DEK, exactly one row persists, and it survives a `lock()` reload.

PR #1170, opened 2026-07-04
fix(core): cap native ONNX intra-op threads to the cgroup CPU quota

BYK:
Companion to #1168. That PR fixed the **memory** axis of container cgroup-blindness; this fixes the **CPU/thread** axis.

## Problem
Native ONNX Runtime sizes its intra-op thread pool to `std::thread::hardware_concurrency()` — the **host** core count — which is cgroup-CPU-blind. In a CPU-quota'd container (Docker/Railway/K8s), a 1–2 vCPU container on a many-core host spawns host-core-count intra-op threads. Each intra-op thread carries its own ORT memory **arena**, so this inflates per-worker RSS (the axis the #1168 memory clamp doesn't cover) and thrashes on context switches against the quota. The worker comment already noted native "scales with cores (#999)" — that's exactly the unbounded behavior in a container.

## Fix
Pass `session_options.intraOpNumThreads = os.availableParallelism()` to the transformers.js pipeline on the **native path only**. `availableParallelism()` is cgroup-CPU-aware (libuv `uv_available_parallelism`: cgroup v2 `cpu.max`, v1 CFS quota, `sched_getaffinity`).

**Only caps when genuinely CPU-restricted** — `availableParallelism() < os.cpus().length`. `nativeIntraOpThreads()` returns `undefined` (leave ORT's default untouched) when the two are equal:
- **No-op on unconstrained hosts** (same guarantee as #1168; e.g. a 4-core dev box: `avail 4 == cpus 4` → `undefined`).
- Avoids *raising* threads above ORT's physical-core default on a **hyper-threaded** host (where logical > physical would otherwise oversubscribe).

WASM is untouched — it's already forced single-thread via `env.backends.onnx.wasm` in the worker. Native-path detection reuses the existing `__LORE_NPM_WASM_PATHS__` signal (set only when the npm bundle falls back to WASM).

## Also
Folds in two review-comment nits from #1168 (doc-only): the `constrainedMemory()` JSDoc (clarify it's the cgroup *limit*, not free-within-limit; the pool-growth gate is per-embed only during the growth window) and a limit≠free note on `containerFreeBytes`.

## Testing
- `nativeIntraOpThreads` unit tests: undefined on unconstrained (`avail==cpus`) and on `avail>cpus` (never raises); caps to the quota when constrained; floors fractional/invalid at 1.
- **Mutation-verified**: forcing "always cap" reds the 2 no-op guards; forcing "never cap" reds the 2 cap guards.
- Full `ort-native` + `embedding-*` suites (222 passed), `tsc --noEmit`, Biome all clean.

## Scope
With #1168 (over-spawn + over-size) + this (thread/arena oversubscription), both cgroup-blind axes are bounded. A single native worker's baseline model RSS can still exceed a sub-GB container — for those, `search.embeddings.provider: "voyage"/"openai"` or `search.embeddings.enabled: false` remain the escape hatches.

### DIFF
--- a/packages/core/src/embedding-worker-types.ts
+++ b/packages/core/src/embedding-worker-types.ts
@@ -324,6 +324,14 @@ export interface WorkerInitData {
    *  bounds the O(L²) attention allocation and keeps a constrained host out of
    *  swap. */
   maxTokens: number;
+  /** Intra-op thread count for native ONNX Runtime, or `undefined` to leave
+   *  ORT's own (host-core-sized) default in place. Computed on the main thread
+   *  via `nativeIntraOpThreads()` (the worker runs as raw .ts and can't
+   *  value-import `ort-native` — same constraint as the stderr-silence flag
+   *  below): caps intra-op threads to the cgroup CPU quota only when the process
+   *  is genuinely CPU-restricted, else `undefined` (a strict no-op). The worker
+   *  applies it on the native path only; WASM is already single-threaded. */
+  intraOpThreads?: number;
   /** Vendored model info for binary mode, or null for npm mode.
    *  In binary mode, model files are pre-extracted to a local dir
    *  and we point transformers.js at that path instead of downloading
--- a/packages/core/src/embedding-worker.ts
+++ b/packages/core/src/embedding-worker.ts
@@ -336,10 +336,30 @@ async function loadPipeline(): Promise<void> {
   // CPU backend (multi-threaded). The npm dist-only bundle redirects
   // onnxruntime-node → onnxruntime-web, which serves "cpu" via its WASM+SIMD
   // backend (API-compatible).
-  pipe = (await pipeline("feature-extraction", modelId, {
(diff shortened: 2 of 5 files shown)


BYK:
CI caught a real issue: the first commit used `await import('./ort-native')` inside the worker, but the embedding worker runs as raw .ts spawned by Node's native resolver (which can't map `./ort-native.js` → the .ts source at runtime — same constraint documented for `maxTokens`/the stderr-silence flag). That crashed worker init on the dev/test path (6 `LocalProvider` real-worker tests in `embedding.test.ts`).

Fixed in 11197824: compute `nativeIntraOpThreads()` on the main thread and thread it through `WorkerInitData.intraOpThreads`; the worker just consumes the number and gates on the native path. Reproduced the failure locally with the vendored model (`LORE_LOCAL_MODEL_PATH`) — the bad import fails the same 6 tests; the fix passes all 53.

PR #1171, opened 2026-07-04
feat(sync): client-side encryption of knowledge content (C-4, #825)

BYK:
## What

Stacked after C-3 (#1169). Encrypts `knowledge.content` + `knowledge.title` on the sync wire so the server only ever stores ciphertext. Local storage stays **plaintext** (FTS + embeddings unaffected; DB file is chmod 0600). **Dormant until C-4b** wires the passphrase UX — with no escrow set up, `encryptionState()` is `off` and sync stays plaintext, so merging is a no-op for current users.

## Design
- **`keystore.encryptionState()`**: `off` (no escrow — never encrypt) / `locked` (escrow pulled via C-3 but identity not installed — can't en/decrypt) / `on` (identity present). Gates on **escrow existence** so we never encrypt with a non-recoverable, auto-minted escrow-less key.
- **`SyncTableMeta.encryptedColumns`** = `[content, title]` for knowledge. Sealed to the per-scope DEK (`crypto.seal`) + base64 on push; decrypted on pull. `content_hash` is computed over the **plaintext** row, so it stays cross-device stable (dedup/skip still work).
- **AAD** = `buildAad(scope, table, column, logicalId)` (length-prefixed, per the C-1 review) — binds each ciphertext to its row/column/scope; it can't be transplanted.
- **Push**: `locked` pauses the knowledge table (`stop`); `on` encrypts; `off` = plaintext.
- **Pull**: the key tables (`account_escrow`, `scope_keys`) are **reordered before `knowledge`** so a fresh device is `locked` (skip, cursor frozen) rather than a plaintext-passthrough `off`. Decryption happens **before** any classify/conflict side effect; an undecryptable envelope with no key (`isEnvelope` backstop) aborts the table with the cursor frozen — **never storing ciphertext as content**.
- **Migration 0011**: widen the free `knowledge` byte cap 8 MB → 12 MB for ciphertext overhead (only the free tier is byte-capped; pro is uncapped).

## Tests
- **core**: `encryptionState` transitions (off → locked → on) driven by escrow/identity presence.
- **gateway (mock)**: push seals content+title (remote holds envelopes, not plaintext); `content_hash` == the plaintext local hash; pull decrypts back with no conflict; `off` pushes plaintext; `locked` skips push **and** pull (cursor frozen); backstop — an `off` device with an envelope in the remote aborts, never corrupting local content.
- **integration (Docker)**: knowledge byte cap widened to 12 MB.

Local: typecheck clean (4 pkgs), lint clean, full suite green (5313), integration 40/40.

## Follow-up (C-4b)
- 🔴 wire `lore sync enable` → `setPassphrase` + unlock UX; enforce `scope == auth.uid()` at the caller; unlock-before-sync so `locked` resolves.
- key rotation (`key_epoch > 0`) — v1 is epoch 0 only.


BYK:
Adversarial correctness + crypto review (read-only, mutation-verified in a throwaway worktree): **SHIP-WITH-FOLLOWUPS**, no BLOCKERs. Default/`off` wire path is byte-identical to pre-PR; encrypted path round-trips; content_hash stays plaintext-stable; pull fails **closed** (never stores ciphertext locally).

**Fixed in-PR:**
1. **Push now fails closed** — when `encryptionState()==='on'` but the scope/DEK can't resolve (empty `user_id`), `pushEntry` returns `stop` instead of leaking plaintext. (+test)
2. **Per-row `knowledge_size_ck` widened** (0011) — content 8192→12288, title 512→1024 so ciphertext-inflated rows aren't silently `poison`-dropped. (+integration test)
+ multibyte-UTF-8 round-trip test.

**Deferred to C-4b (only reachable post-activation / multi-device):**
3. `scope_keys` remote first-write-wins (`ON CONFLICT DO NOTHING`) to prevent a multi-device DEK clobber.
4. Replace `isEnvelope` content-sniffing with an explicit encrypted-row signal (theoretical wedge on a false-positive plaintext).

Local: typecheck + lint clean, full suite green (5314; one unrelated `recall-graph` flake, passes 14/14 in isolation), integration 41/41.

Issue #827, opened 2026-06-19
E — Teams & orgs: shared memory (later)


BYK:
> **Part of epic #821.** Depends on: #825, #826
> _This issue body mirrors the design doc `.opencode/plans/sync-E-teams-orgs.md`._

---

# E — Teams & orgs: shared memory (later)

Parent: `sync-sharing-epic.md`. Depends on: C (encryption/KM) + D (pro tables).
Not scheduled yet — captured so A1–D leave the right seams.

## Goal
Let a `scope_id` be a TEAM/ORG (not just a user), so selected members share
knowledge, entities, distillations, and temporal — converging across all members
and the remote, with encryption gating content.

## What lands here (because A1–D pre-built the seams)
- `membership(scope_id, member_user_id, role)` table.
- 🔴 RLS predicate swap ONLY: `scope_id = auth.uid()` → `is_member(scope_id)`
  on every shareable table. No PK change, no data migration (A1 did the seam).
- 🔴 `effective_tier(scope_id)` replaces `current_tier()`: entitlement attaches
  to the scope; members inherit (A1/D wrote the gate as scope-based).
- 🔴 Group DEK wrapping: wrap the per-scope DEK to each member's identity key
  (C's interface already supports N recipients); rotate DEK on member removal.
- Quota/billing bills the `scope_id` (A1 counter PK is `(scope_id, table_name)`).
- Conflict: concurrent authors → A2's append-only log keeps both versions
  (no silent loss); materialized current picks deterministically; surface
  conflicts for merge.

## MLS escalation (epic decision #6)
- Switch key-agreement from envelope wrapping to MLS (RFC 9420, OpenMLS) when
  groups get large/churny or forward-secrecy / post-compromise security is
  required. Additive via the versioned envelope — old blobs stay readable.

## Invariants (🔴)
- The ONLY isolation change from personal→team is the RLS membership predicate.
- A non-member can neither fetch (RLS) nor decrypt (no wrapped DEK) shared rows.
- Removing a member rotates the DEK for future writes; past data they could
  already read is inherently theirs (document).

## Open
- Roles/permissions granularity (viewer/editor/admin).
- Org = scope-of-scopes (nested) vs flat teams.
- Per-row vs field-level merge for high-conflict shared knowledge.

BYK:
Detailed implementation design pass done → `.opencode/plans/sync-E-teams-orgs-impl.md` (file:line-grounded against the A1–D seams).

[...]

PR #1174, opened 2026-07-04
feat(sync): passphrase UX to activate knowledge encryption (C-4b, #825)

BYK:
## What

Stacked after C-4 (#1171). Wires the passphrase/unlock flow into `lore sync enable` so C-4a's dormant encryption actually turns on. After this, enabling sync guarantees **sync ⟹ end-to-end-encrypted knowledge**.

## Flow — `bootstrapEncryption(pull, prompts)` (pure, injectable for tests)
1. **Pull escrow first** so a fresh device *unlocks* the existing account key rather than minting a divergent one (prevents the multi-device DEK clobber the C-4 review flagged). `pullOnce` doesn't push → no plaintext leak; knowledge stays deferred while locked.
2. Branch on `keystore.encryptionState()`:
   - **off** (first device, no escrow anywhere) → prompt a new passphrase (+ confirm), `setPassphrase` with a generated **recovery code** (shown once) → `on`.
   - **locked** (fresh device, escrow pulled, no local identity) → prompt the existing passphrase / recovery code, unlock (retry ≤5×) → `on`.
   - **on** → no-op.
   - declined/exhausted → `aborted`.

## CLI
- `lore sync enable`: runs the bootstrap **before** the first push. On `aborted`, reverts an enable we just made (`disableSync`) so sync is never left armed-but-plaintext. Hidden passphrase input (muted `readline`); 120-bit Crockford-base32 recovery code, paste-tolerant (dashes/spaces/case normalized).
- `lore sync status`: adds `Encryption: on | locked | off`.
- **No gateway prompt needed**: `unlockWithPassphrase` persists the identity to `account_identity`, and device-1 stores it locally — so the separate gateway process reads `on` straight from the DB.

## scope == auth.uid()
The sync resolver is the only `getScopeKey` caller and always passes `getCurrentUser().user_id` (= auth.uid() = server scope_id); the invariant holds by construction in v1 personal.

## Tests
`bootstrapEncryption` state machine: off→setup(+recovery), locked→unlock, wrong-passphrase-retry→dashed-recovery, on→noop, abort, and **pull-first routes to unlock not setup** (no clobber); + recovery-code generate/normalize. 9 tests.

Local: typecheck + lint clean; targeted sync/keystore suites green; full suite green apart from a **pre-existing** `recall-graph.test.ts` flake (unrelated — this PR only touches the sync CLI; it flakes on main in isolation too).

## Deferred (documented follow-ups)
- Re-encrypt pre-C-4b **plaintext** knowledge already on the remote (mixed-state) — needs a force-repush; rare pre-GA population.
- `scope_keys` remote first-write-wins (multi-device clobber) — prevented in practice by the pull-first bootstrap for v1 personal; real hardening lands with teams (#827).
- Explicit encrypted-row scheme flag (retire `isEnvelope` sniffing); key rotation (`key_epoch>0`).


BYK:
Adversarial review returned **BLOCK** on one real issue (SP1c) — now fixed.

**BLOCKER (fixed): silent-clobber on transient pull failure.** `pullOnce` swallows network errors internally, so the bootstrap couldn't tell "confirmed no remote key" from "pull failed" — a transient failure on a fresh device with a real remote key would route to first-device setup, mint a divergent identity/escrow, and `runSync` push-then-pull would clobber device-1's escrow+DEK (remote has no immutability guard), locking it out.

Fix: `bootstrapEncryption` now takes `syncKeys(): Promise<{confirmed}>` and **only mints a first-device key when `confirmed` is true**. The `cmdEnable` closure positively verifies the remote escrow state (`pullOnce` + a direct `account_escrow` existence query); any error, or a row present while still `off`, → `confirmed:false` → abort (no mint). Fail-safe.

Also addressed:
- **SP8 (highest-value missing test):** added direct `cmdEnable` tests — revert-on-abort (disable + no `runSync`), unreachable-escrow-check refuses setup, not-logged-in, and the happy path.
- **SP2/SP9 nits:** `cmdStatus` wording softened to "on (knowledge encrypted)" (only content/title are sealed); `promptSecret` `_writeToOutput` portability note.

New fail-safe tests are discriminating (removing the guard fails them). 15 tests total; targeted sync/keystore suites green (91).

BYK:
Addressed both Sentry findings:
- **HIGH (promptSecret private API):** rewrote hidden input to read stdin in **raw mode** and never echo keystrokes — no dependency on the private `readline._writeToOutput` (which risked leaking the passphrase in plaintext if that Node internal changed). Handles enter/backspace/Ctrl-C/Ctrl-D.
- **LOW (recovery-code bit extraction):** rewrote `generateRecoveryCode` as a standard base32 accumulator (feed 8 bits, emit 5, mask leftovers) — obviously correct and 32-bit-safe; 120 bits ÷ 5 = 24 symbols exactly.

15/15 tests green, typecheck + lint clean.

BYK:
Addressed the MEDIUM: `normalizeRecoveryCode` now applies Crockford base32 error-correction (O→0, I/L→1) so a hand-transcribed recovery code unlocks even with the ambiguous chars the generator alphabet omits. +test.

BYK:
Hardened the raw-mode `promptSecret`:
- **HIGH (EOF hang):** added an `end` handler so piped/closed stdin resolves instead of hanging.
- **MEDIUM (split UTF-8):** `stdin.setEncoding('utf8')` so the stream's StringDecoder buffers a multi-byte char split across `data` chunks — we always iterate whole code points.

BYK:
Addressed the CRITICAL: **encryption is now required for sync.** On abort, `lore sync enable` disables sync unconditionally (including a pre-existing plaintext sync from before encryption existed), so declining setup can never leave the gateway pushing plaintext. Message updated accordingly. +test (pre-existing sync + abort → disabled).

PR #1179, opened 2026-07-05
fix(sync): scope_keys.wrapped_dek first-write-wins on the server (#825)

BYK:
## What
Defense-in-depth for the multi-device **DEK clobber** both the C-4 and C-4b reviews flagged as the remaining gap.

`scope_keys` holds the per-scope data-encryption key wrapped (HPKE) to a member's account key. RLS lets the owner UPDATE their own row, so a second device that minted a **divergent identity** (bug/race/bypassed client) could overwrite `wrapped_dek` with a DEK wrapped to a *different* key — which **orphans every ciphertext** sealed under the original DEK and **locks the original device out** (it pulls a DEK it can't HPKE-unwrap).

## Fix — migration 0012
A `BEFORE UPDATE` trigger `guard_scope_key_immutable` rejects any change to `wrapped_dek` at the **same-or-lower `key_epoch`** (`check_violation`, 23514):
- **allowed**: INSERT (first write); metadata-only bumps (updated_at/revision/content_hash/is_deleted) that keep `wrapped_dek`; a genuine key **rotation** (strictly higher epoch may carry a new DEK — future).
- **blocked**: changing `wrapped_dek` at the same epoch (the clobber).
- The client already classifies a 23514 **without** "quota" as **poison** (`sync.ts:classifyPushError`), so a blocked clobber-push is dropped + logged and never wedges the table — **no client change needed**.
- Trigger name sorts before `scope_keys_row_quota`/`_set_updated_at`, so it fires first.
- Supabase-only migration — local SQLite schema unaffected.

The C-4b bootstrap already pulls-before-mint to avoid this; 0012 makes it **durable** at the server.

## Non-goal
`account_escrow` stays **mutable** — a passphrase change legitimately re-wraps the account secret and must sync. The rare simultaneous-first-setup race (LWW escrow vs FWW DEK) is documented as **recoverable, no data loss** (the DEK winner stays functional; a device that pulled a mismatched escrow fails closed and recovers on re-sync). A full escrow-consistency guard (claims/epochs) is future work (teams, #827).

## Tests
Tier-1 Postgres integration (real RLS+triggers): metadata-only update OK; same-epoch `wrapped_dek` change **rejected 23514** (original DEK survives); higher-epoch rotation OK. 44/44 integration pass; typecheck + lint clean.


BYK:
Adversarial review: **SHIP**, no BLOCKERs — trigger correctness, ordering (guard fires first), and the 23514→poison→drop-and-advance client path all verified; no legitimate flow changes wrapped_dek at the same epoch; security-neutral, no cross-tenant reach.

Applied the two non-blocking follow-ups:
- Comment clarifying the guard is **UPDATE-scoped defense-in-depth** (self-only bypasses — lower-epoch/DELETE+INSERT — hardened later with #827), not a rollback-proof constraint.
- +2 tests: idempotent same-`wrapped_dek` upsert allowed (production convergence path); a **lower**-epoch dek change is also rejected (`<=` semantics). 46/46 integration pass.

BYK:
Addressed the LOW: collapsed the test helpers to a single `user` param (v1: member_user_id == scope_id == author_id), binding $1 (text, member) and $2 (uuid, scope) explicitly — removes the `scope`-vs-`user` footgun. 46/46 integration green.

PR #1181, opened 2026-07-05
test(sync): end-to-end encrypted knowledge round-trip through the real engine (#825)

BYK:
## What
A Tier-2 integration test (real Postgres + PostgREST + supabase-js, driving the actual `pushOnce`/`pullOnce`) for the full **C-1..C-4b encryption stack** — the layer the mock can't exercise: base64 envelope wire encoding, RLS/triggers on encrypted payloads, the escrow + scope_keys HTTP round-trip, and real HPKE DEK unwrap. This is the highest-fidelity **local** test of encrypted sync before manual testing.

## Tests
- **device-1 encrypts → remote stores CIPHERTEXT → device-2 unlocks + pulls plaintext**: asserts the remote `knowledge.content` AND `title` are real envelopes (not plaintext); a fresh device-2 pulls escrow first (→ `locked`, knowledge deferred), unlocks with the passphrase (→ `on`), pulls again and decrypts back to the exact original plaintext.
- **wrong passphrase does not unlock; knowledge stays deferred.**

`getCurrentUser` is pinned to the test user (encryption scope = the JWT's `auth.uid()`); all other real `supabase` exports are preserved. Test-only — no production code changes.

## Real gap this surfaced (follow-up)
The DEK (`scope_keys`) is minted **lazily inside** the first encrypted push (`getScopeKey` during encrypt), so it lands on the remote **one cycle after** the ciphertext knowledge. A fresh device pulling in that window would mint a **divergent DEK** → with 0012 first-write-wins, that can orphan device-1's data. Documented in-test; fix (**eager-mint the DEK at `lore sync enable`** so scope_keys ships with the first push) will be a separate PR. The test flushes it deterministically with a second push.

Runs under `LORE_INTEGRATION=1` (Docker). 11/11 engine integration tests pass; typecheck + lint clean; the normal suite skips it.

### DIFF
--- a/packages/gateway/test/sync-engine.integration.test.ts
+++ b/packages/gateway/test/sync-engine.integration.test.ts
@@ -13,7 +13,15 @@
  */
 import { execFileSync } from "node:child_process";
 import { createClient, type SupabaseClient } from "@supabase/supabase-js";
-import { db, ensureProject, ltm, setKV, syncData } from "@loreai/core";
+import {
+  crypto,
+  db,
+  ensureProject,
+  keystore,
+  ltm,
+  setKV,
+  syncData,
+} from "@loreai/core";
 import {
   afterAll,
   afterEach,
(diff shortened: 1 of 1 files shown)


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — test is discriminative on every property (fails if push emits plaintext, if locked→unlock→on breaks, or if a wrong passphrase decrypts); runs as a required CI check against real Postgres/PostgREST; `vi.mock` is legitimate + file-scoped.

Applied the review follow-ups:
- **Fidelity**: device-2 wipe now also clears `sync_state` + `sync_outbox` → true fresh-device clean-apply path; assert `conflicts === 0` on both pulls and restored the no-ping-pong `pushed === 0` assertion.
- **Symmetry**: added `title !== "T"`; aligned the cursor literal with the top-level `31_536_000_000`.
- **Lazy-mint DEK gap** filed as #1182 (eager-mint at enable) and referenced in the in-test NOTE.

11/11 integration pass; typecheck + lint clean.

Issue #1182, opened 2026-07-05
Eager-mint the scope DEK at `lore sync enable` (avoid divergent-DEK window)


BYK:
Surfaced by the end-to-end encrypted round-trip test (PR #1181).

## Problem
The per-scope DEK (`scope_keys`) is minted **lazily inside** the first encrypted knowledge push: `pushOnce` → `encryptColumns` → `getScopeKey` → `loadOrCreateScopeKey` INSERTs the row `ON CONFLICT DO NOTHING`. That INSERT's outbox capture fires *after* the `scope_keys` table was already drained in the same push cycle, so:

- **cycle 1**: ciphertext `knowledge` reaches the remote, but `scope_keys` does **not**.
- **cycle 2** (~30s later): `scope_keys` is pushed.

A fresh device-2 that pulls in that window sees the escrow + ciphertext knowledge but **no `scope_keys`**. After unlocking, its `getScopeKey` finds no local row and **mints a divergent DEK** (a new random key wrapped to the same account key). With migration 0012 (`scope_keys.wrapped_dek` first-write-wins), whichever DEK lands first wins:
- device-2's divergent DEK wins → device-1's knowledge (sealed under the original DEK) is **orphaned** and device-1 can no longer decrypt after it re-pulls scope_keys; or
- device-1's real DEK wins → device-2 holds a cached wrong DEK and can't decrypt.

Either way this is a real data-orphaning / lockout risk once multi-device sync is exercised.

## Fix
Eager-mint the DEK when encryption is activated, so `scope_keys` is enqueued **before** the first knowledge push and ships in the same cycle as (or before) the ciphertext:
- In `cmdEnable`'s bootstrap (`packages/gateway/src/cli/sync-cmd.ts`), after the keystore reaches `on`, call `await keystore.getScopeKey(user.user_id)` before `runSync`. This mints + enqueues `scope_keys` so the first sync push includes it.
- Encryption is only ever activated via `lore sync enable`, so this closes the window for all paths (the gateway's later pushes always find scope_keys already established).

## Tests
- Add a Tier-2 integration assertion that after the FIRST push following enable, the remote already holds `scope_keys` (no second push needed) — i.e. remove the two-push workaround in `sync-engine.integration.test.ts` once the fix lands.

Part of #825.

BYK:
Already shipped — code + test both merged in PR #1183 (`bcac023e`), no remaining work.

[...]

PR #1183, opened 2026-07-05
fix(sync): eager-mint the scope DEK at `lore sync enable` (#1182)

BYK:
## Problem
The per-scope DEK (`scope_keys`) was minted **lazily inside** the first encrypted knowledge push (`getScopeKey` within `encryptColumns`). That INSERT's outbox capture fires *after* the `scope_keys` table was already drained in the same cycle, so the ciphertext `knowledge` reaches the remote **one cycle before** `scope_keys`. A fresh device pulling in that window finds ciphertext + escrow but **no key**, and after unlocking mints a **divergent DEK** — which under migration 0012 (`wrapped_dek` first-write-wins) can **orphan** device-1's data or lock a device out.

Surfaced by the end-to-end encrypted round-trip test (#1181), which had to work around it with a second push.

## Fix
`cmdEnable` eager-mints the DEK (`keystore.getScopeKey(user.user_id)`) as soon as encryption reaches `"on"`, **before** `runSync` — so `scope_keys` is enqueued and ships in the **first** push (before/with the encrypted knowledge). Encryption is only ever armed via `lore sync enable`, so this single place closes the window for all paths (the gateway's later pushes always find `scope_keys` already established). `getScopeKey` is idempotent (`ON CONFLICT DO NOTHING`), so re-running `enable` is a no-op.

## Tests
- `sync-cmd.test.ts`: new test asserts `cmdEnable` mints `scope_keys` locally **and** enqueues it in the outbox before the sync push runs.
- Updated the engine e2e NOTE to point at the CLI eager-mint (that test bypasses `cmdEnable`, so it still flushes with a second push).

Typecheck + lint clean; 18 sync-cmd tests + 11 engine integration tests pass.

### DIFF
--- a/packages/gateway/src/cli/sync-cmd.ts
+++ b/packages/gateway/src/cli/sync-cmd.ts
@@ -96,6 +96,24 @@ export async function cmdEnable(
     process.exitCode = 1;
     return;
   }
+
+  // Eager-mint the per-scope DEK now that encryption is "on", so scope_keys is enqueued
+  // and ships with the FIRST push. Otherwise it's minted lazily during the first knowledge
+  // encrypt (getScopeKey inside encryptColumns) and its capture lands a cycle later — a
+  // window where a fresh device pulls the ciphertext without the key and mints a DIVERGENT
+  // DEK, which under 0012 first-write-wins can orphan the data (#1182).
+  // Guard on user_id (mirrors the resolver's fail-closed contract) and swallow a transient
+  // mint failure: it degrades to the pre-fix lazy mint rather than aborting a valid enable.
+  if (user.user_id) {
+    try {
+      await keystore.getScopeKey(user.user_id);
+    } catch (e) {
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial review: **SHIP**, no BLOCKERs — mutation-verified the new test is non-vacuous (removing the eager-mint makes it fail), identity is provably installed at the call site, the minted scope matches the push/pull resolver, and scope_keys is captured + pushed ahead of knowledge in cycle 1 (window closed, not narrowed).

Applied the 3 optional follow-ups: (a) guard `user.user_id` truthiness (mirrors the resolver's fail-closed contract), (b) wrap the eager-mint in try/catch so a transient failure degrades to the lazy mint instead of aborting a valid enable, (c) clear `sync_outbox` in the test `beforeEach` so the enqueue assertion is independently robust. 18/18 pass; typecheck + lint clean.

PR #1187, opened 2026-07-06
fix(sync): don't send a phantom content_hash on join-table deletes (PGRST204)

BYK:
## Bug (observed on a live initial sync)
```
sync: push delete knowledge_entity_refs/…: Could not find the 'content_hash' column of 'knowledge_entity_refs' in the schema cache
```
The tombstone/DELETE push (`sync.ts:316`) **unconditionally** set `content_hash: null`. Join tables like `knowledge_entity_refs` are `versioned:false` and have **no `content_hash` column** remotely (only `is_deleted`), so PostgREST rejects the whole delete with **PGRST204**. Since that's not a `23514`, `classifyPushError` calls it **transient** → the delete retries every cycle forever → **`knowledge_entity_refs` sync wedges**.

The *upsert* path already gates this correctly (`if (tableMeta(table).versioned !== false)` at line 384); the delete path didn't.

## Fix
Gate `content_hash: null` on the tombstone behind `tableMeta(table).versioned !== false`, mirroring the upsert path. Versioned:false tables get a plain `{ is_deleted: true }` soft-delete — the column the join table actually has.

## Test
Tier-2 engine integration (real Postgres + PostgREST): push a ref, delete it locally, push the tombstone → assert it uploads (no PGRST204) and the remote row is `is_deleted=true`. **Verified discriminating** — fails with the gate removed.

12/12 engine integration + 47 unit sync tests pass; typecheck + lint clean.

### DIFF
--- a/packages/gateway/src/sync.ts
+++ b/packages/gateway/src/sync.ts
@@ -298,15 +298,18 @@ async function pushEntry(
   }
 
   if (op === "delete") {
-    // Null the remote content_hash too, so the tombstone honors the
-    // "remoteHash is null for a tombstone" contract on the wire. The pull side
-    // (applyRemote) already treats is_deleted rows as hash-null, but this also
-    // protects un-upgraded readers during a rollout.
-    // Null the remote content_hash too (tombstone "remoteHash is null" contract).
     const tombstone: Record<string, unknown> = {
       is_deleted: true,
-      content_hash: null,
     };
+    // Only versioned tables have a content_hash column remotely; the join table
+    // (knowledge_entity_refs) does not, so sending it is a PGRST204 schema error that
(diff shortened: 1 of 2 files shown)


BYK:
Adversarial review: **SHIP**, no BLOCKERs — fix precisely mirrors the upsert `versioned` gate, no other phantom column reaches the join-table tombstone, composite `.match` targets the right row, and pull symmetry is preserved (soft-deleted ref applies + doesn't resurrect). Test verified discriminating.

- Applied the NIT: tightened `toBeGreaterThanOrEqual(1)` → `toBe(1)`.
- Filed the latent-wedge CONCERN as #1188 (delete branch never classifies errors → PGRST schema errors retry forever; classify PGRST* as poison + route the delete branch through classifyPushError).

PR #1189, opened 2026-07-06
feat(sync): seed the outbox by value so a capped free tier syncs the most useful knowledge first

BYK:
## Why
On the free tier the server rejects knowledge inserts past the row cap (500 rows), and the push drains the outbox in **seq order** — so whatever is enqueued first wins the cap. `seedOutbox` enqueued knowledge in arbitrary storage order (`SELECT … FROM knowledge_current` with no `ORDER BY`, ≈ oldest-created), so genuinely useful / recent entries could be the ones left behind. Free tier should be *limited, not punishing* — keep the **most useful** entries.

## Change
Order the knowledge seed by value:
```sql
ORDER BY confidence DESC,
         COALESCE(last_reinforced_at, updated_at) DESC,
         created_at DESC
```
so the most-confident, most-recently-used/updated entries take the low sequence numbers and survive the cap. `knowledge_current` (v56) already exposes `confidence` + `last_reinforced_at`. No behavior change once under the cap — all entries still sync; only the *order* (and thus what survives a cap) changes.

## Test
Seed three entries created in **non-value** order with distinct confidence → assert the outbox order is value-ranked `[high, mid, low]`. **Verified discriminating** (fails with the `ORDER BY` removed). Full `sync-data` suite: 68/68.

## Follow-up (not in this PR)
This fixes the *initial fill*. Maintaining the top-N as the base grows past the cap (evict a low-value synced row when a more-useful one appears) is a larger feature — will file separately. Entity-graph tables (entities/aliases) also have caps and could get value-ranked seeding later.


BYK:
Adversarial review: **SHIP** — mutation-verified discriminating test; DISTINCT drop provably redundant (UNIQUE idx_knowledge_one_current → one current row per logical_id); ordering behavior-neutral under the cap; columns resolve unambiguously on the v56 view.

Added the optional follow-ups: a **recency tie-break test** (equal confidence → more recently reinforced wins, exercising the secondary sort) and a one-line note that a meta-less current row reads as confidence 1.0 (legacy default) and sorts high.

PR #1190, opened 2026-07-06
fix(sync): de-spam the quota-pause WARN — log once per pause, not every cycle

BYK:
## Problem
A quota-paused table keeps its rows pending and is retried every sync cycle, so the per-table `sync: quota on X — N of N rows` WARN **re-logged the identical message every ~60s forever** (plus a redundant scheduler `quota reached on X` summary). Observed live:
```
[WARN] sync: quota on knowledge_meta — … 500 of 500 rows. Upgrade to sync more.
[WARN] sync: quota on entities — … 30 of 30 rows. …
[WARN] sync: quota reached on knowledge; that table paused until next change
```
…repeating every cycle.

## Fix
- **Dedupe** the per-table quota WARN via a module-level `quotaWarnedTables` set: log **once** when a table enters quota-pause; **clear** the flag when the table next makes real progress (a slot freed / the cap lifted) so a later re-pause warns once more.
- **Remove** the scheduler's per-cycle `quota reached` summary — redundant with the now-deduped, more-informative per-table log.

Behavior is otherwise unchanged (paused rows still retry; nothing is lost).

## Tests
- A persistent pause warns **exactly once** across 3 cycles.
- A pause **re-warns once** after the table makes progress then re-hits the cap.

Both **mutation-verified** (removing the dedupe → first test fails; removing the clear → second test fails). 49/49 sync unit tests pass; typecheck + lint clean.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — exactly-once at a hard cap, correct clear-on-progress placement (incl. the deliberate no-op exclusion), and the user-facing quota signal is preserved (res.quotaHit + the CLI report) after dropping the redundant scheduler summary. Both tests mutation-verified.

Applied the test-hygiene NITs: added `vi.restoreAllMocks()` to `beforeEach` so a spy leaked by a failed assertion can't cascade into later tests, and tightened the log filter to `"quota on knowledge —"` so it can't accidentally match `knowledge_meta`. Left the two optional followups (clear the set on `disableSync`) out — genuinely cosmetic, and the CLI still reports quota regardless.

PR #1192, opened 2026-07-06
feat(sync): show initial-sync progress — "Performing initial sync…" with a live bar

BYK:
## Why
`lore sync enable` printed "Encryption enabled" then went **silent** while the first push uploaded all existing knowledge (500+ rows, paged) — it looked frozen. Give live, per-part feedback.

## What
- **Engine**: thread an optional `onProgress` callback through `syncOnce`/`pushOnce`/`pullOnce`, emitted **per table** (on entry) and **after each page** (so counts move within a big table). No behavior change when absent — the background scheduler passes none.
- **CLI**: `makeSyncProgress` renders a single-line bar that names the current **part** (table) and moves as rows push/pull:
  ```
  Performing initial sync… [██████████░░░░░░] 6/16  pull · knowledge   ↑520 ↓140
  ```
  `runSync` drives it; `lore sync enable` labels it "Performing initial sync…", `lore sync now` "Syncing…". Non-TTY prints the header once (no bar). The bar is cleared before the "Synced: …" summary.

The bar advances one step per (phase, table) visited — push every non-pull-only table, then pull every table — and the ↑/↓ counts move per page for liveness.

## Tests
- `pushOnce`/`pullOnce` emit progress (per table, cumulative counts; one pull event per synced table, emitted **before** the locked/skip guards so a skipped table still advances).
- `makeSyncProgress`: renders + advances the bar on a TTY, names the current table + ↑/↓ counts, clears on `done()`, and prints just the header on a non-TTY.

72 gateway sync/cmd unit tests + 12 engine integration tests pass; typecheck + lint clean.

## Note
Determinate over the known table set (pull totals are unknown via keyset), with per-page ↑/↓ movement for liveness — an honest bar without faking a row total.

### DIFF
--- a/packages/gateway/src/cli/sync-cmd.ts
+++ b/packages/gateway/src/cli/sync-cmd.ts
@@ -13,6 +13,7 @@
 import { randomBytes } from "node:crypto";
 import { createInterface } from "node:readline";
 import { keystore, syncData, getKV } from "@loreai/core";
+import type { SyncProgressFn } from "../sync";
 
 export async function commandSync(
   positionals: string[],
@@ -114,7 +115,7 @@ export async function cmdEnable(
     }
   }
 
-  await runSync();
+  await runSync("Performing initial sync…");
 }
 
 /** Prompt callbacks for the encryption bootstrap (injectable for tests). */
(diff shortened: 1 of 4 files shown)


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — backward-compatible (all existing callers omit `onProgress`; scheduler passes none), step/total math sound (bar fills to exactly 100% by construction; per-page repeats dedupe via the Set), and no behavior change when the callback is absent.

Addressed the review:
- **CONCERN (coverage):** added a test that puts the device in a `locked` state so `knowledge` is genuinely skipped in pull, and asserts its progress event **still fires** — this guards the load-bearing "emit BEFORE the skip guard" placement (mutation-verified: relocating/removing the emit fails it).
- **NIT (defense-in-depth):** wrapped the renderer's `onProgress` body in try/catch so a rendering failure can never abort a sync cycle.
- Left the narrow-terminal wrap-clear nit (purely cosmetic on sub-width terminals).

PR #1193, opened 2026-07-06
fix(sync): classify PGRST schema errors as poison + route deletes through the classifier (#1188)

### LINKED ISSUE #1188: sync: delete-push branch never classifies errors → PGRST/schema errors wedge the table forever
Follow-up from PR #1187 (adversarial review CONCERN).

## Problem
The delete/tombstone branch in `packages/gateway/src/sync.ts` (~line 328) returns `"stop"` on **any** error, never calling `classifyPushError`. Combined with `classifyPushError` mapping everything that isn't `23514` to **transient** (incl. PostgREST **PGRST204** "column not found"), a genuine schema mismatch on a *delete* retries every cycle forever and wedges the table — with no drop-past-cursor escape hatch (the upsert path has one via the `poison` classification at sync.ts:395-410).

PR #1187 fixed the one concrete instance (phantom `content_hash` on `versioned:false` join-table deletes), but the branch remains structurally fragile against future schema drift.

## Fix
1. Classify PostgREST schema errors (`PGRST204` and similar `PGRST*` codes) as **poison** (permanent), not transient, in `classifyPushError`.
2. Route the delete branch through `classifyPushError` so a poison tombstone is recorded + advanced past the cursor (like the upsert poison path) instead of wedging the table.

## Notes
- Keep transient (network/5xx) → `stop` (retry, never lose a write).
- Add a regression test: a delete that hits a permanent schema error is dropped-and-advanced, not retried forever.

BYK:
Closes #1188.

## Problem
The delete/tombstone push swallowed **every** error as transient (`return "stop"`), and `classifyPushError` mapped anything that wasn't `23514` to transient too. So a permanent PostgREST schema error on a **delete** (e.g. PGRST204 "column not found") retried every cycle forever and **wedged the table** — the exact class of bug #1187 fixed at the payload level, but the underlying classification gap remained.

## Fix
- **`classifyPushError`**: PGRST**1xx** (request parsing) + PGRST**2xx** (schema cache: relationship/function/column, incl. PGRST204) are a permanent payload↔schema mismatch for a request we generate → **poison**. PGRST**0xx** (connection) and PGRST**3xx** (JWT/auth) stay **transient**. Exported for direct unit testing.
- **Delete branch**: route the error through `classifyPushError`. **Poison** → record the conflict + advance past it (mirrors the upsert poison path) so one bad tombstone can't wedge the table. **Quota/transient** → keep pending and retry (never lose a delete).

## Tests
- `classifyPushError` boundary table: 1xx/2xx→poison, 0xx/3xx→transient, `23514` quota-vs-check.
- A delete hitting **PGRST204** is recorded (`rejected_unsyncable`) + the cursor **advances** past it (not wedged).
- A delete hitting a **transient** error stays pending (cursor **held**, never dropped).

**Mutation-verified** (removing the PGRST rule fails the poison + unit tests). 55 unit + 12 engine integration tests pass; typecheck + lint clean.

## Scope note
Only the *delete* branch changes behavior. The upsert path already classified + handled poison; this makes the delete path symmetric and hardens both against future schema drift.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — classification boundary sound (PGRST1xx/2xx = permanent payload↔schema mismatch → poison; PGRST0xx connection + PGRST3xx auth stay transient, incl. the genuinely-transient `PGRST002` schema-cache-load), delete poison path mirrors the upsert poison path exactly (no res.pushed++, records conflict, marks sync_state, advances), transient/quota still hold the cursor (never lose a delete), and pull-side symmetry is preserved.

Addressed the main followup: added a test for the **upsert** path's new PGRST204→drop behavior — the shared classifier hardens both paths symmetrically (a permanent schema error on an upsert now drops+records instead of wedging), so it's now covered on both sides.

Noted-and-accepted (non-blocking): a `PGRST204` in the sub-second window after a remote migration (before PostgREST reloads its schema cache) would drop+record one row rather than the old whole-table wedge — acceptable since #1187 already root-fixed the actual trigger (this is a backstop), Supabase auto-reloads the cache on DDL, and the row re-enqueues on the next reconcile. The pre-existing non-PGRST raw-SQLSTATE gap (e.g. 22P02) stays transient by the safe 'never lose a write' default.

Issue #1191, opened 2026-07-06
sync: maintain the top-N most-useful entries under a free-tier cap (ongoing eviction, not just seed)


BYK:
Follow-up from #1189 (value-ranked seed).

## Context
#1189 makes the *initial* `seedOutbox` value-ranked (confidence, then recency), so a fresh `lore sync enable` fills the free-tier cap with the most useful knowledge. But that only orders the **initial fill**. Once a table is at the cap:
- New/updated entries enqueued later just hit quota and pause — even if they're more useful than something already synced.
- So the synced set becomes "the best N *at enable time*", not "the best N *now*".

The product principle (user): free tier should be **limited, not punishing** — it should always hold the most useful entries as the knowledge base grows.

## Proposed design (sketch)
Continuous top-N maintenance with eviction:
1. **Cap discovery**: learn N per (tier, table) from the server quota error (`"N of N rows"`) and cache it; or mirror `plan_limits`.
2. When a push is quota-paused, compare the pending entry's value against the **least-valuable currently-synced** entry (track a local "synced value" index). If the new entry ranks higher, **evict** the lowest (delete the remote row to free a slot) and push the new one.
3. Guard against thrash (hysteresis / min value delta) so a churn of near-equal entries doesn't ping-pong remote rows.
4. Keep it scoped to `knowledge` first; consider entities/aliases later.

## Risks / notes
- Deleting remote rows to free slots must respect the append-only + tombstone model and not race concurrent devices.
- Interacts with the confidence CRDT (value can change post-sync) — eviction decisions should use the materialized confidence.
- Pro tier (50k cap) makes this low-urgency for paying users; it's a free-tier UX quality feature.

## Also
- Entity-graph tables (entities/aliases/relations) also have caps and currently seed in arbitrary order — value/recency-rank their seed too (smaller follow-up).

BYK:
Done — implemented **server-side** (cleaner than the original client-side "synced value index" sketch in the description: eviction fires on INSERT-over-quota, so churn is bounded by write rate rather than the 60s poll rate, the budget can't be client-bypassed, and no client-side value index to keep coherent).

Continuous top-N eviction now covers every capped table:

| PR | Migration | Table(s) | Value signal |
|----|-----------|----------|--------------|
| #1203 | 0013 | `knowledge` (+cascade `knowledge_meta`/`_crdt`) | `knowledge_meta.base_confidence` |
[...]

PR #1194, opened 2026-07-06
feat(sync): value-rank the entity-graph seed (entities/aliases/relations) — #1191a

BYK:
Part of #1191 (the value-ranked-seed half). Extends #1189 from knowledge to the entity graph.

## Why
The push drains the outbox in **seq order** and the server rejects once a table's row cap is hit (free tier: **entities 30**, aliases/relations **300**). The entity-graph tables were seeded in arbitrary storage order, so useful entities could be the ones left behind — the same 'don't punish free users' gap #1189 fixed for knowledge.

## Change
New `seedSelect(table)` value-ranks each non-knowledge seed:
- **entities**: most-referenced first (`COUNT(*)` over `knowledge_entity_refs`), then recency (`updated_at`, `created_at`). A heavily-referenced entity is more useful than a one-off, and the cap is small (30) so the right 30 matter most.
- **entity_relations**: recency (`updated_at`, then `created_at`).
- **entity_aliases**: recency (`created_at` — no `updated_at` column).
- any other table: unordered (no per-row value signal / high enough cap).

No behavior change under the cap (everything still syncs); only *what survives a cap* changes.

## Tests
- entities seed by **reference count** — inserted in the reverse of the expected order **and** with recency as the reverse of ref-count, so neither storage order nor a recency-only sort could reproduce `[eA, eB, eC]`; only ref-count ranking does.
- relations/aliases seed **newest-first**.
- **Mutation-verified** (reverting `seedSelect` to unordered fails both). `beforeEach` now also clears `entity_aliases`/`entity_relations`.

71 `sync-data` tests pass; typecheck + lint clean.

## Follow-up (noted, not here)
`knowledge_meta`/`knowledge_meta_crdt` seeds are still unordered, so under the cap they may not align 1:1 with the value-ranked `knowledge` set (pre-existing since #1189; minor — a mis-synced meta just reads as confidence 1.0 on a peer). Left to the #1191b consistency pass. The bigger continuous **top-N eviction** (#1191b) is also still open.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — `SELECT e.*` returns only entity columns (rowIdOf/hashRowColumns unaffected), the ref-count JOIN can't duplicate rows (subquery groups by entity_id, LEFT JOIN keeps 0-ref entities), index-assisted (COVERING idx_knowledge_entity_refs_entity), behavior-preserving under the cap, and both tests are mutation-verified non-vacuous.

Applied the in-scope followup: added a stable **`id` tiebreak** to all three orderings so re-seeds are fully deterministic on value ties (entities/relations/aliases). Tests still green (ties aren't exercised, so [eA,eB,eC]/[rNew,rOld]/[aNew,aOld] hold).

Tracking the noted `knowledge_meta`/`knowledge_meta_crdt` alignment (seed the meta in the same value order as `knowledge` so surviving meta matches surviving knowledge under a cap) as part of the #1191b consistency pass — it degrades gracefully today (missing meta → confidence 1.0 via COALESCE).

PR #1196, opened 2026-07-06
feat(sync): align the knowledge_meta seed with knowledge's value order

BYK:
Follow-up to #1189 / #1194 (flagged in both reviews). Completes the value-ranked-seed coherence for the confidence register.

## Why
`knowledge_meta` shares knowledge's free-tier row cap (**500**), but it was seeded in **arbitrary storage order**. So under the cap the surviving meta set wouldn't match the surviving (confidence-ranked) `knowledge` set — a synced entry could reach a peer **without its confidence** (reading as the default 1.0 via the `knowledge_current` COALESCE).

## Change
Seed `knowledge_meta` in the **same value order** as the knowledge seed:
```sql
ORDER BY confidence DESC, COALESCE(last_reinforced_at, updated_at) DESC, logical_id
```
so the top-N meta rows line up with the top-N knowledge rows. Deleted entries are typically low-confidence (`pruneDeadEntries`/consolidation) so they sort last and don't steal live slots — a plain order (no JOIN) aligns in practice with no behavior change to *which* rows seed. `knowledge_meta_crdt` has a 10× cap (5000) that won't bind for a single device, so it stays default/unordered.

## Tests
- `knowledge_meta` seeds by confidence — created low→mid→high so storage order is the **reverse** of the expected `[high, mid, low]`; **mutation-verified** (removing the case → unordered → fails).
- `beforeEach` now also clears `knowledge_meta`/`knowledge_meta_crdt` so the register doesn't leak across cases.

72 `sync-data` tests pass; typecheck + lint clean.

## Remaining
This closes the seed-coherence half of #1191. The bigger continuous **top-N eviction** (#1191b — keep the best-N as the base grows past the cap, not just at seed time) is still open and wants a design pass first.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — order mirrors the knowledge seed exactly (confidence → recency, logical_id tiebreak), ordering by the materialized `confidence` (not the synced `base_confidence`) is correct since the knowledge seed ranks by the same materialized value, and the beforeEach cleanup is FK-safe. Mutation-verified in an isolated worktree (removing the case → [low,mid,high] insertion order → fails).

Filed the one noted follow-up (1191): seed meta **live-only** via `JOIN knowledge_current` — a row-set change (out of scope here) that both fully closes the deleted-high-confidence-entry slot-steal edge and cleans up orphaned meta on the remote.

BYK:
Correction: the meta-live-only follow-up is filed as https://github.com/BYK/loreai/issues/1197 (my earlier comment linked the wrong number).

PR #1198, opened 2026-07-06
feat(sync): seed knowledge_meta for live entries only (JOIN knowledge_current)

### LINKED ISSUE #1197: sync: seed knowledge_meta/crdt live-only (JOIN knowledge_current) to fully align with knowledge + drop orphaned meta
Follow-up from PR #1196 (adversarial review).

#1196 aligned the `knowledge_meta` seed **order** with the value-ranked `knowledge` seed. But `knowledge_meta` still seeds **all** register rows, including rows for **deleted** logical entries (`remove()` intentionally keeps the meta register row — a synced convergent register, excluded from `LOGICAL_ID_BOOKKEEPING_TABLES`).

## Residual gaps
1. **Deleted high-confidence entry can steal a live slot.** `pruneDeadEntries` deletes low-confidence entries (safe), but **curator consolidation** can delete a *high-confidence* entry. Its lingering meta sorts to the top and, if the 500 cap binds, evicts a *live* entry's meta → that entry reads as default confidence 1.0 on a peer (self-heals on next reinforce/update). Narrow, bounded, non-blocking.
2. **Orphaned meta on the remote.** Meta (and crdt) rows for deleted entries sync even though no live `knowledge` row exists — pre-existing since the register was added.

## Fix
Seed `knowledge_meta` (and consider `knowledge_meta_crdt`) restricted to live entries:
```sql
SELECT m.* FROM knowledge_meta m
  JOIN knowledge k ON k.logical_id = m.logical_id AND k.is_current = 1 AND k.is_deleted = 0
 ORDER BY m.confidence DESC, COALESCE(m.last_reinforced_at, m.updated_at) DESC, m.logical_id
```
Row-**set** change (stops seeding deleted/orphaned meta) → needs its own test and a note on reaping existing orphaned remote meta (ties into #909 reaper). Relates to #1191.

BYK:
Closes #1197. Follow-up to #1196.

## Why
#1196 aligned the `knowledge_meta` seed **order** with the value-ranked `knowledge` seed, but still seeded **all** register rows — including rows for **deleted** logical entries (`remove()` intentionally keeps the meta register row). Two gaps:
- A curator-consolidation-deleted **high-confidence** entry's lingering meta sorted to the top and, under the shared **500-row** cap, could **evict a live entry's meta** (→ that entry reads as default confidence 1.0 on a peer).
- Meta for deleted entries was pushed to the remote as **orphaned** rows (no live knowledge).

## Change
Restrict the `knowledge_meta` seed to **live** entries:
```sql
SELECT m.* FROM knowledge_meta m
  JOIN knowledge k ON k.logical_id = m.logical_id AND k.is_current = 1 AND k.is_deleted = 0
 ORDER BY m.confidence DESC, COALESCE(m.last_reinforced_at, m.updated_at) DESC, m.logical_id
```
Mirrors `knowledge_current`, same value order as the knowledge seed. `knowledge_meta_crdt` stays default (its 10× cap won't bind for a single device; orphaned crdt rows are inert — noted in #1197 as a further option).

This is a row-**set** change (stops seeding deleted/orphaned meta), so it ships separately from #1196's order-only change.

## Tests
- **deleted-entry meta is NOT seeded** (live-only) — mutation-verified (reverting to the non-JOIN seed re-includes the deleted entry's meta and fails).
- the existing confidence-order test still holds.

73 `sync-data` tests pass; typecheck + lint clean.

## Note
Existing orphaned meta already on the remote (from before this change) isn't reaped here — that's the server-side reaper (#909).


BYK:
Adversarial review: **SHIP**, no BLOCKERs — the JOIN can't duplicate (partial-unique idx_knowledge_one_current → ≤1 current per logical_id), `SELECT m.*` is column-identical for rowId/hash, and the 'no meta-delete propagation' design is intentional + consistent across seed and reconcile (remove() keeps the register row → reconcile's NOT EXISTS won't tombstone). Predicate byte-matches `knowledge_current`. Mutation-verified. Applied the one worthwhile NIT (assert the deleted entry's meta row still physically exists so the test can't pass for the wrong reason). Orphaned pre-existing remote meta is benign and is reaper #909's job.

PR #1203, opened 2026-07-07
feat(supabase): server-side continuous top-N eviction for knowledge (#1191b PR1)

BYK:
PR1 of #1191b (server-side first, per the cost analysis). Client latency-polish is PR2.

## Why
The value-ranked seed makes the best-N sync at `enable` time, but once a table is at the free-tier row cap, new writes **pause** — so an active free user's freshest, high-confidence knowledge stops backing up while older, lower-value rows keep their slots. This makes the server **keep the best-N as the base grows**.

## Why server-side (the cost answer)
Eviction fires on **INSERT-over-quota**, so churn is bounded by the **write rate, not the 60s poll rate** — it structurally cannot thrash the way a client poll-loop could, needs no extra round-trips, and the churn budget **can't be bypassed by a client**.

## Design (migration 0013)
- **Value = `knowledge_meta.base_confidence`** (already synced). Ranked off that rather than a denormalized column on `knowledge` — such a column would track confidence, so every reinforce/decay would re-sync the **big content row** (the exact churn the knowledge_meta split removed). New entries start ~1.0, so the incoming row is essentially never the lowest → ranking existing rows is exact in practice.
- **HARD delete, not a tombstone** — local SQLite is the source of truth, so an evicted remote row is a **cache eviction (never data loss)** and must not propagate as a deletion to peers. **Cascades** to the entry's `knowledge_meta` + `knowledge_meta_crdt` by `logical_id` → trio stays aligned, their slots free too, so the paired meta/crdt INSERTs (pushed after knowledge) fit — **no push-order change, no new column**.
- **Cost guard:** per-scope hourly **circuit breaker** (`sync_eviction_budget`, RLS-locked, definer-only). Over the budget → falls back to raising (pause). Budget = `lore.eviction_budget` GUC (default 200; tests lower it).
- **Scope (PR1):** knowledge, **row cap only**. Byte-cap overflow still raises; entities/aliases/relations still raise until they get a self-contained rank (the `sync_rank` column — PR2). Pro is uncapped → eviction never runs there.

## Tests (Tier-1 real Postgres + RLS/triggers)
- Evicts the **lowest-confidence** entry + cascades its meta; net count stays at cap.
- Cascades the **crdt** row too.
- **Circuit breaker** pauses after the budget.
- Eviction is **knowledge-only** (entities still raise).
- **Byte-cap** overflow still raises (row eviction only).

**Mutation-verified**: disabling the eviction branch fails the three eviction-dependent tests (the raise-behavior tests still pass). 51 integration tests pass; typecheck + lint clean.

## Follow-ups
- PR2: client-side latency polish + the `sync_rank` column that extends exact eviction to **entities/aliases/relations**.
- Byte-cap eviction (evict-to-fit-bytes) and tombstone-first eviction are noted in the design doc.


BYK:
Hardening (pre-review): moved the circuit-breaker cap from a session GUC (`lore.eviction_budget`) to an **RLS-locked `sync_config` table**. A session GUC is client-`SET`-table, which would let a client raise its own eviction budget and force runaway churn — defeating the whole point of server-side enforcement. Added a test that `authenticated` gets `42501` on `sync_config` (read + write). 52 integration tests pass.

BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — machine-diffed against 0010 (all prior quota behavior byte-identical; eviction is the only addition), scope-isolated (victim query pinned to `new.scope_id`, cross-tenant deletion impossible), transactionally atomic (a later byte-raise rolls back the eviction+budget increment), breaker race-free under the existing advisory lock, hard-delete produces no peer-propagating tombstone, and not client-bypassable. Both SHOULD-FIX items applied:

- **F1** — gated eviction to `user_tier = 'free'` (pro `knowledge` is capped at 50000, so it *would* have evicted; a paying user now gets a recoverable **pause** signal instead of silent remote eviction). Comment corrected. Added a discriminating test (a capped pro pauses, evicts nothing).
- **F2** — added `revoke all on sync_config, sync_eviction_budget from anon, authenticated` (mirrors 0007's `user_table_usage`), so the `42501` assertion holds against **real Supabase**, not just the shim's missing default GRANT.

53 integration tests pass; mutation-verified (disabling eviction fails 3 tests; removing the free-gate fails the pro test). typecheck + lint clean.

BYK:
@sentry Reviewed — the suggested `coalesce(..., 0.0)` would introduce a **worse** bug, so keeping `1.0` (added a comment explaining why).

Meta-less is the **normal transient state during a push**: the client pushes all `knowledge` rows *before* their `knowledge_meta` (push order within a cycle). So a just-admitted entry is briefly meta-less on the remote. With `0.0`, the next at-cap INSERT would rank that fresh entry **lowest and evict it before its meta arrives** — evicting the newest, highest-value data (and fresh entries would evict each other during a multi-entry push at cap). `1.0` protects them, and is consistent with the `knowledge_current` view (`COALESCE(confidence, 1.0)` — meta-less = default 1.0 system-wide).

Genuinely-orphaned rows (permanent meta loss) are rare, bounded by the entry count, and belong to the reaper (#909) / A2 compaction — not the eviction path. No code change; annotated the invariant at the query.

PR #1206, opened 2026-07-07
feat(supabase): server-side recency eviction for entity_aliases + entity_relations (#1191b PR2a)

BYK:
PR2a of #1191b's entity-graph extension. Stacked on the merged PR1 (#1203).

## Why
After PR1 (knowledge eviction), `entity_aliases` (free cap 300) and `entity_relations` (300) still **hard-pause** at the row cap. This extends the same relief valve to them — by **recency**.

## Design (migration 0014)
On a free-tier INSERT that would exceed the row cap, evict the **OLDEST** live row for that scope (`updated_at ASC, id ASC`), then admit the incoming.
- Recency is **self-contained** on the row (`updated_at`), and the incoming row is always the **newest** → it's never its own victim and always survives → **no incoming-value guard needed** (unlike knowledge/confidence).
- **Leaf tables** (nothing references an alias/relation) → plain hard DELETE, **no cascade**.
- Everything else inherited from 0013 **verbatim**: FREE-tier only, ROW cap only (byte-cap still raises), hard delete (cache eviction, not a peer-propagating tombstone — local is source of truth), and the per-scope hourly **circuit breaker** (`sync_eviction_budget` + `sync_config`). The only change is the eviction block: it now fires for `{knowledge, entity_aliases, entity_relations}` and branches per-table.

## Not in scope
- `entities` → **PR2b**: its value (ref-count) isn't self-contained and a *new* entity starts low-value, so it needs the synced **`sync_rank`** column + an incoming-value guard (cross-cutting: local + client + remote).

## Tests (Tier-1 real Postgres + RLS/triggers)
- Evicts the **oldest relation by recency** — the oldest row is given a **higher id**, so the test isolates `updated_at` ordering from id order.
- Evicts the **oldest alias**.
- Existing knowledge-eviction + breaker + pro-gate tests still pass against the 0014 function (regression coverage).

**Mutation-verified**: restricting eviction back to knowledge-only fails both new tests. 55 integration tests pass; lint + typecheck clean.


BYK:
Review: **SHIP**, no BLOCKERs — knowledge path proven byte-identical to 0013 (behavior-preserving refactor of the shared eviction bookkeeping); entity path is scope-safe, `%I` injection-safe (table name from the fixed IN-set), cascade-complete (true leaf tables — zero FKs reference them), correctly accounted, and free-tier/row-cap/budget gated. Applied the two non-blocking doc fixes:
- Reworded the eviction comments: the load-bearing invariant is the **BEFORE-INSERT timing** (incoming isn't in the table yet → never its own victim), not "incoming is always newest" (a client can push an INSERT with an arbitrary `updated_at`, so that framing was a false premise).
- Fixed the stale `entities` test comment (eviction now covers aliases/relations; only entities awaits `sync_rank` in PR2b).

Noted (non-blocking): the per-scope circuit breaker is aggregate across evicting tables (cross-table starvation possible but falls through to the safe pause — acceptable as a cost backstop).

BYK:
@sentry This finding is benign and the suggested fix would **break sync** — keeping `updated_at` client-set on INSERT (intentionally):

1. **No quota bypass.** Eviction removes exactly one row (the oldest) to admit one, so the row count *stays at the cap* regardless of `updated_at` values — a client cannot exceed the cap. `updated_at` only influences *which* of the client's **own** rows is evicted.
2. **Scope-isolated.** The victim query is pinned to `new.scope_id = auth.uid()`, so a client can only reprioritize eviction among **its own** rows (its own remote cache) — no cross-tenant effect, and local SQLite is the source of truth so nothing is lost.
3. **The suggested fix is harmful.** Server-stamping `updated_at` on INSERT would clobber the client's timestamp, which is **load-bearing** for the `(updated_at, id)` keyset **pull cursor** and last-writer-wins conflict resolution. `sync_set_updated_at` is deliberately BEFORE **UPDATE** only.

This is also a pre-existing property (clients have always controlled INSERT `updated_at`); eviction merely reads the field. No code change; the invariant comment already documents that the real guard is the BEFORE-INSERT timing (incoming is never its own victim), not the timestamp value.

PR #1207, opened 2026-07-07
feat(sync): server-side entity eviction via synced sync_rank (#1191b PR2b)

BYK:
PR2b of #1191b's entity-graph extension — the cross-cutting one. Completes the relief valve (0013 knowledge, 0014 aliases/relations, this = entities).

## Why entities need a column (not a JOIN)
Knowledge ranks by confidence, aliases/relations by recency — both **self-contained** on the row. An entity's value is its **ref-count**, which is (1) a JOIN, and (2) **unknown at INSERT**: a new entity starts at ~0 refs and its `knowledge_entity_refs` aren't synced yet, so the server can't value the *incoming* row via a JOIN. And a new entity is genuinely **low-value** (unlike high-confidence new knowledge), so "evict lowest existing, admit" would let a 0-ref newcomer displace a well-referenced entity. So the client maintains a synced **`sync_rank`** column (= ref-count) on the entity row, and the server evicts **with an incoming-value guard**.

## Design
**Local (v65):**
- `entities.sync_rank INTEGER NOT NULL DEFAULT 0`, backfilled from the ref count.
- Maintained in **app code** (`recomputeEntityRank`) at `linkKnowledge` / `unlinkKnowledge` / `syncEntityRefs` (before∪after affected set) / `mergeEntities` — **not a trigger**, because `recursive_triggers` is OFF so a trigger on `knowledge_entity_refs` wouldn't fire the entities sync-capture/FTS triggers (the rank change would never sync).
- `sync_rank` added to `entities` syncColumns (entities is versioned → the hash change drives the re-push).

**Remote (0015):**
- `entities.sync_rank` column + eviction branch: evict the lowest `sync_rank` **only if** `new.sync_rank > floor` (guard). A sub-floor newcomer **pauses** until its rank rises.
- Cascades the victim's `entity_aliases` / `knowledge_entity_refs` / `entity_relations` (meaningless without the entity; frees their slots). HARD deletes, FREE-tier + row-cap + circuit-breaker gated as 0013/0014.

**Ordering:** the remote column (0015) auto-deploys on merge *before* any client ships the new syncColumns → no unknown-column push error.

## Tests
- **Local**: `sync_rank` tracks link/unlink (incl. INSERT-OR-IGNORE idempotence) and `syncEntityRefs` gain+loss (before∪after recompute).
- **Integration (real Postgres)**: evicts the lowest-rank entity + cascades its alias/ref/relation; the **value guard** pauses a sub-floor newcomer (nothing evicted).

**Mutation-verified**: dropping the entities branch fails the eviction test; weakening the guard fails the guard test. Core 2686 pass, 57 integration pass, typecheck + lint clean.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — remote migration proven a clean/minimal/idempotent extension of 0014 (byte-diffed), eviction branch + incoming-value guard (strict `>`, tie→pause) + cascade + scope-isolation + breaker gating all verified, and both integration tests discriminate (drop branch → eviction test fails; weaken guard → guard test fails; the tie boundary is covered by the pre-existing all-rank-0 entities test).

**Fixed F-1 (the review's SHOULD-FIX) in-PR** rather than deferring — it undermined the feature's core value signal on a common path. `sync_rank` was maintained on ref *add* but not on knowledge *deletion*, leaving stale-high ranks that never self-heal (hash unchanged → never re-pushes → remote keeps the inflated value → evicts higher-value entities instead):
- `ltm.remove()` (common curator path): capture the entities losing a ref **before** the explicit ref-delete, then `recomputeEntityRank` each.
- `clearProject`/`deleteProject`/`clearKnowledge` (bulk FK-cascade purges): `resyncStaleEntityRanks()` after the knowledge delete — repairs **only drifted rows** (correlated `WHERE sync_rank <> live-count`), so unchanged entities don't churn the outbox.
- Added tests: `ltm.remove` recompute + `resyncStaleEntityRanks` drift repair.

Non-blocking NITs noted (stale test comment at the all-rank-0 entities test; `recoverMissingObjects` entities recreate lacks sync_rank — same latent pattern as every ALTER-added column). Core 2688 pass, typecheck + lint clean.

BYK:
CI fix: adding `sync_rank` to entities syncColumns broke the hand-written sync engine mock in `sync.test.ts` (its `REMOTE_COLUMNS` phantom-column check rejected the new column). Added `sync_rank` to the mock's entities column set — 56 sync.test.ts tests + full gateway suite (2722) green again.

BYK:
Addressed the Sentry finding (the review's NIT-5): added `sync_rank INTEGER NOT NULL DEFAULT 0` to the `recoverMissingObjects` entities `CREATE TABLE` — that definition is meant to be current (it already includes the ALTER-added `embedding`), so a version-65 DB that lost its entities table would otherwise self-heal *without* `sync_rank` and, since the v65 ALTER won't re-run at version 65, the column would be permanently missing. Also fixed the now-stale NIT-2 comment: the `entities` cap test now correctly documents that it exercises the **strict-`>` tie guard** (all default rank 0 → 0 > 0 false → pause), not "entities aren't evictable." db.test 109 + 57 integration pass; typecheck + lint clean.

PR #1210, opened 2026-07-07
fix(sync): eviction polish — knowledge_meta cap headroom + relation cascade only when both endpoints gone (#1191b)

BYK:
Two fixes surfaced by running `lore sync now` against production after the eviction PRs (#1203/#1206/#1207) landed.

## FIX 1 — the misleading "knowledge_meta limit" message
`knowledge` (free cap 500) *churns silently* via eviction (evict-lowest-confidence, admit-new — no guard). But `knowledge_meta` (also 500) has **no independent eviction** — it only freed slots via knowledge's cascade. So a user with >500 live knowledge sees the overflow's *metas* quota-block, surfacing **"reached your knowledge_meta limit"** — which reads like a metadata problem, not "you have more knowledge than the free tier syncs."

**Fix:** give free `knowledge_meta` the same **10× headroom `knowledge_meta_crdt` already has** (500→**5000** rows, 512KB→4MB). Metas for the ~500 synced knowledge always fit — which also makes confidence-based eviction *accurate* (no meta-less `COALESCE(1.0)` over-protection at steady state). A few orphan metas for the knowledge-overflow are inert (the eviction JOIN never reaches them) and are the reaper's (#909) job.

## FIX 2 — cascade a relation only when BOTH endpoints are gone
0015 cascade-deleted **every** `entity_relations` row touching an evicted entity (`entity_a=victim OR entity_b=victim`). Because the client marks a relation synced after its first push, it **never re-pushes** a remotely-cascaded relation — so relations touching a *still-synced* entity were **permanently lost** from the remote (local↔remote divergence), not just churned.

**Fix:** keep a relation while its **other endpoint is still a live synced entity** (that entity's edge is preserved); delete only fully-orphaned relations (both endpoints absent/soft-deleted). `entity_aliases` and `knowledge_entity_refs` still cascade unconditionally — they belong to the single evicted entity.

## Migration 0016
`plan_limits` bump + `create or replace enforce_row_quota` (reproduces 0015 byte-for-byte except the relation cascade + adds the `not exists` other-endpoint guard).

## Tests (integration, real Postgres)
- Cap-headroom value (free knowledge_meta = 5000).
- Both-gone cascade: relation to a surviving endpoint **kept**, relation to an absent endpoint **dropped**.
- Updated the 0015 cascade test — `rel1` (e-lo↔e-hi) is now **kept** since e-hi survives.
- Both-gone **mutation-verified**: reverting to the OR-cascade fails both relevant tests. 59 integration pass, typecheck + lint clean.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — 0016's `enforce_row_quota` byte-diffed clean against 0015 (only the relation cascade changed), both-gone logic + cascade ordering + maintain_usage/recursion + idempotency + orphan-meta inertness all verified, and the tests discriminate (revert to OR-cascade fails both relation tests).

**Folded in review follow-up #1 (self-relation hardening):** added `and e.id <> victim` to the both-gone `not exists` so a `(victim, victim)` self-relation is correctly treated as fully-orphaned and deleted (previously it'd survive as an inert orphan, since the victim is still live at cascade time). The client already guards against self-relations (`entities.ts:1399`), so this is defense-in-depth for direct-PostgREST writers. Added a discriminating test (mutation-verified: removing the guard fails it).

Follow-up #2 (a live-client relation to a remotely-soft-deleted/never-synced endpoint is dropped and not re-pushed) is a cosmetic bounded-cache residual — leaving as noted, covered by the reaper (#909). 60 integration pass.

PR #1213, opened 2026-07-07
feat(core): local knowledge version compaction + de-CASCADE ref FKs (#909)

BYK:
A2 sub-PR 4 (client half of #909). Bounds local storage by pruning superseded knowledge version rows, keeping a bounded recent window per logical entry.

## The prereq gap this closes
#913 made the **outbox** anchor-free, but the ref FKs were a *separate* anchor dependency it didn't touch:
```
knowledge_refs.from_id/to_id       → knowledge(id) ON DELETE CASCADE
knowledge_entity_refs.knowledge_id → knowledge(id) ON DELETE CASCADE
```
Refs key on `logical_id`, which `== the v1 anchor row's id`. So pruning the anchor would **cascade-delete a live entry's wiki-refs + entity-links** — silent data loss. (Confirmed with the user that de-CASCADE is safe: the local FK never guarded against malicious clients — that's all remote RLS/quota/CHECK — and a server-side sweep stays fully possible.)

## What changed
- **Migration v66** — recreate both ref tables **without** the `knowledge(id)` FK (leaf tables; orphan-filtered copy so the *retained* `entity_id→entities CASCADE` on `knowledge_entity_refs` can't fail the migration). Aligns local with the remote, which never had these FKs. `recoverMissingObjects` + schema-version assertions updated (65→66).
- **`clearProject` / `deleteProject` / `clearKnowledge`** — purge `knowledge_entity_refs` + `knowledge_refs` explicitly by logical_id (they used to lean on the CASCADE).
- **`compactKnowledgeVersions()`** — under `withSyncApplying` (local-only, **no outbox rows**): keep the head (current/death-cert) + the 2 most-recent superseded per `logical_id`, drop the rest and anything **>60 days** — including the v1 anchor, now safe. Wired into the idle global sweep (batched, re-drains when full).

## Why the anchor prune is safe
Refs resolve by `logical_id` everywhere; the current version always carries it, so a ref still resolves to the live entry after the anchor row is gone, and `cleanDeadRefs` GCs a ref whose logical_id has no live version. The FTS delete trigger is `WHEN old.is_current=1`-guarded, so pruning superseded rows never touches FTS.

## Tests
- Compaction: count cap, age cap (>60d within the window), head-never-touched, **anchor-prune-safe** (entity + wiki refs survive & resolve), **sync-suppression** (0 outbox rows).
- Ref cleanup for all 3 delete paths after the FK removal.
- Full core (2697) + gateway (2722) suites green; typecheck + lint clean.

## Deferred
Server-side reaper (remote orphan refs + the orphan `knowledge_meta` rows surfaced in 0016 testing) stays a server task per #909.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS** (5 mutations killed against their guarding tests; all 10 scrutiny points PASS except one). Addressed the one pre-merge finding + both NITs:

**Finding #2 (migration atomicity) — FIXED.** The v66 recreate ran via `database.exec()` (CREATE _new / copy / DROP / RENAME), which is **not** atomic — a crash mid-recreate would leave a lingering `_new` table that boot-loops the retry (`table already exists` ≠ the `duplicate column name` the runner tolerates), uniquely violating the codebase's migration retry-safety invariant. Refactored into an atomic JS `SAVEPOINT` step (`applyRefFkDrop`, special-cased by `REF_FK_DROP_MIGRATION_INDEX`) — matches the `applySessionRollup` precedent, so a crash rolls the whole recreate back and the retry starts from the original tables (no boot-loop, no data loss).

**NIT — cross-project wiki-ref coverage:** added a test — clearing project A drops the A→B wiki ref (matching the old CASCADE) but **preserves project B's entry** (the DELETE is project-scoped).

**NIT — stale `refs.test.ts` comment:** the `PRAGMA foreign_keys OFF/ON` toggle was a no-op now that the `knowledge_refs` FK is gone — removed it, comment updated to reflect the dead-ref state comes from the (absent) FK.

Full core suite 2698 pass, typecheck + lint clean. All other points (head-safety, keep-2 + age-cap, `withSyncApplying` suppression, FTS-safe superseded deletes, faithful CASCADE reproduction) were verified + mutation-guarded.

BYK:
Addressed the Sentry finding (`clearKnowledge` multi-DELETE without a transaction): wrapped the whole purge sequence (transfers + logical_id bookkeeping + injections + contradictions + the two new ref deletes + the knowledge rows) in `BEGIN IMMEDIATE`/`COMMIT` with `ROLLBACK` on error — now matching `clearProject`/`deleteProject`. This closes the pre-existing gap that my two added ref-deletes had lengthened. `resyncStaleEntityRanks` + vec-orphan reclaim + cache/`.lore.md` regen stay post-commit (idempotent follow-ups). Full core suite 2698 pass, typecheck + lint clean.

PR #1215, opened 2026-07-07
fix(gateway): don't let one orphan pull row abort the whole sync

BYK:
## The crash
A full initial sync on a fresh device crashed with `FOREIGN KEY constraint failed` (SQLite errcode 787) while pulling `entity_aliases`, aborting the **entire** multi-table sync:
```
Performing initial sync… [█████████████░░░] 16/19  pull · entity_aliases  ↓501
Error: FOREIGN KEY constraint failed  { code: 'ERR_SQLITE_ERROR', errcode: 787 }
```

## Root cause
The pull loop (sync.ts) applies each remote row with **local** FKs enforced, but only catches `EncryptedContentUnavailable` — **any other error re-throws and kills the whole cycle**.

The offending row is an **orphan on the FK-less remote**: `entity_aliases` / `knowledge_entity_refs` / `entity_relations` carry a local `entity_id → entities` FK, but the remote has **no FKs**, so an alias/ref can legitimately exist for an entity that was never admitted under the free-tier entity cap (30). A fresh client pulls that orphan and the local FK rejects it (787). This is the same "local FK vs FK-less remote" mismatch #1213 addressed for knowledge refs — here on the entity graph, exposed by a full re-pull.

## Fix
Mirror the push-side **poison** handling: wrap each per-row apply (in the primary pull loop **and** `drainTimestamp`) so a **local constraint violation** (`isSqliteConstraintError`: `SQLITE_CONSTRAINT_*` or `ERR_SQLITE_ERROR` "constraint failed") **skips that one row** — recording it to `sync_conflicts` and advancing the cursor past it — instead of aborting the sync. `EncryptedContentUnavailable` still defers the whole table (re-pulls when the key is available); any other (unexpected) error still propagates.

## Test
A pull carrying a valid alias + an orphan alias (absent entity) now completes without throwing — the valid alias applies, the orphan is skipped and recorded as a conflict (`r.conflicts >= 1`). Full sync suite (62) green; typecheck + lint clean.

## Deferred
The remote still *accumulates* these orphans (aliases/refs pushed for entities the cap rejected). Physically GC'ing them is the server-side reaper's job (#909) — same as the orphan `knowledge_meta` rows surfaced in 0016 testing. This PR makes the client **resilient** to them.

## Note on the companion report (session/`sync.enabled` not persisting)
Both the login session and the `sync.enabled` flag live in the same `team_config` table in `lore.db` — both being absent means the nightly opened a different/fresh `lore.db` (new device / data dir), which is also what triggered the full re-pull. That's per-device local state (intentionally not synced), not a persistence bug. Tracking separately.


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs (primary-loop guard mutation-proven; full gateway suite 2740 pass). Addressed the follow-ups:

- **drainTimestamp coverage gap** (reviewer's Mutation C survived — the drain path was unguarded): added a regression test that seeds >PAGE (200) aliases on one exact `updated_at` (forcing the primary page to hand off to `drainTimestamp`) with an orphan sorted into the drain batch. Confirms the drain path skips the orphan + completes without throwing.
- **NIT:** fixed the `isSqliteConstraintError` comment (`better-sqlite3` → `bun:sqlite`, the actual second driver).

Two low-severity gaps noted for the reaper follow-up (#909), no code change:
1. A skipped alias whose entity is later admitted won't re-pull until its remote `updated_at` bumps or the cursor resets (inherent to the lossy free-tier cap).
2. A full cursor reset would re-record each orphan into `sync_conflicts` (bounded; matches the existing conflict-path INSERT convention).

BYK:
Re: Sentry `isSqliteConstraintError` "too broad" (catches UNIQUE, not just FK) — **verified reachable but keeping the broad catch by design; tracked as   .**

The suggested narrowing (FK-only) is the wrong fix: the pull loop previously caught **only** `EncryptedContentUnavailable`, so a UNIQUE-violating pulled row would **also crash the entire multi-table sync** — the exact bug class this PR fixes. Narrowing would leave UNIQUE (and CHECK/NOT NULL) as latent whole-sync aborts.

- `applyRemoteUpsert` uses `ON CONFLICT(id)` only, so a secondary `UNIQUE(alias_type, alias_value)` genuinely throws → confirmed reachable.
- The row is **not silently dropped**: it's recorded to `sync_conflicts` (diagnosable + recoverable), mirroring the push-side poison convention.
- vs the status quo (crash), this is strictly better for **both** FK and UNIQUE.

The remaining valid part of the concern — a UNIQUE conflict can leave devices divergent — is a separate **conflict-resolution** enhancement (convergent remote-wins/tiebreak), filed as   . Not a regression from this PR.

BYK:
(Correcting the blank link above — the follow-up is **#1217**: convergent conflict resolution for pulled rows that violate a local-only UNIQUE.)

Issue #1217, opened 2026-07-07
sync: convergent conflict resolution for pulled rows that violate a local-only UNIQUE


BYK:
Split out from PR #1215 (Sentry MEDIUM finding).

## Context
PR #1215 makes the pull-apply resilient: a pulled row that violates a **local** constraint the FK-less remote doesn't enforce is skipped + recorded to `sync_conflicts` + the cursor advances, instead of aborting the whole multi-table sync. Two constraint classes:
- **FK orphan** (`entity_aliases`/`knowledge_entity_refs`/`entity_relations`.`entity_id → entities`): parent entity never admitted under the free-tier cap. Skipping is **correct** — the row genuinely can't apply.
- **Secondary UNIQUE** (`entity_aliases UNIQUE(alias_type, alias_value)`): reachable because `applyRemoteUpsert` uses `ON CONFLICT(id)` only, so a NEW remote alias id carrying a duplicate `(alias_type, alias_value)` throws. Skipping avoids the crash but can leave devices **divergent**.

## Why not narrow #1215 to FK-only
That re-introduces the crash class for a UNIQUE-violating pulled row (the pull loop previously caught only `EncryptedContentUnavailable`, so it would abort the whole sync). #1215 is strictly better (crash → logged skip) for **both** classes.

## Ask
Add deterministic **remote-wins-or-tiebreak** resolution for a pulled row colliding with a local row on a local-only UNIQUE: e.g. on `entity_aliases`, delete the local row holding the conflicting `(alias_type, alias_value)` (recording it) then apply the remote — or a stable tiebreak (lowest entity_id wins) applied symmetrically so both devices converge. Reuse `sync_conflicts`. Low severity (aliases are fuzzy lookup aids), but worth converging.

PR #1222, opened 2026-07-08
fix(gateway): don't flood WARN logs on expected orphan pull-skips

BYK:
Follow-up to #1215. Reported after testing the nightly: sync no longer crashes (good!), but `sync now` produced a flood of:
```
[WARN ] sync: pull knowledge_entity_refs skipped orphan row <id>: FOREIGN KEY constraint failed
```
one per skipped row (hundreds, on a fresh device whose knowledge graph references entities beyond the free-tier 30-entity cap).

## Root cause
#1215's per-row skip logged via `log.notice`, which is **not** debug-gated, writes `[WARN]` to the log file, **and** reports to the Sentry sink at warning severity. So every expected orphan skip = a visible WARN + a Sentry warning event.

## Fix
FK orphans (parent entity not synced under the cap) are **expected + high-volume**, so:
- count them in a new `SyncResult.skipped` — surfaced once in the `lore sync now` summary ("…, N skipped (records not present locally — e.g. entities beyond your plan's sync cap)") — and still record each to `sync_conflicts` (diagnosable/recoverable);
- **don't log per row** for the FK case; emit **one debug-gated (`log.info`) summary per table** instead;
- a rarer **non-FK** constraint skip (UNIQUE/CHECK/NOT NULL) stays **visible** via `log.notice` — those are unexpected and low-volume.

## Also
- `.gitignore`: add `.cache/` — the transformers.js model-download cache lands in `packages/*/.cache` (a ~137MB onnx + tokenizer); this keeps it from being committed by accident.

## Tests
- New assertion: an FK-orphan pull-skip emits **no** per-row `log.notice`.
- The two existing orphan-skip tests now assert `res.skipped` (not `res.conflicts`) — orphan skips are semantically distinct from LWW conflicts.
- Full sync suites (63) green; typecheck + lint clean (the 2 lint errors on `references.ts`/`ltm-metadata.test.ts` are pre-existing on main, untouched here).

Note: the remote still *accumulates* these orphans (refs/aliases pushed for entities the cap rejected) — physically GC'ing them stays the server-side reaper's job (#909); convergent UNIQUE-conflict resolution is #1217.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ dist-vendor/
 .sea-staging/
 .node-cache/
 local_cache/
+.cache/
 *.tgz
 coverage/
 *.db
--- a/packages/gateway/src/cli/sync-cmd.ts
+++ b/packages/gateway/src/cli/sync-cmd.ts
@@ -471,6 +471,9 @@ async function runSync(header = "Syncing…"): Promise<void> {
     `Synced: pushed ${r.pushed}, pulled ${r.pulled}` +
       (r.conflicts
         ? `, ${r.conflicts} conflict(s) resolved (remote wins)`
+        : "") +
+      (r.skipped
+        ? `, ${r.skipped} skipped (couldn't apply locally — mostly records referencing data beyond your plan's sync cap)`
         : ""),
   );
   if (r.quotaHit) {
--- a/packages/gateway/src/sync.ts
+++ b/packages/gateway/src/sync.ts
@@ -33,6 +33,12 @@ export interface SyncResult {
   pushed: number;
   pulled: number;
   conflicts: number;
+  /**
+   * Pulled rows skipped because they violate a LOCAL constraint the FK-less
+   * remote doesn't enforce (mostly orphan refs/aliases whose entity wasn't synced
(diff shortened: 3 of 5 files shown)


BYK:
Adversarial review: **SHIP**, no BLOCKERs (classifier empirically FK-exclusive across node:sqlite/bun; level routing + `skipped` accounting mutation-verified). Closed the one coverage NIT:

- Added a positive test: a **non-FK** constraint skip (`entity_aliases` `UNIQUE(alias_type, alias_value)` — FK passes, UNIQUE fails) **stays visible** via `log.notice` and still counts in `res.skipped`. Locks the "rarer skip stays loud" contract (the FK path emits no per-row notice; this one must). Full sync suites (64) green.

BYK:
Fixed the Sentry LOW (summary wording over-asserted an FK cause for a mixed FK+non-FK count): the debug per-table summary now reads "skipped N row(s) failing a local constraint (mostly orphans whose parent wasn't synced under the plan cap)" and the CLI summary "…couldn't apply locally — mostly records referencing data beyond your plan's sync cap". Non-FK skips still also emit their own accurate per-row notice, so the specific cause is never hidden.

PR #1225, opened 2026-07-08
fix(core): checkpoint the SQLite WAL so it can't grow unbounded (#1221)

### LINKED ISSUE #1221: SQLite WAL not checkpointing: 10.8 GB db with a 5.4 GB lore.db-wal
## Summary

On a heavily-used self-hosted install, the SQLite WAL is not checkpointing: the main DB is **10.8 GB** and `lore.db-wal` is **5.4 GB** (plus `-shm`). A WAL that large means passive/auto-checkpoints are not completing, so the WAL grows unbounded.

```
-rw------- 10885758976  lore.db        (10.8 GB)
-rw-------  5479789552  lore.db-wal    (5.4 GB)
-rw-r--r--       32768  lore.db-shm
```

## Why this matters

- Unbounded disk growth (WAL never truncates).
- Read/query slowdowns (readers must scan a huge WAL).
- Long crash-recovery / checkpoint-on-open times.
- A stuck checkpoint usually means a **long-lived read transaction** is pinning the WAL (e.g. the read-worker / vector pool, or the vec0 re-chunk backfill holding a snapshot), or effective `wal_autocheckpoint` is too high / disabled.

## Ask (not yet root-caused — filing per request)

- Confirm whether a long-lived reader (read-worker pool, vector pool, backfill cursor) holds a snapshot that blocks `PRAGMA wal_checkpoint(PASSIVE)`.
- Consider a periodic `wal_checkpoint(TRUNCATE)` during idle, and/or revisit `wal_autocheckpoint` sizing.
- Add a startup/idle log if WAL exceeds a threshold (e.g. > N MB) so this is visible rather than silent.

## Environment

- Self-hosted, opencode plugin (in-process gateway, Bun), `storage_mode=vec0 vec=on`.
- ~6k distillations, large temporal corpus.

BYK:
Closes #1221. Reported from a heavily-used self-hosted install: **10.8 GB `lore.db` with a 5.4 GB `lore.db-wal`** — the WAL was never truncating.

## Root cause: WAL checkpoint starvation
There was **no `wal_autocheckpoint` tuning, no `journal_size_limit`, no explicit `wal_checkpoint`, and no checkpoint on close** — the WAL relied entirely on SQLite's default **PASSIVE** auto-checkpoint. Lore's topology defeats it:
- **one writer** (the main connection) emits a continuous stream of small commits — live traffic **plus** the ~hour-long temporal-embedding backfill;
- **persistent reader-pool connections** (default 2, `db/reader.ts`) serve near-continuous recall, and each query holds a **WAL read-mark**.

A PASSIVE checkpoint can back-fill frames but can **never** get the "no reader below the mark" instant required to *reset* the WAL to frame 0 — so every commit appends past the high-water mark and the WAL grows without bound. Textbook checkpoint starvation.

## Fix
- **`db.ts`**:
  - `PRAGMA journal_size_limit = 64 MiB` at open — truncates the WAL back to the cap after any resetting checkpoint (backstop; default is `-1` = never).
  - `checkpointWal()` — a **TRUNCATE** checkpoint on the writer connection that resets **and** shrinks the file when no reader pins an older snapshot; when one does, it checkpoints what it can and returns `busy` (a harmless no-op retried next tick). Never blocks indefinitely.
  - `walSizeBytes()` — cheap `-wal` stat for telemetry.
  - a TRUNCATE checkpoint on **`close()`** so a graceful shutdown doesn't leave a huge WAL to recover.
- **`idle.ts`**: run `checkpointWal()` on the existing 30s idle tick once the WAL passes an 8 MiB floor — reclaims during the quiet windows between agent turns (when the reader pool is idle). Emits a **notice** if a >512 MiB WAL still won't checkpoint (a reader is continuously pinning), else a debug summary.

Subtlety handled: for `TRUNCATE` mode the `log`/`checkpointed` result columns report the *post-truncation* state (`0/0`) even on success, so the reclaim is measured by the `-wal` **file-size delta**, not those counters (verified empirically on node:sqlite).

## Scope note
The **10.8 GB main file** is a *separate* symptom: `PRAGMA auto_vacuum = INCREMENTAL` is a no-op on a pre-existing DB (SQLite bakes `auto_vacuum` at creation; changing it needs a one-off `VACUUM`), and `incremental_vacuum` only runs at the vec0 cutover. That's a heavier, separate fix — not in this PR.

## Tests
- `checkpointWal(TRUNCATE)` reclaims a deliberately-grown WAL to zero on a single connection (auto-checkpoint disabled for a deterministic pre-size); `reclaimedBytes` measured by file delta.
- `journal_size_limit` is set to 64 MiB at open.
- Full core suite (128 files, 2714) green; typecheck + lint clean.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -7,7 +7,7 @@ import {
   repartitionVec0Project,
 } from "./db/vec-store";
 import { join, dirname } from "node:path";
-import { chmodSync, mkdirSync } from "node:fs";
+import { chmodSync, mkdirSync, statSync } from "node:fs";
 import { getGitRemote } from "./git";
 import { isHostedMode } from "./hosted";
 import { dataDir } from "./data-dir";
@@ -2208,10 +2208,17 @@ export function db(): Database {
   // Retry for up to 5s when another connection holds the write lock (e.g.
   // backgroundDistill's BEGIN IMMEDIATE overlapping with a recall query).
   // Default is 0ms which throws SQLITE_BUSY immediately.
-  database.exec("PRAGMA busy_timeout = 5000");
+  database.exec(`PRAGMA busy_timeout = ${BUSY_TIMEOUT_MS}`);
   // Return freed pages to the OS incrementally on each transaction commit
   // instead of accumulating a free-page list that bloats the file.
   database.exec("PRAGMA auto_vacuum = INCREMENTAL");
+  // Bound the on-disk WAL (#1221). SQLite's default PASSIVE auto-checkpoint is
+  // starved by the persistent reader-pool connections (each recall query holds a
+  // WAL read-mark), so the WAL can grow unbounded (a 5.4 GB -wal was observed).
+  // journal_size_limit truncates the WAL back to this cap after any checkpoint
+  // that resets it — the backstop for the active TRUNCATE checkpoint the idle
+  // scheduler runs via checkpointWal(). Default is -1 (never truncate).
+  database.exec(`PRAGMA journal_size_limit = ${WAL_JOURNAL_SIZE_LIMIT_BYTES}`);
   migrate(database);
   installSyncCapture(database);
   // Load the sqlite-vec native extension (if available) on the RAW connection,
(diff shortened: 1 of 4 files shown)


BYK:
Adversarial review caught a real **BLOCKER** (thanks): running `wal_checkpoint(TRUNCATE)` under the connection's `busy_timeout=5000` synchronously busy-waits up to ~5s when a reader pins the WAL — and that's *exactly* when the feature engages (readers active → WAL grows past the floor → checkpoint attempt). Empirically 5016ms of event-loop stall.

**Fixed** by dropping the write-lock retry to 0 around both checkpoints:
- `checkpointWal()` sets `PRAGMA busy_timeout = 0` before the TRUNCATE and restores `BUSY_TIMEOUT_MS` (now a shared const) in a `finally`. Under a pinning reader it returns `busy` in ~1ms; in a quiet window it still fully truncates.
- `close()` sets `busy_timeout = 0` before its TRUNCATE so a reader mid-query can't hang shutdown (an un-truncated WAL is just recovered on next open).

**Regression test added** (closes the coverage gap the reviewer flagged): a second `openReaderConnection` holds an open read transaction (pins a WAL snapshot); `checkpointWal()` must return `busy=true` in **< 2000ms** (would be ~5000ms without the guard). Passes 3/3; full core suite (128 files, 2715) green, typecheck + lint clean.

Also confirmed from the review: reclaim-by-file-delta is sound (single writer = the sync caller, so nothing grows the WAL between the two stats); `log/checkpointed` reporting 0/0 for TRUNCATE is empirically verified; the 10.8 GB main-file (auto_vacuum) is correctly out of scope.

PR #1227, opened 2026-07-08
feat(core): automatic free-page reclaim on the idle tick (#1221)

BYK:
Companion to #1226 (`lore data vacuum`). This is **(a)** of the #1221 follow-up — it keeps the main DB file bounded **without** a manual command, so space freed by prunes / knowledge compaction / eviction returns to the OS continuously instead of re-accumulating.

## What runs on the idle tick
On the existing 30s idle tick, **before** the WAL checkpoint (so the checkpoint flushes any truncation the same tick):

- **`auto_vacuum=INCREMENTAL` DBs** (the normal case): `incrementalVacuum()` in **bounded batches** (~8 MiB/tick) once the freelist passes a **16 MiB floor**. Cheap; `busy_timeout=0` so it can't stall the loop (same guard as the WAL checkpoint).
- **legacy `auto_vacuum=NONE` DBs** (created before auto_vacuum was set — their freed pages are *never* returned): **once per process**, if the DB is small (**≤128 MiB**) **and** no session is active → run a full `vacuum()` inline to reclaim **and** convert to INCREMENTAL (so the branch above takes over). If it's too large to compact inline → a one-time **notice** pointing at `lore data vacuum` (never block the event loop for minutes).

## How
- **`db.ts`**: `autoVacuumMode()`; `incrementalVacuum(pages)` — runs `PRAGMA incremental_vacuum(N)` under a `busy_timeout=0` guard (like `checkpointWal`), reclaim measured by the **freelist delta** (the file physically shrinks after the next checkpoint, since the truncation lands in the WAL).
- **`idle.ts`**: the reclaim block + constants (`FREELIST_RECLAIM_FLOOR_BYTES` 16 MiB, `INCREMENTAL_VACUUM_BATCH_PAGES` 2048, `SAFE_AUTO_VACUUM_BYTES` 128 MiB) + a one-shot per-process legacy flag.

## Safety
- `incrementalVacuum` is light (freelist + WAL append, not an exclusive rewrite) and `busy_timeout=0` — no event-loop stall (the class of bug caught in #1225).
- The only heavy op (full `vacuum()`) is gated to **small legacy DBs + no active session + once per process**; large legacy DBs (like the reported 10.8 GB one) get the notice, not a multi-minute inline VACUUM.

## Tests
- a fresh DB reports `auto_vacuum=INCREMENTAL`;
- `incrementalVacuum()` clears the freelist (`reclaimedBytes>0`, `freelistBytes()===0`) after a delete, without a full VACUUM.
- Full core suite (130 files, 2720) + gateway idle-work (43) green; typecheck + lint clean.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -3166,11 +3166,60 @@ export function freelistBytes(): number {
  * file size before/after. Followed by a resetting WAL checkpoint so the VACUUM's
  * churn doesn't linger in the -wal.
  */
-export function vacuum(): { beforeBytes: number; afterBytes: number } {
+export function vacuum(opts?: { noWait?: boolean }): {
+  beforeBytes: number;
+  afterBytes: number;
+} {
+  const conn = db();
   const beforeBytes = dbFileSizeBytes();
-  db().exec("VACUUM");
+  // The mode-converting VACUUM needs a near-exclusive moment; a reader pinning a WAL
+  // snapshot would otherwise make it busy-wait the whole BUSY_TIMEOUT_MS (~5s stall).
+  // `noWait` (used by the idle auto-reclaim, which runs on the event loop) drops the
+  // retry to 0 so it fails fast instead of stalling — the caller retries later. The
+  // explicit `lore data vacuum` command omits it so it waits for the lock as expected.
+  // Register the desired auto_vacuum mode so the following VACUUM bakes it into the
+  // header: this converts a legacy auto_vacuum=NONE DB to INCREMENTAL (the whole point
+  // for #1221). Explicit here — not relying on the open-time pragma — so the idle
+  // small-DB convert reliably flips the mode and never re-VACUUMs on the next tick.
+  // Mirrors the migrate() conversion path. No-op when already INCREMENTAL.
+  conn.exec("PRAGMA auto_vacuum = INCREMENTAL");
+  if (opts?.noWait) conn.exec("PRAGMA busy_timeout = 0");
+  try {
+    conn.exec("VACUUM");
+  } finally {
+    if (opts?.noWait) conn.exec(`PRAGMA busy_timeout = ${BUSY_TIMEOUT_MS}`);
+  }
   checkpointWal();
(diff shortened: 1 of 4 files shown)


BYK:
Review caught the BLOCKER (#3) + a SHOULD-FIX (#6) — both fixed:

**#3 (BLOCKER — legacy VACUUM ~5s stall, same class as #1225):** `vacuum()` now takes `{ noWait }`. The idle auto-convert calls `vacuum({ noWait: true })` → `busy_timeout=0` around the VACUUM (restored in `finally`), so a stray reader can't make it busy-wait; it fails fast (SQLITE_BUSY, caught → retried a later tick) instead of stalling the loop. The explicit `lore data vacuum` command still omits `noWait` so it waits for the lock as the user expects. Regression test added: with a reader pinning the DB, `vacuum({ noWait: true })` returns in **< 2000ms** (would be ~5000ms without the guard).

**#6 (SHOULD-FIX — one-shot flag burned unconditionally):** restructured the legacy branch. The flag (now `legacyVacuumNoticeShown`) is only set when the **too-large-DB notice** is actually shown. The small-DB auto-convert path has **no flag** — it self-gates via the `auto_vacuum` mode flip on success, and on a busy gateway (`sessions.size>0`) or SQLITE_BUSY it simply retries a later idle tick instead of being silently skipped for the whole process.

Full core suite (130 files, 2721) + idle-work (43) green; typecheck + lint clean.

BYK:
Addressed the second Sentry comment (db.ts `vacuum()`): it was technically a false positive — the open-time `PRAGMA auto_vacuum = INCREMENTAL` already makes the bare VACUUM convert a legacy NONE DB (empirically verified: mode flips to 2, so the idle branch self-gates and doesn't loop). But I made it **explicit** anyway — `vacuum()` now runs `PRAGMA auto_vacuum = INCREMENTAL` before the VACUUM (no-op when already INCREMENTAL), mirroring the `migrate()` conversion path. This removes the fragile reliance on the open pragma and definitively rules out any re-VACUUM loop. Tests + typecheck + lint green.

(The first Sentry comment on idle.ts:541 was already resolved by the #6 restructure.)

PR #1230, opened 2026-07-08
feat(supabase): raise free-tier entity-graph caps (#1221 follow-up)

BYK:
Follow-up to the #1221 work / the entity-cap concern surfaced during live sync testing (a user with **115 entities synced only 30**, and the refs/aliases to the other ~85 became orphans skipped on pull — #1215/#1222).

## Why
The original free caps (`0003_sync_limits.sql`) sized the entity graph at **entities=30** against **knowledge=500** — a 16× tighter cap on the *cheap, high-value* graph than on the content it indexes. A real 500-entry knowledge base extracts ~100–300 entities, so at 30 the graph that makes knowledge navigable is mostly evicted. Entities are ~200 B/row, so the content cap (knowledge=500) should be the real free-tier limit and the graph should be proportionate.

## Change (migration 0017, row caps only)
| table | before | after |
|---|---|---|
| entities | 30 | **300** |
| entity_aliases | 300 | **3000** |
| entity_relations | 300 | **1500** |
| knowledge_entity_refs | 2000 | **5000** |

knowledge (500) and **pro tier unchanged**. ~1 MB more per free user.

## Byte caps unchanged — on purpose
The free-tier `max_bytes` (0007: entities 1 MB, aliases/relations 2 MB, refs 1 MB) already hold the new row counts with headroom, so the **row cap binds first** and the byte caps stay as the anti-abuse backstop:
| table | new rows | ~bytes/row (user cols) | total | byte cap | per-row budget |
|---|---|---|---|---|---|
| entities | 300 | ~300 | ~90 KB | 1 MB | ~3.5 KB |
| entity_aliases | 3000 | ~122 | ~366 KB | 2 MB | ~699 B |
| entity_relations | 1500 | ~183 | ~275 KB | 2 MB | ~1.4 KB |
| knowledge_entity_refs | 5000 | ~72 | ~360 KB | 1 MB | ~210 B |

## Rollout
Server-side only — takes effect when the migration auto-deploys; **no client release needed** (the 0015 server-side eviction just raises its threshold). Existing users' next sync admits more of their graph.

## Validation
- Migration is a plain idempotent `UPDATE` of existing `plan_limits` rows.
- Updated the integration-test `afterEach` restore map (`FREE_LIMITS`) to match.
- **Full Tier-1 Postgres integration suite (60 tests) green** with 0017 applied (real Docker Postgres). Typecheck + lint clean.

### DIFF
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -621,7 +621,7 @@ function rowIdExpr(m: SyncTableMeta): string {
  * the MOST USEFUL rows (mirrors the knowledge seed; the push drains the outbox in seq
  * order and the server rejects once the cap is hit, so seed order = what survives):
  *  - entities: most-referenced first (by knowledge_entity_refs count), then recency —
- *    a heavily-referenced entity is more useful than a one-off, and the cap is small (30).
+ *    a heavily-referenced entity is more useful than a one-off when the free cap bites.
  *  - entity_relations: recency (updated_at, then created_at).
  *  - entity_aliases: recency (created_at — the table has no updated_at).
  *  - knowledge_meta: LIVE entries only (JOIN knowledge_current), same value order as
--- a/packages/gateway/test/sync-security.integration.test.ts
+++ b/packages/gateway/test/sync-security.integration.test.ts
@@ -15,10 +15,10 @@ import { type PgHarness, startPgHarness } from "./helpers/pg-harness";
 /** Free-tier defaults (mirror supabase/migrations/0003 + 0009) — restored after quota tests. */
 const FREE_LIMITS: Record<string, number> = {
   knowledge: 500,
-  entities: 30,
-  entity_aliases: 300,
-  entity_relations: 300,
-  knowledge_entity_refs: 2000,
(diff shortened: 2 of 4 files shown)


BYK:
Addressed the review:

**Must-fix (#5) — stale second `afterEach`:** the eviction block's `afterEach` (line 872-875) still restored the OLD caps (entities 30, aliases 300, relations 300). Fixed — it now restores from `FREE_LIMITS` (entities 300, aliases 3000, relations 1500) so it can never drift from the migration again. Was inert today (later blocks use fresh users / their own setCaps) but a real landmine; good catch.

**Doc nits:** updated `supabase/README.md` cap list (entities 300 / aliases 3000 / relations 1500 / refs 5000, + note re: 0017) and genericized the stale `30-row` illustration; genericized the `sync-data.ts` seed comment ('when the free cap bites' instead of '(30)').

Re-ran the full Tier-1 Postgres integration suite (60 tests) — green. Typecheck + lint clean.

Confirmed from the review: migration targets exactly the 4 free rows, type-safe/idempotent, caps read dynamically from plan_limits (eviction/quota pick up 0017 automatically), byte caps hold for realistic rows (row cap binds first on all four; refs tightest at ~72 B vs 210 B budget), pro tier + knowledge untouched.

Issue #961, opened 2026-06-24
Publish a coding-agent memory benchmark (no competitor measures this)


BYK:
## Summary

Design and publish a **coding-agent memory benchmark** — a credible eval that measures the thing Lore actually does, which no competitor currently measures.

## Why

Every memory system in this space (ByteRover/Cipher, Mem0, Zep, Hindsight, HonCho, Chronos, etc.) evaluates on **conversational/chat-assistant** benchmarks — primarily **LoCoMo** and **LongMemEval**. These measure "remember personal facts across chat sessions." None of them measure **coding-agent memory**: decision recall, gotcha reuse, cross-session refactor continuity, convention adherence, etc.

This came out of the ByteRover competitive analysis. ByteRover's arXiv paper (arXiv:2604.01599) claims LoCoMo 96.1% / LongMemEval-S 92.8%, but:

- It evaluates conversational recall, **not coding** — despite ByteRover being sold as coding-agent memory.
- The headline numbers are inflated by a separate **"justifier" model** (Gemini 3.1 Pro synthesizes the answer from retrieved context before a Gemini 3 Flash judge scores it). Their own ablation shows removing the justifier path collapses accuracy (92.8 → 63.4).
- Their ablation also shows the marquee "relation graph" / OOD-detection features contribute ~0.4pp (and multi-session *improves* without the relation graph).

**Whoever publishes a credible coding-memory benchmark defines the category.** We already have the rigor culture (adversarial review, mutation testing) to do this honestly, and it doubles as marketing leverage + a regression guard for our own recall/distillation quality.

## What it should measure (axes)

- **Decision recall**: given a multi-session history, does the agent recall a prior architectural/design decision when it becomes relevant again?
- **Gotcha reuse**: does the agent avoid re-introducing a previously-discovered bug class / footgun?
- **Convention adherence**: does it follow project conventions captured earlier (test framework, naming, file layout)?
- **Cross-session refactor continuity**: pick up an in-progress refactor across a session boundary without re-deriving context.
- **Entity/alias resolution**: resolve people/services/tools by alias across sessions.
- **Negative controls**: does it *avoid* surfacing stale/superseded knowledge after it was updated or deleted?

## Methodology guardrails (learn from competitors' mistakes)

- **No "justifier" inflation**: do not let a frontier model synthesize the answer from retrieved context as part of scoring. Score the agent's actual output, or clearly separate retrieval-quality metrics from end-task metrics.
[...]

PR #1231, opened 2026-07-08
feat(distill): retain exact literal/style values (hex, CSS, dimensions, version tags) in observations (#961)

BYK:
## Why

The retrieval diagnostic (#961, harness in #1211) found that the specific facts users ask to recall **verbatim** — hex colors (`#1164a3`), CSS declarations (`border-radius: 10px`, `inset 0 8px 0 0`), pixel dimensions (`18px`), version tags (`v45`), selectors (`[class*="ThreadList"]`) — were dropped from distillations as "styling/config" noise. Distillations held only **~6%** of expected facts vs **~90%** in raw temporal.

`DISTILLATION_SYSTEM` and `RECURSIVE_SYSTEM` already preserve file paths, function names, exact counts, and error codes — but never *named* literal style/config values, so a worker model treats them as skippable prose.

## What

- **`DISTILLATION_SYSTEM`**: new *"Exact literal & style values"* bullet + a worked BAD/GOOD example (chat-panel frame) that explicitly frames styling/config as high-value, not noise.
- **`RECURSIVE_SYSTEM`**: extends *Technical Changes* to carry those literals through meta-consolidation ("never paraphrase a style/config value into prose — 'rounded corners' loses 'border-radius: 10px'").

Complements #1228 (which surfaces these from raw temporal at injection time): richer distillations mean the specifics also survive compaction into the **durable** pool, so they're available even after raw messages age out.

## Tests
- `distillation-prompt-specifics.test.ts` locks in the guidance tokens so the intent can't silently regress.
- `distillation.test.ts` unaffected (121 pass); `tsc` clean.

Follow-up: validate lift via the eval harness (distillation-quality run) — expect the distillation-only passive share to rise from the ~6% floor.

Refs #961

### DIFF
--- a/packages/core/src/prompt.ts
+++ b/packages/core/src/prompt.ts
@@ -63,6 +63,7 @@ In coding sessions, the following must be recorded exactly as they appear:
 - Directory contents when enumerated: "src/auth/ contains: jwt.ts, password.ts, middleware.ts, rate-limiter.ts, routes.ts"
 - Test results: "47 tests passing, 3 failing in auth.test.ts"
 - Environment variables: DATABASE_URL, NEXT_PUBLIC_API_URL, CODECOV_TOKEN
+- Exact literal & style values — record the LITERAL token, never a paraphrase: hex colors (#1164a3, #0d2244), CSS declarations (border-radius: 10px, inset 0 8px 0 0, box-shadow: inset 0 8px 0 0, margin-left: 8px), pixel/size dimensions (18px, 8px), version/tag labels (v45), commit SHAs, and CSS/DOM selectors ([class*="ThreadList"], .virtual-list). "Styling", "visual", or "config" details are NOT low-value — they are exactly what the user asks to recall verbatim later.
 
 BAD: 🟡 Added password hashing to the auth module.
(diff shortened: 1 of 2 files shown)


BYK:
Closing after a design discussion. Storing repo-resident specifics (hex, CSS, selectors, exact values) in distillations is the wrong lever: a stored value is a **cache of the file** that goes stale the moment the code changes, so the distillation/meta would confidently serve a wrong value. For anything that lives in the repo, the correct source of truth is the file itself (and `git blame` for historical values) — and since retrieving a memorized value costs a tool call anyway, a `Read` is the same cost for a *correct* answer.

Lore already has the right mechanism: `extractReferences()` + `knowledge_refs` store `file:line`/symbol pointers per entry, and `referenceValidation` (`knowledge_ref_validity`/`knowledge_symbol_presence`) rots them when they drift. So we keep the **where** in memory and let the model read the **what**.

The genuinely non-file facts (past test/lint outputs, decisions, user-stated numbers) are already served by the passive temporal injection in #1228. No distillation value-bloat needed. Superseded by a pointer/read nudge instead.

PR #1233, opened 2026-07-09
feat(supabase): server-side tombstone reaper (#909)

BYK:
Completes the A2 append-only lifecycle: **create → update → delete → REAP**. Client compaction (#1213) is done; this is the deferred server piece, pulled forward.

## Why
Soft-deleted rows (`is_deleted=true`) stay physically on the remote and **count toward the per-scope row/byte quota** (A1 physical accounting), so a heavy deleter can wedge their own cap with dead tombstones. Content is already scrubbed to `''`/null on delete (#897) — only the row slot remains. This purges tombstones older than a retention window and reclaims the slot.

## Migration 0018
- **`reap_tombstones(retention_days int default 90) returns bigint`** — SECURITY DEFINER, loops all **7 synced tables** (each has `is_deleted` + `updated_at`: knowledge, entities, entity_aliases, entity_relations, knowledge_entity_refs, knowledge_meta, knowledge_meta_crdt), `delete where is_deleted and updated_at < now() - window`, returns the count. The DELETEs fire the existing `maintain_usage` AFTER-DELETE trigger, so **per-scope usage counters auto-decrement** — no accounting here. `revoke all from public`; `grant execute to service_role` for manual admin runs.
- **pg_cron daily schedule**, guarded behind `pg_available_extensions` so it runs on Supabase and is skipped by the stock `postgres:16-alpine` integration harness (the function itself is exercised directly in tests). Idempotent (unschedule-if-exists then reschedule).

## Correctness (resurrection constraint, #909)
A client offline longer than the window could miss a deletion and, if it then modifies+pushes the stale row, resurrect it. Age is measured from `updated_at` (≈ deletion time — client value on INSERT, `sync_set_updated_at` on the UPDATE path), and the 90-day default exceeds the expected offline period. A per-scope **purge-epoch** (force a stale-cursor client to re-pull) is the robust follow-up — deferred, documented in the migration.

**No client change needed:** the client already can't resurrect a synced-unchanged row — `seedOutbox` only enqueues rows whose `content_hash` diverges from `sync_state` (sync-data.ts:609), so a reaped tombstone is never re-pushed; and a vanished remote row simply isn't returned on pull (no crash).

## Tests (Tier-1 Postgres, real Docker)
- reaps old tombstones **across scopes**, keeps live + recent-within-window, **decrements the usage counter** by the reaped count, return value = total;
- `retention_days=0` reaps every tombstone but **never a live row** (even an ancient live row).
- Full suite **62 tests green** (60 existing + 2 new). Typecheck + lint clean.

### DIFF
--- a/packages/gateway/test/sync-security.integration.test.ts
+++ b/packages/gateway/test/sync-security.integration.test.ts
@@ -1424,3 +1424,91 @@ describe.skipIf(gate())(
     });
   },
 );
+
+describe.skipIf(gate())("sync migrations — tombstone reaper (#909)", () => {
+  async function usageRow(scope: string, table: string) {
+    const { rows } = await h.client.query(
+      "select coalesce(row_count,0)::int as rows, coalesce(byte_count,0)::int as bytes from public.user_table_usage where scope_id=$1 and table_name=$2",
+      [scope, table],
+    );
+    return rows[0] ?? { rows: 0, bytes: 0 };
+  }
+
+  // Ancient (well past any window) vs recent (inside the 90-day window).
+  const ANCIENT = "2020-01-01T00:00:00Z";
+  const recentTs = () => new Date(Date.now() - 5 * 86_400_000).toISOString();
+
+  const insK = (scope: string, id: string, deleted: boolean, ts: string) =>
+    h.asUser(scope, (c) =>
+      c.query(
+        "insert into public.knowledge (id, scope_id, category, title, content, is_deleted, updated_at) values ($1,$2,'pattern','t','c',$3,$4)",
+        [id, scope, deleted, ts],
+      ),
+    );
+
+  it("reaps tombstones older than the window, keeps live + recent, decrements usage, across scopes", async () => {
+    const a = await h.createUser();
+    const b = await h.createUser();
+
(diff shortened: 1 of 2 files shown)


BYK:
Addressed the review's follow-ups (verdict was SHIP-WITH-FOLLOWUPS, no BLOCKERs):

1. **account_escrow/scope_keys exclusion documented** — reworded the table-list comment: those keystore tables also carry `is_deleted`/`updated_at` but are intentionally excluded (single-row-per-scope in v1 → delete+recreate is a same-PK upsert, no tombstone accumulation; and we never want a background task deleting encryption-key state). Noted the teams-mode `scope_keys` revisit.
2. **Privilege-boundary regression test added** — asserts a normal `authenticated` user calling `reap_tombstones(90)` is rejected with `permission denied` (guards the `revoke ... from public`).
3. **byte_count decrement asserted** — the main test now checks both `row_count` (3→2) and `byte_count` (strictly lower) after the reap.

Full Tier-1 Postgres integration suite now **63 tests** green. Typecheck + lint clean.

The optional `<`-boundary / per-row O(n) batching notes are noted as future considerations; the pg_cron-not-in-shared_preload_libraries edge only affects a non-Supabase deploy (Supabase preloads it), so it's not relevant for the production target.

PR #1234, opened 2026-07-09
fix(sync): converge pulled entity_alias UNIQUE collisions instead of skipping (#1217)

### LINKED ISSUE #1217: sync: convergent conflict resolution for pulled rows that violate a local-only UNIQUE

BYK:
Closes #1217 (split out from the #1215 orphan-resilience hotfix; Sentry MEDIUM).

## Problem
A pulled `entity_aliases` row that violates the **local** `UNIQUE(alias_type, alias_value)` — a constraint the FK-less remote doesn't enforce — was **skipped** (#1215) to avoid crashing the multi-table pull. But a skip leaves two devices **permanently divergent**: each keeps its own alias for the same (type, value) and neither ever converges. Two devices can independently mint an alias for the same (type, value) with different ids.

## Fix — deterministic, symmetric convergence
The **lower alias id wins on every device**, so the outcome is identical regardless of pull order:
- **remote id < local id** → remote wins: drop the losing **local** alias, apply the remote — *only if* the remote's entity is admitted locally; otherwise it's an FK orphan and the valid local alias stays (falls back to the generic skip).
- **local id < remote id** → local wins: discard the remote loser.

Why min-id and not "remote-wins": "remote always wins" does **not** converge (each device deletes its own and pulls the other's, ping-ponging). A symmetric deterministic tiebreak does.

The losing **local** row is deleted **without** sync-suppression, so its removal propagates via the outbox and the duplicate is cleaned off the remote too (then reaped by #909). The discarded row is recorded to `sync_conflicts` (`resolution=alias_unique_superseded`) so the lost alias is diagnosable, not silently destroyed. **Only** `entity_aliases` UNIQUE errors take this path — FK-orphan / CHECK / NOT NULL skips are unchanged.

## Changes
- `packages/core/src/sync-data.ts` — `resolveAliasUniqueConflict(remote, reapply)`: the tiebreak + FK-orphan guard + `sync_conflicts` recording + loser delete.
- `packages/gateway/src/sync.ts` — `isUniqueConstraintError` + route `entity_aliases` UNIQUE collisions through the resolver in the pull catch (falls back to the generic skip otherwise).

## Tests
- **core unit** (`sync-alias-converge.test.ts`): remote-wins, local-wins, **both pull orders converge to the same winner**, FK-orphan fall-through (valid local kept), no-collision fall-through.
- **gateway wiring** (through the real `pullOnce`): local-wins and remote-wins branches, asserting the winner present / loser gone / `sync_conflicts` recorded / no "skipped a row" notice. The obsolete "UNIQUE stays a visible skip" test is replaced by these.
- Full core+gateway suites green (one unrelated pre-existing flaky `session-rotation-merge` test passes in isolation). Typecheck + lint clean.

### DIFF
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -1340,6 +1340,75 @@ export function recordConflict(
     );
 }
 
+/**
+ * Deterministic convergence for a pulled `entity_aliases` row that violates the local
+ * `UNIQUE(alias_type, alias_value)` the FK-less remote doesn't enforce (#1217). Two
+ * devices can independently mint an alias for the same (type, value) with DIFFERENT ids;
+ * without a rule each device keeps the OTHER's copy and they never converge. The tiebreak
+ * is symmetric — the LOWER alias id wins on EVERY device — so the outcome is identical
+ * regardless of pull order.
+ *
+ * `reapply` re-applies the winning remote row; it's invoked ONLY when the remote wins,
+ * after the losing local row is dropped so the collision is cleared. Returns true when
+ * the conflict is resolved (the caller advances the cursor + counts it), or false to fall
+ * back to the generic constraint skip — either because there is no local (type, value)
+ * collision, or because the remote alias is an FK orphan (its entity isn't admitted under
+ * the cap), in which case the valid local alias must stay.
+ *
+ * The losing local row is deleted WITHOUT sync-suppression, so its removal propagates via
+ * the outbox and the duplicate is cleaned off the remote too (then reaped, #909). The
+ * discarded row is recorded to `sync_conflicts`.
+ */
+export function resolveAliasUniqueConflict(
+  remote: Record<string, unknown>,
+  reapply: () => void,
+): boolean {
+  const at = remote.alias_type;
+  const av = remote.alias_value;
+  const remoteId = remote.id;
+  if (
+    typeof at !== "string" ||
+    typeof av !== "string" ||
+    typeof remoteId !== "string"
+  )
+    return false;
+  const local = db()
+    .query(
+      "SELECT id, entity_id, alias_type, alias_value FROM entity_aliases WHERE alias_type = ? AND alias_value = ?",
+    )
+    .get(at, av) as
+    | { id: string; entity_id: string; alias_type: string; alias_value: string }
+    | undefined;
+  // No local collision, or the very same row (an ordinary update, not a conflict) →
+  // let the caller apply/skip it normally.
+  if (!local || local.id === remoteId) return false;
+
+  if (remoteId < local.id) {
+    // Remote alias has the lower id → it wins. But only proceed if it can actually
+    // apply: its entity must be admitted locally, else it's an FK orphan and the valid
+    // local alias must stay (the caller skips the orphan remote row).
+    const entityId = remote.entity_id;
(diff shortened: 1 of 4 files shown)


BYK:
Addressed the review (verdict was SHIP-WITH-FOLLOWUPS, no BLOCKERs):

**C1 (SHOULD-FIX) — fixed in-PR, not deferred.** The `drainTimestamp` path (>PAGE rows sharing one exact ms) previously routed alias UNIQUE collisions to the generic skip, so the convergence guarantee had a gap there. Factored the whole resolve-or-skip catch into a shared `handlePulledRowConstraint(meta, remote, rid, e, res, touchedFts, enc)` now called by BOTH the primary pull loop and `drainTimestamp` — so they converge identically. Added a regression test (`drainTimestamp also converges an alias UNIQUE collision`) that forces the drain path (205 rows at one ms + a lower-id collider landing in the drain batch) and asserts the collision resolves (remote wins, local loser dropped + recorded) rather than skipping. Would fail if the drain path were ever un-wired.

C2 (clearSyncState on the loser) and C3 (integration coverage of the FK-orphan-collision case) are benign per the review — the outbox delete + push already reconciles the loser's sync_state, and the FK-orphan guard is unit- + mutation-tested. Leaving both as-is.

Confirmed from the review: min-alias-id tiebreak is symmetric/deterministic (byte-order string compare, converges for all pull orders — mutation-proven, reversing the tiebreak fails 6 tests); FK-orphan guard prevents dropping a valid local alias regardless of constraint ordering; the reapply can't wedge and self-heals; only `entity_aliases` UNIQUE takes this path (generic FK/CHECK/NOT NULL skips unchanged).

Full gateway sync suite 61 tests + core resolver suite 5 tests green. Typecheck + lint clean.

PR #1235, opened 2026-07-09
feat(sync): safe-reap watermark so no active device misses a deletion (#909)

BYK:
Follow-up to the #909 tombstone reaper (#1233): the per-scope **purge epoch**, in a form that's actually correct for the eviction model.

## Why not the literal purge-epoch
#909 proposed a stale-cursor client "re-pull from scratch instead of trusting its cursor." That only helps if paired with **reconcile-by-absence** (delete local rows absent from the remote), which conflicts with an established invariant: the remote is a **bounded cache** (eviction #1203/#1206/#1207), so a row absent from it is ambiguous — *evicted-but-valid* vs *deleted-and-reaped*. Reconcile-by-absence would wrongly drop valid local data (`reconcile()` deliberately never treats remote-absence as a delete). The reaped tombstone is physically gone, so no pull can recover it.

## The correct form — server-side safe-reap watermark
Instead of forcing the client to reconcile, the **server holds off reaping** until **every active device in the scope has pulled past a tombstone**. Then no active device can miss a deletion, and no client reconcile is ever needed.

- `updated_at < min(pulled_through)` across the scope's active devices ⇒ every active device's keyset cursor already passed the tombstone ⇒ each applied the deletion ⇒ safe to reap.
- Strictly better than the fixed 90-day window: **precise** protection within the active window (a slow device is never outrun) + **faster quota reclaim** (reap as soon as all devices catch up, not after a blanket 90 days). Same residual bound only for a device abandoned past the active-TTL.

## Server — migration `0019_reaper_watermark.sql`
- `sync_device_progress(scope_id, device_id, table_name, pulled_through, last_seen)` — a **control table** (not synced; the client writes it directly out-of-band). `last_seen` is **server-stamped** by a trigger so a client can't fake activity to stall reaping; `pulled_through` is client-set (over/under-claiming only self-harms). RLS by scope, a rows-per-scope cap, and an index for the reaper's correlated `min()`.
- `reap_tombstones(retention_days=90, active_ttl_days=90)` redefined: per-scope correlated `delete … where updated_at < coalesce(min(active pulled_through), now()-retention)`. Falls back to the retention window for a scope with no active device (dormant / pre-feature). Also cleans up progress rows for devices abandoned beyond 2× the window. Drop-first avoids overload ambiguity — the pg_cron `reap_tombstones(90)` call re-binds (2nd arg defaulted).

## Client — `packages/gateway/src/sync.ts`
- `reportDeviceProgress()`: after each `syncOnce` pull, best-effort upsert of this device's per-table `pulled_through` (device id = the shared `replicaId()`, reused from the confidence CRDT). Never throws — a reporting failure never breaks sync.

## Tests
- **Tier-1 Postgres integration** (real Docker Postgres + migration 0019, 70 tests total): watermark holds a tombstone until the **slowest active device** passes it; an **abandoned** device is excluded (can't hold back reaping); **no-active-device** falls back to the retention window; an active device sitting **behind** a tombstone blocks its reaping regardless of age; abandoned progress-row **cleanup**; `last_seen` **server-stamping**; **RLS** isolation between scopes.
- **Gateway unit**: `syncOnce` reports per-table device progress; the mock now accepts array upserts (faithful to supabase-js).
- Full core+gateway suites green (5484 pass), typecheck + lint clean.

Residual (documented): a device abandoned longer than the active-TTL and later revived may still miss deletions reaped in its absence — the same bound as the prior fixed window, now the *only* gap rather than a blanket one.

### DIFF
--- a/packages/core/src/ltm.ts
+++ b/packages/core/src/ltm.ts
@@ -194,8 +194,9 @@ function insertMeta(
 // resists a single penalty) — the accepted cost of a faithful additive CRDT.
 
 /** Stable per-device replica id (UUIDv7), minted once into KV. Not cached: the
- *  test suite swaps the DB between cases, so always read it from the live DB. */
-function replicaId(): string {
+ *  test suite swaps the DB between cases, so always read it from the live DB. Also
+ *  reused as the device id for sync's server-side reaper watermark (#909). */
+export function replicaId(): string {
   let id = getKV("sync.replica_id");
   if (!id) {
     id = uuidv7();
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -30,6 +30,11 @@ import {
 import { putWrappedScopeKey } from "./crypto/keystore";
 import { rematerializeConfidence } from "./ltm";
 
+// The stable per-device id (also used by the confidence CRDT) doubles as the device id
+// for sync's server-side reaper watermark (#909). Re-exported so the gateway can report
+// this device's pull progress under `syncData.replicaId()`.
+export { replicaId } from "./ltm";
+
 /**
  * Run `fn` with this connection's sync capture suppressed (re-entrant,
  * connection-scoped). Re-exported from `db` so the apply helpers and the engine
--- a/packages/gateway/src/sync.ts
+++ b/packages/gateway/src/sync.ts
@@ -973,6 +973,40 @@ function applyRemote(
 // syncOnce — push then pull
 // ---------------------------------------------------------------------------
 
+/**
+ * Report this device's per-table pull progress so the server-side reaper only reaps a
+ * tombstone once EVERY active device in the scope has pulled past it (#909 watermark) —
+ * no client reconcile-by-absence needed, which the eviction model forbids. Best-effort:
+ * a failure never breaks sync. `last_seen` is server-stamped (a client can't fake
+ * activity); `pulled_through` is this device's keyset cursor timestamp per table.
+ */
+async function reportDeviceProgress(client: SupabaseClient): Promise<void> {
+  try {
+    const user = await getCurrentUser();
+    if (!user?.user_id) return;
+    const deviceId = syncData.replicaId();
+    const rows: Array<Record<string, unknown>> = [];
+    for (const meta of syncData.syncedTables("basic")) {
+      if (meta.pullOnly) continue; // pull-only tables (e.g. profiles) are never reaped
+      const ms = parseCursor(getKV(pullKey(meta.table))).ms;
+      rows.push({
+        scope_id: user.user_id,
+        device_id: deviceId,
+        table_name: meta.table,
+        pulled_through: new Date(ms).toISOString(),
+      });
+    }
+    if (rows.length === 0) return;
+    const { error } = await client
+      .from("sync_device_progress")
+      .upsert(rows, { onConflict: "scope_id,device_id,table_name" });
+    if (error)
+      log.info(`sync: device-progress report skipped: ${error.message}`);
+  } catch (e) {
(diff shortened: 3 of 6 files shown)


BYK:
Addressed the review (verdict was SHIP-WITH-FOLLOWUPS, no BLOCKERs) — all three in-PR:

**SHOULD-FIX #1 — cap trigger blocked updates at the boundary.** The BEFORE INSERT cap fired for an `insert … on conflict do update` even when it resolved to an UPDATE, so a re-report at ≥700 rows would wrongly raise. Fixed: the cap now only counts a genuinely NEW `(scope, device, table)` row (`NOT EXISTS(pk) AND count >= 700`). Added an integration test proving a re-report (update path) never trips it and upserts in place.

**CONCERN #2 — rollout window.** Switched the cutoff from `coalesce(watermark, now()-retention)` to **`least(watermark, now()-retention)`**. The retention floor is now always preserved: a device that isn't reporting yet (mid-rollout / old gateway) keeps the same protection as the old fixed window, while an active-but-slow device is still protected *beyond* the window (its cursor holds the watermark). So it's now at-least-as-safe as the fixed window in every case rather than strictly-better-but-with-a-rollout-gap. `least` also ignores a NULL watermark, so the no-active-device fallback still resolves to the floor. Updated the header/inline comments to this accurate framing.

**CONCERN #3 — strict `<` boundary test.** Added an integration test: two active devices, a tombstone at *exactly* the slower device's cursor ms → asserts it is **not** reaped (a `<`→`<=` mutation would now be caught).

NITs 4/5: fixed the cap comment (no longer claims "7×100") and renamed the test's `reapable` → `reported` (it's the non-pull-only superset; the 2 extra keystore rows are inert — the reaper only queries its own table_names).

Rewrote the watermark integration tests to relative dates (`daysAgo(n)`) so they exercise the watermark *below* the retention floor and stay robust to wall-clock. Verify: **72** integration tests (real Postgres + migration 0019) + **62** gateway unit, typecheck + lint clean.

BYK:
Re: the Sentry `BUG_PREDICTION` on sync.ts:991-996 (empty-table cursor → 1970 `pulled_through`) — evaluated; it's a **false positive**, and the suggested fix would introduce a real pull bug. No change.

**No bloat for the cited case (empty table).** A table that's empty for the scope has **no tombstones**, so a 1970 watermark blocks reaping of… nothing. The bot's example ("a new user's knowledge table") reaps zero rows with or without a cursor — there is nothing to bloat.

**A 1970 report for a NON-empty table means the device hasn't pulled it yet — and blocking is correct.** `reportDeviceProgress` runs *after* `pullOnce`, so a device that actually has rows reports its advanced cursor. A device reports 1970 for a non-empty table only when it genuinely hasn't pulled that table (first sync mid-flight, or encryption `locked`). In that state the device hasn't seen those tombstones, so holding off reaping is exactly the safety guarantee this PR exists to provide — reaping them would let that device miss the deletions. The only lingering-tombstone case is a device that is active yet perpetually locked/never-pulling a table; that's the correct conservative cost, and it self-heals the moment the device pulls (its cursor advances past 1970).

**The suggested fix is unsafe.** "Advance the cursor to the query time even when no rows are returned" would move the keyset cursor **past rows the device hasn't pulled** (anything written at/after the query instant, or an encrypted table skipped without a pull), silently dropping them — a pull-correctness regression. `pullOnce` deliberately advances the cursor **only past rows it actually applied**; that invariant is load-bearing for both pull completeness and this watermark.

This is the same case the pre-merge adversarial review examined (scrutiny point 6, PASS): per-table (only the empty table's own watermark is affected) and harmless. Erring toward *not* reaping is the intended direction.

PR #1236, opened 2026-07-09
fix(sync): converge pulled entity_relations UNIQUE collisions instead of skipping (#1217)

BYK:
Follow-up to #1234 (alias UNIQUE convergence). `entity_relations` has the **same** local-only secondary UNIQUE the FK-less remote doesn't enforce:

- **Local:** `UNIQUE(entity_a, entity_b, relation)` (db.ts:695)
- **Remote:** PK `(owner_user_id, id)`, no triple UNIQUE (0002); `idColumns: ["id"]` → `applyRemoteUpsert` keys `ON CONFLICT(id)`

So two devices can each mint the same relation triple under different `id`s → both live on the remote → a pulling device's second one violates its local triple UNIQUE → it was **skipped** (#1215), leaving the devices permanently divergent (each keeps its own id for the same edge).

## Fix
Deterministic convergence, mirroring the alias resolver: the **lower `id` wins on every device**, applied symmetrically, so both converge regardless of pull order. The loser is recorded to `sync_conflicts`; when the local row loses, its **unsuppressed** delete propagates so the remote duplicate is cleaned up (and reaped, #909).

Unlike aliases, **no FK-orphan guard is needed**: the collision matches on `entity_a` AND `entity_b`, so the remote's endpoints are identical to the colliding local row's — which exist locally (that row's own FK) — so the winner can always apply. The delete-then-reapply also self-heals on the next pull if the reapply ever threw (same property as aliases).

## Scope
- `packages/core/src/sync-data.ts`: `resolveRelationUniqueConflict`.
- `packages/gateway/src/sync.ts`: `handlePulledRowConstraint` routes `entity_relations` UNIQUE errors through it (both tables share one `isUniqueConstraintError` branch; the generic FK/CHECK/NOT-NULL skip is unchanged).
- **Tests:** core unit (`sync-relation-converge.test.ts`, 5: remote-wins, local-wins, both-orders-converge, no-collision, same-id) + gateway wiring through `pullOnce` (local-wins asserts no "skipped a row" notice; remote-wins asserts the local loser is dropped + the remote applied).

This closes the last table with a divergent-on-skip local-only UNIQUE. `knowledge` is version-aware (keyed by logical_id, no cross-device same-triple collision) and `knowledge_entity_refs` uses a composite PK resolved via `ON CONFLICT` — neither is subject to this pattern.

Full core+gateway suites green (the one `cache-stability.e2e` blip is pre-existing cross-test flakiness — passes 16/16 in isolation; this PR touches only sync). Typecheck + lint clean.

### DIFF
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -1414,6 +1414,67 @@ export function resolveAliasUniqueConflict(
   return true;
 }
 
+/**
+ * #1217, mirror of resolveAliasUniqueConflict for entity_relations' local-only
+ * UNIQUE(entity_a, entity_b, relation): the FK-less remote keys only on `id`, so two
+ * devices can push the SAME relation triple under different ids. On pull, the second one
+ * violates the local triple UNIQUE and was skipped (→ divergence). Resolve it the same
+ * way: the lower `id` wins on EVERY device, so both converge regardless of pull order.
+ *
+ * Unlike aliases, NO FK-orphan guard is needed: the collision matches on entity_a AND
+ * entity_b, so the remote's endpoints are identical to the colliding local row's — which
+ * exist locally (that row's own FK) — so the winner can always apply. (And even if the
+ * reapply threw, the local loser is already gone, so the next pull re-applies cleanly.)
+ */
+export function resolveRelationUniqueConflict(
+  remote: Record<string, unknown>,
+  reapply: () => void,
+): boolean {
+  const ea = remote.entity_a;
+  const eb = remote.entity_b;
+  const rel = remote.relation;
+  const remoteId = remote.id;
+  if (
+    typeof ea !== "string" ||
+    typeof eb !== "string" ||
+    typeof rel !== "string" ||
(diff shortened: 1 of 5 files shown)


BYK:
Added the reviewer's NIT #1 (fan-out guard) to this PR: `sync-registry-contract.test.ts` now scans **every** synced table's live schema (`PRAGMA index_list`/`index_info`) for a local-only secondary UNIQUE — a UNIQUE whose columns differ from the remote's upsert key (`origin != 'pk'` and `!= idColumns`) — and asserts the table has registered convergence handling (`entity_aliases`→resolver, `entity_relations`→resolver, `knowledge`→version-aware apply). A future synced table that adds such a UNIQUE now fails this contract until it gets a resolver + routing, rather than silently regressing to divergent-on-skip.

NIT #2 (entity-merge triple mutation) is inherited from the alias resolver and self-heals on fuzzy graph edges — left as-is, shared behavior.

107 tests green across the 4 sync convergence/contract/wiring files; typecheck + lint clean.

Issue #826, opened 2026-06-19
D — Pro tables: distillations + referenced temporal (encrypted)


BYK:
> **Part of epic #821.** Depends on: #822, #823, #824, #825
> _This issue body mirrors the design doc `.opencode/plans/sync-D-pro-tables.md`._

---

# D — Pro tables: distillations + referenced temporal (encrypted)

Parent: `sync-sharing-epic.md`. Depends on: A1 + A2 (uniform append model) + C
(encryption) + B (tier). Encrypted from day one — never ships plaintext.

## Goal
Back up + cross-device sync the `distillations` table and the
distillation-referenced subset of `temporal_messages`, encrypted, bounded, and
without the firehose / self-destructing-backup traps found in review round 1.

## What syncs
- All `distillations` rows (every generation, archived or not).
- `temporal_messages` where `id ∈ ⋃ distillations.source_ids` (uniform across
  generations — `metaDistill` carries forward gen-0 source ids). Undistilled raw
  messages NEVER leave the device.

## Capture (distillation-driven; epic + B5)
- `captureStrategy` on `SyncTableMeta`: distillations = `distillation-driven`,
  temporal_messages = captured via distillations only (no per-row trigger).
- TEMP trigger on `distillations` INSERT enqueues the distillation AND
  `INSERT INTO sync_outbox … SELECT 'temporal_messages', value … FROM
  json_each(new.source_ids)`. UPDATE (archived flip) re-enqueues the distillation.
- 🔴 `installSyncCapture` must (re)install pro triggers when the tier mirror (B)
  first resolves `tier='pro'` — capture install runs before tier is known.

## Lifecycle (review round 1 fixes — epic decisions #1/#2/#3)
- 🔴 Prune is sync-invisible: `temporal.prune()` (deletes distilled temporal +
  archived distillations, `temporal.ts:690/730/742`) must clear `sync_state`
  WITHOUT a tombstone. Run prune under a capture-suppressed context.
- 🔴 `seedOutbox`/`reconcile` MUST be subset-aware for temporal (only the
  json_each set) — never enqueue an undistilled message (B2).
- 🔴 Restore exemption: restored rows keyed off local residency, not original
  `created_at`, so the next idle prune doesn't delete them (B3). Set
  `distilled=1` on restore.
- Remote bounded by per-scope quota (A1 counter), not a TTL reaper.
- `distillations` meta: `versioned: true` (archived flips), `ftsTables:
[...]

PR #1240, opened 2026-07-09
feat(supabase): pro-tier remote mirror for distillations + temporal (D-1, #826)

BYK:
First PR of the **D** stack (epic #821, pro-tier encrypted backup). **Server-only and inert** — adds the remote Postgres mirror + tier gate for `distillations` and the distillation-referenced subset of `temporal_messages`. No client change yet, so the tables sit empty until the client half (D-2/D-3) lands. Safe to deploy.

## `0020_pro_sync.sql` (copies the 0009 add-synced-table structure)
- **`current_tier()`** — SECURITY DEFINER, reads the caller's own `profiles.tier` (bypasses profiles RLS for its single scoped row), coalesces to `'free'`.
- **`distillations`** — VERSIONED (`content_hash`/`revision`); the `archived` flip is a real UPDATE that re-pushes. Encrypted-on-wire cols: `narrative, facts, observations`. `source_ids` stays cleartext (id references only).
- **`temporal_messages`** — APPEND-ONLY (`versioned:false`, no `content_hash`/`revision`, no tombstones). Encrypted-on-wire cols: `content, metadata`. The local `distilled` residency flag is **not** mirrored.
- **Scope seam** (0007): `scope_id`/`author_id` default `auth.uid()`. RLS **USING open** (a downgraded ex-pro can still PULL their backup); **WITH CHECK** pins `scope_id`+`author_id` to the writer AND requires `current_tier()='pro'` → free write is `42501`.
- **Size CHECKs** with base64/envelope headroom on the encrypted cols; id/hash caps.
- **`plan_limits` PRO rows only** (no free row — free is RLS-denied); generous caps (distillations 200k/256MB, temporal 1M/512MB — tunable, flagged for product).
- Both tables are `(scope_id, id)`-keyed → the shared `enforce_row_quota` **generic branch** already fits, so **no function redefinition** — just attach the quota + `maintain_usage` triggers. Idempotent usage backfill.

## Tier-1 integration tests (real Postgres, +10)
`current_tier()` mirror; pro `plan_limits` seeded + **no** free row; free write → `42501`; pro write ok + downgraded read-works/write-blocked; forged `author_id` → `42501`; `scope_id` immutable → `23514`; cross-scope isolation; pro row cap → `23514`; pro byte cap → `23514`; encrypted-col size CHECK → `23514`. **82/82 pass** against real Postgres locally.

## Stack / lifecycle
D-1 (this) → D-2 (client tier-aware plumbing) → D-3 (pro metas + distillation-driven capture + subset seed + prune sync-suppression — arms it) → D-4 (Tier-2 encrypted round-trip). The remote is bounded by the per-scope **quota counter, not a TTL reaper** (epic decision); the client's local prune stays sync-invisible (D-3) so a pruned local row never deletes the remote backup.

Open (flagged, non-blocking): pro quota sizing (placeholders); quota-pause vs server-side oldest-eviction on fullness (D uses pause per epic); encrypting `temporal.metadata` (defaulted yes).

### DIFF
--- a/packages/gateway/test/sync-security.integration.test.ts
+++ b/packages/gateway/test/sync-security.integration.test.ts
@@ -1702,3 +1702,247 @@ describe.skipIf(gate())("sync migrations — reaper watermark (#909)", () => {
     expect(rows[0].n).toBe(1); // upsert updated in place, did not duplicate
   });
 });
+
+describe.skipIf(gate())(
+  "sync migrations — pro tier gate + quota (D, #826)",
+  () => {
+    // Pro-tier plan_limits defaults (0020) — restored after each test so cap
+    // mutations never leak. These tables have NO 'free' row (free is RLS-denied),
+    // so the shared quota suite's FREE_LIMITS afterEach never touches them.
+    const PRO_ROW_CAP: Record<string, number> = {
+      distillations: 200000,
+      temporal_messages: 1000000,
+    };
+    const PRO_BYTE_CAP: Record<string, number> = {
+      distillations: 268435456,
+      temporal_messages: 536870912,
+    };
+
+    const makePro = (uid: string) =>
+      h.asService((c) =>
+        c.query("update public.profiles set tier='pro' where id=$1", [uid]),
+      );
+    const setFree = (uid: string) =>
+      h.asService((c) =>
+        c.query("update public.profiles set tier='free' where id=$1", [uid]),
+      );
+    const setProRowCap = (table: string, n: number) =>
+      h.client.query(
+        "update public.plan_limits set max_rows=$1 where tier='pro' and table_name=$2",
+        [n, table],
+      );
+    const setProByteCap = (table: string, n: number) =>
+      h.client.query(
+        "update public.plan_limits set max_bytes=$1 where tier='pro' and table_name=$2",
+        [n, table],
+      );
+
+    afterEach(async () => {
+      if (gate()) return;
+      for (const [t, n] of Object.entries(PRO_ROW_CAP))
+        await setProRowCap(t, n);
+      for (const [t, n] of Object.entries(PRO_BYTE_CAP))
+        await setProByteCap(t, n);
+    });
+
+    it("current_tier() reflects the profiles mirror", async () => {
+      const a = await h.createUser();
+      const free = await h.asUser(a, (c) =>
+        c.query("select public.current_tier() as t"),
+      );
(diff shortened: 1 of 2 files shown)


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS, no BLOCKERs** (tier gate, RLS scope isolation, author/scope forgery rejection, at-cap oracle-close, scope immutability all verified + mutation-tested; quota reuse claim empirically proven — generic id-probe → 23514 not 42703; migration idempotent; usage backfill non-doubling).

Addressed the one real defect + a coverage NIT in this PR:
- **#5 (archived schema/comment mismatch) — FIXED.** Added `archived boolean not null default false` to the remote `distillations` table so the documented archived-flip re-push contract is actually satisfiable: D-3 can include `archived` in `syncColumns` → the flip changes `content_hash` → re-push → `sync_set_updated_at` bumps the cursor → peers re-pull. (Was: no remote column → would PGRST204-stall or silently drop archive state.)
- **N2 — added** a forged-`scope_id` INSERT test (mirrors the forged-author_id one; → 42501 via RLS WITH CHECK after the trigger's oracle-close early-return).

Deferred (as noted in the PR): DELETE self-harm on own backup (N1, product call), pro quota sizing (N3), quota-pause-vs-eviction (N4). **83/83** integration tests pass against real Postgres.

BYK:
Re: Sentry MEDIUM on `temporal_messages` RLS allowing UPDATE (append-only) — acknowledged, **accepted for D-1, tracked for D-3**.

- **Not a security/correctness issue:** the `for all` policy still scopes every op to `scope_id = auth.uid()`, so a direct-PostgREST UPDATE could only rewrite the user's OWN backup rows (self-harm), never cross-scope; `enforce_row_quota` still gates byte growth on UPDATE; and an errant update just bumps `updated_at` so peers re-pull it (the versioned:false + `(scope_id, updated_at)` cursor handles it without a crash or divergence).
- **Consistent** with the existing append-only-ish `knowledge_meta_crdt` (also `versioned:false`, `for all`).
- **Defense-in-depth follow-up:** when D-3 arms these tables client-side I'll tighten `temporal_messages` to deny UPDATE at the RLS layer (split `for all` into SELECT/INSERT/DELETE policies — `distillations` legitimately keeps UPDATE for the archived flip). Tracked in `.opencode/plans/sync-D-pro-tables-exec.md`.

The tables are inert until D-3, so there is no exposure in the meantime.

PR #1241, opened 2026-07-09
feat(sync): tier-aware synced-table selection (D-2, #826)

BYK:
Second PR of the **D** stack (#826). **Pure plumbing refactor** so the engine syncs the table set for the user's PLAN tier instead of a hardcoded `"basic"`. `SYNCED_TABLES.pro`/`max` are still empty → **behavior-identical today** (everyone gets basic); it just removes the hardcoded `"basic"` so D-3 can populate the pro set and have it sync automatically.

## core (`sync-data.ts`)
- **`META_BY_TABLE`** now indexes ALL tiers (basic ∪ pro ∪ max) → `meta()`/`metaFor()` resolve a table regardless of tier (a table's shape is tier-independent; tier only gates whether it's *synced*).
- **`syncedTablesFor(tier)`** — the CUMULATIVE set: basic ⊆ pro ⊆ max. This is what the engine actually syncs (a Pro user gets basic **plus** pro). `syncedTables(tier)` stays as the non-cumulative single-tier accessor.
- **`currentSyncTier()`** — maps the plan tier (`currentTier()`: free|pro|max) → SyncTier (free→basic, pro→pro, max→max); unknown plan → basic (safe default).
- **`metaFor(table)`** — tier-independent meta resolver.
- `seedOutbox`/`reconcile`/`enableSync` default to `currentSyncTier()` and iterate `syncedTablesFor(tier)`; `assertSyncInvariants` covers the cumulative set.

## gateway (`sync.ts`, `cli/sync-cmd.ts`)
- The 5 `sync.ts` + 2 CLI hardcoded `syncedTables("basic")` sites → `syncedTablesFor(currentSyncTier())`; `tableMeta()` resolves via tier-independent `metaFor()` (avoids a tier-timing edge for pro tables); `cmdEnable` calls `enableSync()` (defaults to the current plan's SyncTier).

## Tests
+3 unit tests: `syncedTablesFor` cumulative invariant (basic ⊆ pro ⊆ max), `currentSyncTier` plan mapping incl. unknown→basic, `metaFor` resolve/throw. **Full core+gateway suite green (5501 passed / 101 skipped, 0 fail)** — zero behavior change with pro/max empty.

Stack: D-1 (#1240, merged) → **D-2 (this)** → D-3 (pro metas + distillation-driven capture + subset seed + prune sync-suppression — arms it) → D-4 (Tier-2 encrypted round-trip).

### DIFF
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -318,12 +318,48 @@ export const SYNCED_TABLES: Record<SyncTier, SyncTableMeta[]> = {
   max: [],
 };
 
+// Every registered table across ALL tiers — so meta()/metaFor() resolve a table
+// regardless of the caller's current tier (a Pro table's shape is tier-independent;
+// tier only gates whether it is SYNCED, not whether its meta exists).
 const META_BY_TABLE = new Map<string, SyncTableMeta>();
-for (const t of SYNCED_TABLES.basic) META_BY_TABLE.set(t.table, t);
+for (const t of [
+  ...SYNCED_TABLES.basic,
+  ...SYNCED_TABLES.pro,
+  ...SYNCED_TABLES.max,
+])
+  META_BY_TABLE.set(t.table, t);
 
-/** The synced table set for a tier (only `basic` is populated today). */
+/** The tables registered for exactly one SyncTier (non-cumulative). */
 export function syncedTables(tier: SyncTier = "basic"): SyncTableMeta[] {
   return SYNCED_TABLES[tier];
+}
+
+/**
+ * The CUMULATIVE synced table set at a SyncTier: basic ⊆ pro ⊆ max. This is what
+ * the engine actually syncs — a Pro user gets the basic tables PLUS the pro ones.
+ * (pro/max are empty until #826/D populates them, so this equals `basic` today.)
+ */
+export function syncedTablesFor(tier: SyncTier): SyncTableMeta[] {
+  const out = [...SYNCED_TABLES.basic];
+  if (tier === "pro" || tier === "max") out.push(...SYNCED_TABLES.pro);
+  if (tier === "max") out.push(...SYNCED_TABLES.max);
+  return out;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


BYK:
Adversarial review: **SHIP-WITH-FOLLOWUPS** — core contract (zero behavior change while pro/max empty) proven via mutation testing + full green core suite; typecheck clean; the 3 new tests are discriminating (both mutations killed).

**Fixed the one real finding inline (#2):** `sync.ts:296` — the outbox **prune-floor** computation still used `syncedTables("basic")` (my `replaceAll` missed it because there `syncData` and `.syncedTables(...)` are split across two lines in the chained call). No-op today (basic==cumulative), but post-D-3 the prune floor would ignore pro tables → `pruneOutbox(minCursor)` could delete unpushed pro outbox rows → silent pro data loss. Now uses `syncedTablesFor(currentSyncTier())`, matching the push loop. 85/85 gateway sync tests pass, typecheck clean.

A behavioral regression test for this is **impossible in D-2** (basic==cumulative today, so no test can distinguish the two sets — the same property that makes D-2 'zero behavior change'). Recorded as a D-3 follow-up (prune-floor-includes-pro test once `pro` is non-empty), along with the reviewer's other D-3 items (registry-contract + `assertSyncInvariants` cumulative; `syncedTablesFor` dedup guard) in `.opencode/plans/sync-D-pro-tables-exec.md`.

BYK:
Re: Sentry on `assertSyncInvariants` (:1557) — **latent, cannot manifest today, tracked for D-3.** Invariant #3 builds its `known` set from the tier-scoped `syncedTablesFor(currentSyncTier())`, so post-D-3 a pro→free downgrade that leaves pro-table `sync_state` rows would false-flag 'references unregistered table.' With `pro`/`max` empty today there are no pro tables or pro `sync_state` rows, so it can't fire. D-3 fix (recorded in the plan): base invariant #3's `known` set on `META_BY_TABLE` keys (all registered tables, tier-independent) since `sync_state` may legitimately hold rows for any registered table regardless of current tier.

PR #1243, opened 2026-07-10
feat(supabase): enforce append-only on temporal_messages (D-3a, #826)

BYK:
Closes the D-1 Sentry MEDIUM: 0020 gave both pro tables a single `for all` RLS policy, but `temporal_messages` is **append-only** (the client only ever inserts; a message's content is never rewritten in place). `for all` still permitted UPDATE, letting a direct-PostgREST client rewrite its own stored conversation content (self-harm only, but contradicts the model + the `versioned:false` / `(scope_id, updated_at)` pull-cursor assumptions).

## `0021_temporal_append_only.sql` (belt-and-suspenders)
- **REVOKE UPDATE** on `temporal_messages` from `authenticated` → hard `42501`, not a silent 0-row no-op.
- Replace the `for all` policy with explicit **SELECT / INSERT(pro) / DELETE** policies (no UPDATE) so the policy set itself reflects append-only. Reads stay open (downgraded pull); DELETE stays allowed (self-purge, no tier gate — the local prune is sync-invisible, the server reaper handles remote lifecycle).
- **`distillations` untouched:** it's versioned and legitimately UPDATEs (the archived flip re-pushes), so it keeps its `for all` policy.

## Tier-1 tests (+2)
temporal UPDATE → `42501` with content unchanged + INSERT/SELECT/DELETE still work; distillations UPDATE (archived flip) still works. **85/85** pass against real Postgres.

Stack: D-1 (#1240) + D-2 (#1241) merged → **D-3a (this, small server follow-up)** → D-3b (client arming) → D-4 (Tier-2 round-trip).

### DIFF
--- a/packages/gateway/test/sync-security.integration.test.ts
+++ b/packages/gateway/test/sync-security.integration.test.ts
@@ -1944,5 +1944,81 @@ describe.skipIf(gate())(
       );
       expect(err.code).toBe("23514");
     });
+
+    it("temporal_messages is append-only: UPDATE is denied (42501); INSERT/SELECT/DELETE allowed (0021)", async () => {
+      const a = await h.createUser();
+      await makePro(a);
+      await h.asUser(a, (c) =>
+        c.query(
+          "insert into public.temporal_messages (id, scope_id, content) values ('t1',$1,'orig')",
+          [a],
+        ),
+      );
+      // UPDATE is revoked at the privilege level → hard 42501, not a silent no-op.
+      const err = await expectError(() =>
+        h.asUser(a, (c) =>
+          c.query(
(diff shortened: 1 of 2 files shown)


BYK:
Review: **SHIP, no blockers** — REVOKE mutation-proven load-bearing (removing it fails exactly the append-only test); INSERT/SELECT/DELETE + all quota/trigger paths intact; distillations untouched; idempotent + correctly ordered. Added the C1 hardening nit: a test asserting `authenticated` holds exactly {DELETE, INSERT, SELECT} (no UPDATE) on `temporal_messages` via `information_schema.role_table_grants` — pins the invariant against a future accidental re-grant. C2 (comment) was already covered. 86/86 pass.

PR #1244, opened 2026-07-10
feat(sync): arm Pro-tier distillation + temporal backup, client side (D-3b, #826)

BYK:
Wires the **client half** of pro-tier sync: the compressed conversation memory (distillations + the distillation-referenced subset of `temporal_messages`) now syncs for pro users. **Free users are entirely unaffected** (tier stays basic, pro tables untouched). Pro sync is pre-GA. The pull/restore side (residency exemption, encrypted round-trip) lands in D-4.

## Registry (`sync-data.ts`)
- `SyncTableMeta` gains **`captureStrategy`** (`"row"` | `"distillation-fanout"` | `"none"`).
- `SYNCED_TABLES.pro` registers **distillations** (fanout capture, encrypted narrative/facts/observations) and **temporal_messages** (append-only `versioned:false`, `captureStrategy "none"`, encrypted content/metadata).

## Capture (`db.ts`)
- `installSyncCapture` installs a **tier-gated distillation-fanout trigger**: on INSERT enqueue the distillation AND fan out one `temporal_messages` outbox row per id in `source_ids` (`json_each`) — 🔴 this referenced subset is the ONLY temporal that ever syncs; an **undistilled message is never enqueued**. On UPDATE re-enqueue the distillation (archived flip). Pro-only; the else-branch DROPs it on downgrade.
- **`reinstallSyncCapture()`** exported to reconcile the trigger set on a tier flip.
- (`json_each`-in-trigger validated against node:sqlite before building on it.)

## Seed / reconcile (`sync-data.ts`)
- `seedSelect`: `temporal_messages` restricted to the `json_each`-referenced subset; distillations recency-ordered.
- `reconcile` **SKIPS the delete-tombstone pass** for the append-only Pro tables — a local prune / project cleanup / eviction must never wipe the remote backup (remote lifecycle is the server reaper's job, epic #821).

## Prune (`temporal.ts`)
- the 3 delete passes run under `withSyncApplying` + clear the pruned rows' `sync_state` (sync-invisible eviction; keeps `sync_state` bounded; no tombstone).

## Gateway (`sync.ts`)
- `syncOnce` detects a plan-tier flip across the pull → reinstalls capture + reconciles, so a free→pro upgrade (or a fresh Pro device) arms capture and seeds pre-existing distillations.

## D-2 follow-ups folded in
- registry-contract battery + `assertSyncInvariants` now cover the **cumulative all-tier set** (+ a cross-tier dedup guard); invariant #3's known-set spans all tiers (a downgraded ex-pro's Pro `sync_state` is not "drift").

## Tests
+6 unit tests (fanout capture / tier-gating / archived flip / subset seed / no-tombstone reconcile / sync-invisible prune) + registry-contract now exercises the pro tables. **Full core+gateway suite green (5517 passed / 104 skipped, 0 fail)**; typecheck + lint clean.

Stack: D-1 (#1240) + D-2 (#1241) + D-3a (#1243) merged → **D-3b (this)** → D-4 (pull/restore residency + Tier-2 encrypted round-trip).

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -2390,7 +2390,58 @@ function installSyncCapture(database: Database) {
       INSERT INTO sync_outbox (table_name, row_id, op, changed_at)
       VALUES ('scope_keys', new.member_user_id, 'upsert', ${ts});
     END;`;
+  // D (#826): Pro-tier distillation-fanout capture. Installed ONLY when the plan
+  // tier (from the pulled profiles mirror) is pro/max — a free user creating
+  // distillations must not accrue un-pushable pro outbox entries that no push cursor
+  // ever drains. On a tier flip the gateway calls reinstallSyncCapture() to add/drop
+  // these (TEMP + IF NOT EXISTS is idempotent; the else-branch DROP removes them on
+  // downgrade).
+  const tierRow = database
+    .query("SELECT tier FROM profiles LIMIT 1")
+    .all()[0] as { tier?: string } | undefined;
+  const isProTier = tierRow?.tier === "pro" || tierRow?.tier === "max";
+  if (isProTier) {
+    // On INSERT: enqueue the distillation AND fan out one temporal_messages outbox
+    // row per referenced source id (json_each over source_ids). This REFERENCED
+    // SUBSET is the ONLY temporal that ever syncs — temporal_messages has NO capture
+    // trigger of its own (captureStrategy "none"), so an undistilled message is never
+    // enqueued. On UPDATE (the archived flip): re-enqueue the distillation only (its
+    // source set is immutable; a no-op non-archived UPDATE such as embedding backfill
+    // is deduped away by the push-side content_hash check). No DELETE: the local prune
+    // is sync-invisible (temporal.prune runs under capture-suppression + clears
+    // sync_state), so a pruned local row must NOT tombstone the remote backup.
+    sql += `
+      CREATE TEMP TRIGGER IF NOT EXISTS distillations_outbox_ins
+      AFTER INSERT ON distillations WHEN (${gate})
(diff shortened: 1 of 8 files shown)


BYK:
## Adversarial review → **BLOCKER found + fixed**

Review verdict was BLOCK on one issue (findings 1–6, 8–10 all PASS, free-user contract airtight, all 6 mutations killed their target tests):

**🔴 BLOCKER (finding 7):** `pushEntry` sent `is_deleted: false` on every upsert, but remote `temporal_messages` (0020) has **no `is_deleted` column** (deliberately — append-only) → PGRST204 → every temporal row silently poisoned → the temporal backup was 100% non-functional. The unit mocks didn't catch it because `REMOTE_COLUMNS` had no entry for the pro tables (no column validation).

**Fix:**
- Added `appendOnly` meta flag (temporal_messages); `pushEntry` omits `is_deleted` for append-only tables (both upsert + delete branches). `versioned` is the wrong gate — the join table is `versioned:false` yet *has* `is_deleted`.
- Made the test mock **faithful**: registered distillations (with `is_deleted`) + temporal_messages (**without**) in `REMOTE_COLUMNS`, and added a Pro-push test asserting the distillation carries `is_deleted` while temporal does not, and the undistilled message never syncs. **Verified discriminating** — removing the `appendOnly` flag flips it to PGRST204 → test fails.

**Non-blocking CONCERNs also addressed:**
- `json_valid` guard on the temporal seed select (a corrupt `source_ids` can't abort the whole seed).
- Added a discriminating test for the invariant-#3 all-tiers `known` set (reviewer's mutation F had survived).

Full core+gateway suite green (5519 passed / 104 skipped, 0 fail); typecheck + lint clean.

PR #1245, opened 2026-07-10
feat(sync): restore-residency for pulled temporal messages (D-4a, #826)

BYK:
The **pull/restore half** of Pro-tier temporal backup. A pulled/restored message lands on a device with the SOURCE device's original `created_at`, but must survive locally for a full retention window rather than being evicted on the very next idle tick (B3). Completes the functional Pro-sync feature (D-3b armed the push side).

## `migration v67`
`temporal_messages.restored_at` — a **LOCAL-ONLY** residency clock (never synced; not in `syncColumns`). NULL for native rows.

## `applyRemoteTemporal` (sync-data.ts)
A pulled temporal row is marked **`distilled=1`** (it was ALREADY distilled on the source device — never re-distill it; it belongs in the archival tier, not the active context window) and stamped **`restored_at=now`**. 🔴 BOTH set only on INSERT and PRESERVED on conflict-update, so a device pulling its OWN pushed rows back keeps its native flags (no prune-clock reset). Routed in `sync.ts`'s `applyRemote` dispatch; content/metadata arrive decrypted (C-4).

## `prune` (temporal.ts)
The TTL pass + size-cap eviction key off **`COALESCE(restored_at, created_at)`** so a restored message ages out on its LOCAL residency clock, not its old origin `created_at`. Native rows unchanged (`restored_at` NULL → `created_at` governs).

## Tests
+3 unit tests: restored survives TTL prune while a native old row is pruned; `applyRemoteTemporal` INSERT sets `distilled=1`+`restored_at`; self-pull-back preserves native flags. Schema-version assertions bumped 66→67. **Full core+gateway suite green (5527 passed / 104 skipped, 0 fail)**; typecheck + lint clean.

Stack: D-1 (#1240) + D-2 (#1241) + D-3a (#1243) + D-3b (#1244) merged → **D-4a (this)** → D-4b (Tier-2 real-engine encrypted round-trip, end-to-end validation).

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1674,6 +1674,17 @@ const MIGRATIONS: string[] = [
   `
   -- Version 66 (#909): see applyRefFkDrop — no-op SQL marker.
   `,
+  // Version 67 (#826/D): temporal_messages.restored_at — a LOCAL-ONLY residency clock.
+  // A pulled/restored message keeps its ORIGIN created_at (for ordering + recall) but is
+  // stamped restored_at=now on apply (applyRemoteTemporal); the TTL/size prune keys off
+  // COALESCE(restored_at, created_at) so a restored message gets a full LOCAL retention
+  // window instead of being evicted on the very next idle tick despite its old origin
+  // created_at (B3). NULL for native rows (created_at governs). NEVER synced (not in
+  // temporal_messages' syncColumns). Re-runs harmlessly on recovery (stripAppliedAlters
+  // drops the duplicate ALTER).
+  `
+  ALTER TABLE temporal_messages ADD COLUMN restored_at INTEGER;
+  `,
 ];
 
 // Index of the migration whose work is performed by a column-presence-aware JS
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -1079,6 +1079,67 @@ export function applyRemoteUpsert(
   });
 }
 
+/**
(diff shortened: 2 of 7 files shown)


BYK:
## Adversarial review → **SHIP-WITH-FOLLOWUPS** (all 🔴 invariants mutation-verified)

Review confirmed via mutation testing: no wire leak of restored_at/distilled (structurally impossible — not in syncColumns); self-pull-back does NOT reset the prune clock (mutation B killed the test); no native-row prune regression; migration idempotent + version normalized to 67; restored rows are recall-searchable (FTS trigger fires under withApplying); typecheck clean.

**Folded in the SHOULD-FIX + a NIT:**
- The reviewer found a surviving mutant: removing *only* the `distilled` half of the conflict-exclusion passed all 130 tests (the self-pull-back test seeded `distilled=1`, so the flip was a no-op). **Added** a test seeding a local `distilled=0` (active) row + pulling the same id back, asserting distilled stays 0 and restored_at stays NULL. **Verified discriminating** — removing `c !== "distilled"` now fails it.
- Memoized the `PRAGMA table_info` in `applyRemoteTemporal` (was per-pulled-row → N+1 at the Pro cap).

**Deferred (noted for the epic, pre-existing / not this PR):** the pull path has no `ensureProject` — a fresh device pulling rows for an unknown project FK-violates (`project_id NOT NULL REFERENCES projects`). This predates D-4a (temporal + distillations already applied via the generic upsert). Will address in D-4b / the epic.

Full core+gateway suite green (5528 passed / 104 skipped, 0 fail).

Issue #1246, opened 2026-07-10
sync: pull path must ensureProject before applying rows (project_id FK) for true cross-device restore


BYK:
## Problem

Every synced content table has `project_id TEXT NOT NULL REFERENCES projects(id)` (db.ts:62/100/280/1809 — knowledge, entities, temporal_messages, distillations, …), but the pull/apply path (`applyRemote` in `packages/gateway/src/sync.ts` → `applyRemoteUpsert`/`applyRemoteKnowledge`/`applyRemoteTemporal`) **never calls `ensureProject`**.

So when a device pulls rows for a `project_id` it doesn't have a local `projects` row for (a genuinely fresh device, or one that hasn't opened that project yet), the INSERT hits a **FK violation** with `foreign_keys = ON`. Post-#1215 orphan-resilience treats a constraint violation as a **poison row → skipped + recorded to `sync_conflicts` + cursor advanced**. The rows are silently dropped and (because their `sync_state` isn't set) won't retry without a cursor reset.

## Impact

- Undermines the core cross-device **restore** story: device-2 pulling another device's knowledge/entities/distillations/temporal for a project it hasn't materialized locally drops those rows.
- **Pre-existing** for basic-tier tables (knowledge/entities) — not introduced by the D epic — but newly relevant now that Pro temporal/distillation backup (#826, D-3b/#1244 + D-4a/#1245) makes multi-device restore a first-class feature.

Surfaced in the D-4a (#1245) adversarial review as an epic-level follow-up.

## Fix direction

On pull, `ensureProject(project_id)` for each distinct incoming `project_id` before applying its rows (under apply-suppression so it doesn't enqueue). Open questions: what project *path* to synthesize for a project known only by id on a fresh device (the `projects` row needs a path); whether to defer materialization until the local git remote matches; interaction with `UNATTRIBUTED_PROJECT_PREFIX`. Likely wants a dedicated design pass.

## Validation

The **D-4b Tier-2 real-engine round-trip** (device-1 push → device-2 restore, encrypted) will exercise exactly this path against real Postgres/PostgREST and should be built to assert the restored rows actually land (which will fail today for an unknown project, proving the gap).

BYK:
## Design locked — converge via git_remote (reuse ensureProject machinery)

[...]

PR #1248, opened 2026-07-10
fix(sync): make projects delete-invisible so reconcile never tombstones the mapping (#1246)

BYK:
Follow-up to #1247, addressing a Sentry `BUG_PREDICTION` (MEDIUM) posted right after that PR auto-merged.

## Bug
`projects` had no `captureStrategy`, so `reconcile()`'s delete-tombstone pass (sync-data.ts:964, which skips only `pullOnly` + `captureStrategy !== "row"` tables) **processed** it — enqueuing a `delete` for any `sync_state` row whose local row is gone. Two triggers:
- `convergeProjectsByRemote` merges a loser and deletes it locally; its `sync_state` lingers → reconcile tombstones it.
- a genuine `deleteProject`.

Either tombstones the **shared** remote id→git_remote mapping, which would **race a merge loser's re-keyed content** AND **nuke another device's still-active project** (id→git_remote is shared identity, not per-device state).

## Fix
A dedicated **`deleteInvisible`** flag on `SyncTableMeta` (projects sets it); `reconcile` skips the delete-tombstone pass for it.

Why not reuse existing flags:
- `captureStrategy: "none"` means *"no own trigger, enqueued via fanout"* — projects HAS its own INSERT/UPDATE trigger, so "none" would break the registry-contract `trigger IFF captureStrategy != none` test.
- `appendOnly` omits `is_deleted` for a remote table lacking the column — projects is versioned + updatable and has `is_deleted`.

`seedOutbox` gates only on `pullOnly`, so projects still seeds.

## Test
`sync-projects.test.ts`: a merged loser with a lingering `sync_state` row is NOT tombstoned by a subsequent `reconcile` — verified it FAILS without the flag (mutation-checked). Full suite **5543 green**; typecheck + lint clean.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -2404,11 +2404,12 @@ function installSyncCapture(database: Database) {
   // #1246: projects identity mapping (id→git_remote). Gated on git_remote IS NOT NULL —
   // only remote-backed projects sync (a remote-less project's random id can't correlate
   // cross-device; its content would FK-poison on a peer). The UPDATE trigger fires the
-  // git_remote backfill (null→remote) into the outbox. NO DELETE: a merge
-  // (convergeProjectsByRemote) deletes the loser locally, but the remote loser row is left
-  // as benign orphaned data (never tombstoned — is_deleted stays false, so the reaper does
-  // NOT collect it; it is one row/merge, within quota). A DELETE tombstone would instead
-  // race the loser's still-orphaned remote content. path is DEVICE-LOCAL, never synced.
+  // git_remote backfill (null→remote) into the outbox. NO DELETE trigger AND projects is
+  // deleteInvisible (reconcile skips its delete-tombstone pass), so a local deletion — a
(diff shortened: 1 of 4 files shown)


BYK:
## Adversarial review → SHIP-WITH-FOLLOWUPS (no BLOCKERs, mutation-verified)

Confirmed: the guard is positioned after seedOutbox + before the delete-tombstone INSERT (suppresses only DELETE-tombstoning; projects still seeds + pushes upserts via its triggers); `deleteInvisible` is genuinely distinct from `appendOnly` (is_deleted payload omission) and `captureStrategy: "none"` (no-own-trigger contract); genuine-delete non-propagation is the intended safety property (a project deleted on device A must not nuke device B's still-active project); regression test mutation-proven discriminating.

**Folded in 2 review nits:** fixed the misattributed test comment ("captureStrategy none" → "deleteInvisible"), and added a registry-contract guard asserting a `deleteInvisible` table has NO DELETE capture trigger (else a trigger would tombstone directly, bypassing the reconcile skip — closes the future-footgun the reviewer flagged). 53/53 in the two suites.

PR #1249, opened 2026-07-10
test(sync): Tier-2 end-to-end project-identity restore round-trip (D-4b, #1246)

BYK:
Proves **P1** (#1247) against the **real Postgres + PostgREST** stack (real migrations incl. `0022`, RLS, encryption, supabase-js) — the layer that catches the `22008`/`PGRST204`/keyset class of bugs the hand-written mock hides.

## Scenario
device-1 (encryption armed) creates a **remote-backed** project (`git_remote` set) + knowledge under it, pushes → device-2 (fresh, same account) wipes all local state and pulls:

- asserts the remote `projects.git_remote` is **CIPHERTEXT** (a repo URL is identifying metadata — never in the clear);
- **pull #1 (locked):** escrow + scope_keys arrive first → projects + knowledge deferred (never stored as ciphertext);
- after unlock, **pull #2:** projects **decrypts + seeds the synthetic-path FK parent** (registered *before* knowledge in the pull order), then the content applies **under it** — without P1 this knowledge would reference a nonexistent local project and poison-skip;
- asserts the seeded row has `path = lore:project/<id>` + the **decrypted** git_remote, the content is restored, `resolveProjectByRemoteOrPath(url) → proj-A` (a later checkout would **adopt** it), and the final push is `0` (**no ping-pong** — pulled rows applied under capture-suppression).

## Notes
Additive only: one `LORE_INTEGRATION`-gated test + the encryption describe's `beforeEach` now also clears local+remote `projects`. **13/13** engine integration tests pass with Docker; typecheck + lint clean.

Completes the P1 stack: P1 #1247 + hardening #1248 + **this validation**. Next: **P2** (gate content capture/seed on `git_remote` across the knowledge/entity families — efficiency/privacy optimization).

### DIFF
--- a/packages/gateway/test/sync-engine.integration.test.ts
+++ b/packages/gateway/test/sync-engine.integration.test.ts
@@ -19,6 +19,7 @@ import {
   ensureProject,
   keystore,
   ltm,
+  resolveProjectByRemoteOrPath,
   setKV,
   syncData,
 } from "@loreai/core";
@@ -445,8 +446,12 @@ describe.skipIf(SKIP)(
       db().exec(
         "DELETE FROM account_identity; DELETE FROM account_escrow; DELETE FROM scope_keys",
       );
+      // #1246: the synced test projects (remote-backed /dev1/* + pulled lore:project/*).
+      db().exec(
+        "DELETE FROM projects WHERE path LIKE '/dev1/%' OR path LIKE 'lore:project/%'",
+      );
       keystore.lock();
-      for (const t of ["account_escrow", "scope_keys"]) {
+      for (const t of ["account_escrow", "scope_keys", "projects"]) {
         await h.client.query(`DELETE FROM public.${t}`);
       }
     });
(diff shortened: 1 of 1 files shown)


BYK:

Verified discriminating via two mutations in a throwaway worktree (each kills exactly this test): (A) disabling the `applyRemoteProject` dispatch → projects poison-skips on NOT NULL path → restore assertions fail; (B) removing `encryptedColumns` → git_remote hits the wire in cleartext → ciphertext assertion fails. Encryption/lock/pull-order claims confirmed against the real engine (projects registered after key tables, before knowledge; deferred while locked, cursor frozen, re-pulled after unlock); no state leakage into sibling tests; final push=0 (created_at round-trips exactly, content_hash over plaintext). 13/13 with Docker, typecheck clean.

Folded in the two test-hygiene nits (aligned with the sibling knowledge test): assert `encryptionState()==='off'` before pull #1, and assert `scope_keys` landed on the remote after the flush push. The convergence path (convergeProjectsByRemote) e2e coverage is noted as a future case (out of D-4b's single-project-restore scope).

PR #1251, opened 2026-07-10
feat(sync): gate knowledge + entities content on git_remote (P2a, #1246)

BYK:
**P2** stops syncing PROJECT-SCOPED content of a REMOTE-LESS project — its random per-device id can't correlate cross-device, so uploading a local/scratch repo's private data is wasteful and a privacy leak (on a peer it just logical-orphans; for Pro tables it would FK-poison). **P2a** covers the direct-`project_id` basic tables (`knowledge`, `entities`); the indirect children (metas/refs/aliases/relations, **P2b**) and the Pro fanout (**P2c**) follow.

## Changes
- **Capture** (`db.ts` `installSyncCapture`): `knowledge` + `entities` triggers gain a git_remote gate — a row syncs iff its project is remote-backed **OR** it is GLOBAL/cross-project (`cross_project=1` or NULL `project_id`, which retain an origin project_id but aren't tied to its remote and must **always** sync). Children (`entity_aliases`/`entity_relations`) stay ungated here (P2b gates them through their parent entity).
- **Seed** (`sync-data.ts`): the `entities` seedSelect and the `knowledge_current` seed path get the same gate via a shared `remoteBackedGate()` fragment (kept in lockstep with the trigger).
- **Re-seed on backfill** (the novel mechanism): `ensureProject`'s git_remote NULL→set backfill fires a new `onProjectRemoteBackfilled` hook → `reseedProjectContent(projectId)`, re-enqueuing the project-scoped content the gate held back while it was remote-less. **Without this, a project that gains a remote later would silently STOP being backed up** until the next `lore sync enable` — a regression vs today. A registered hook avoids the `db.ts`↔`sync-data.ts` circular import; scoped, no transaction wrapper (`ensureProject` may run inside one).

## Tests
New `content git_remote gate — P2a` describe (capture/seed gating, global exemption, re-seed after backfill, no-op when disabled). Existing sync fixtures updated to remote-backed projects (they exercise content that must sync). **Full suite 5549 green**, typecheck + lint clean.

Part of the P2 stack (P2a here → P2b children → P2c Pro fanout). Design in `.opencode/plans/project-identity-sync.md`.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -30,6 +30,26 @@ function fireProjectMutation(): void {
 }
 
 /**
+ * #1246 (P2): fired when a project's git_remote flips NULL→set (ensureProject lazy
+ * backfill). Content created while the project was remote-less was NOT captured (the
+ * P2 git_remote gate), so it must be re-enqueued now that the project can correlate —
+ * otherwise it would silently stop being backed up until the next `lore sync enable`.
+ * The sync layer registers a handler (avoids a db.ts↔sync-data.ts circular import).
+ */
+let onProjectRemoteBackfilledCb: ((projectId: string) => void) | null = null;
+
+/** Register a handler for git_remote NULL→set backfill. Only one is supported. */
+export function onProjectRemoteBackfilled(
+  cb: ((projectId: string) => void) | null,
+): void {
+  onProjectRemoteBackfilledCb = cb;
+}
+
+function fireProjectRemoteBackfilled(projectId: string): void {
+  onProjectRemoteBackfilledCb?.(projectId);
+}
+
+/**
  * Extract the repository name from a normalized git remote URL.
  *
  * Examples:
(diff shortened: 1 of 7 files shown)


BYK:

Mutation-verified in a throwaway worktree — 5 mutations each kill the new P2a tests (non-vacuous):
| Mutation | Result |
|---|---|
| remove capture `projectGate` | capture + reseed tests fail |
| neuter seed `remoteBackedGate` | seed test fails |
| drop global/cross clauses from capture | global-exemption + reseed fail |
| reseed hook no-op | reseed test fails |
| drop `cross_project=0` in reseed | reseed test fails |

Confirmed: (a) capture gate (`%R`→ref) and seed `remoteBackedGate()` are logically identical (EXISTS ≡ IN for non-NULL, NULL short-circuited) — **no drift**; (b) global/NULL-project exemption present + discriminated in **both** paths (global knowledge never stops syncing); (c) `reseedProjectContent` opens **no** transaction → cannot nest-BEGIN inside `ensureProject`; fires only on genuine NULL→set, no loop; (d) reseed keys (knowledge=logical_id, entities=id) match the push plan exactly; (e) fixtures stamp git_remote via a raw suppressed UPDATE (not `ensureProject`) so they don't trigger reseed or mask regressions. 163 sync tests + typecheck clean.

Non-blocking follow-up noted: a rare gateway+CLI concurrent double-fire of `reseedProjectContent` yields benign duplicate upsert outbox entries (idempotent at push, pruned normally).

PR #1252, opened 2026-07-10
feat(sync): gate indirect child content on parent syncability (P2b, #1246)

BYK:
Extends **P2a** (#1251) from the direct-`project_id` tables to the **CHILDREN** that have no `project_id` of their own — they sync iff their **parent** knowledge/entity is itself syncable (remote-backed OR global/cross-project). Closes the remaining privacy/efficiency gap: a remote-less project's confidence register, entity aliases, relations, and refs no longer upload.

## Changes
- **Capture** (`db.ts installSyncCapture`): new `knowledgeParentGate`/`entityParentGate` builders (reuse `directGate`). Gate **INSERT/UPDATE** of `knowledge_meta`, `knowledge_meta_crdt` (parent knowledge via logical_id), `entity_aliases` (parent entity), `entity_relations` (**both** endpoints), `knowledge_entity_refs` INSERT (**both** parents). **DELETE stays UNGATED** — a delete of a never-synced child pushes as an idempotent no-op, and gating it would depend on the parent still existing, but the parent (entity) is cascade-deleted FIRST (`ON DELETE CASCADE`), which would spuriously drop a legitimate delete of a *synced* child.
- **Seed** (`sync-data.ts`): mirror gates on the child seedSelects (`knowledge_meta` JOIN gains the git_remote condition; `entity_relations`/`entity_aliases`/`knowledge_entity_refs`/`knowledge_meta_crdt` get parent-gate WHEREs) via matching `knowledgeParentGate`/`entityParentGate` helpers — kept in lockstep with the triggers.
- **Re-seed on backfill**: `reseedProjectContent` extended to re-enqueue the project's children when it gains a remote — single-parent children keyed off the project's knowledge/entities; the two-parent children (relations, refs) additionally require the OTHER endpoint syncable.

## Tests
New `child content git_remote gate — P2b` describe (per-table capture gating, both-endpoints relation rule, global-parent exemption, seed gating, re-seed of all 5 child tables after backfill). Fixtures injecting metas/refs updated to have remote-backed parents. **Full suite 5554 + integration 106 green**, typecheck + lint clean.

P2 stack: P2a (#1251, merged) → **P2b (here)** → P2c (Pro fanout). Design in `.opencode/plans/project-identity-sync.md`.

### DIFF
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -2308,6 +2308,19 @@ function installSyncCapture(database: Database) {
     "(SELECT value FROM team_config WHERE key='sync.enabled')='1' " +
     "AND NOT EXISTS (SELECT 1 FROM temp._sync_applying)";
   const ts = "CAST(strftime('%s','now') AS INTEGER)*1000";
+  // P2 (#1246) content git_remote gates. A row syncs only if it is REMOTE-BACKED or
+  // GLOBAL — a remote-less project's random id can't correlate cross-device, so its
+  // private data must not upload. `directGate(c)` for a table WITH project_id/cross_project
+  // (knowledge, entities): cross_project=1 (promoted/global) OR NULL project_id OR the
+  // project has a git_remote. `knowledgeParentGate`/`entityParentGate` gate a CHILD (no
+  // project_id of its own) on whether its parent knowledge/entity is itself syncable.
+  const directGate = (c: string) =>
+    `(${c}.cross_project = 1 OR ${c}.project_id IS NULL OR ` +
+    `EXISTS (SELECT 1 FROM projects p WHERE p.id = ${c}.project_id AND p.git_remote IS NOT NULL))`;
+  const knowledgeParentGate = (idExpr: string) =>
+    `EXISTS (SELECT 1 FROM knowledge k WHERE COALESCE(k.logical_id, k.id) = ${idExpr} AND ${directGate("k")})`;
+  const entityParentGate = (idExpr: string) =>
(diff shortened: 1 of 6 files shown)


BYK:

Mutation-verified in a throwaway worktree — 4/4 mutants killed by the exact intended test:
| Mutation | Result |
|---|---|
| remove entity_aliases capture parent gate | capture test fails |
| remove entity_relations 2nd-endpoint gate | both-endpoints test fails |
| remove entity_relations reseed enqueue | reseed test fails |
| remove entity_aliases seed gate | seed test fails |

Confirmed: (1) capture↔seed **lockstep** for all 5 children (`directGate("k")` ≡ `remoteBackedGate("k.")`; EXISTS ≡ IN for non-NULL); (2) **DELETE stays ungated** — verified push of a delete for a non-existent remote row is an idempotent no-op (not a wedge), and the parent entity is cascade-deleted first (ON DELETE CASCADE) so a gated DELETE would spuriously drop a legit tombstone; (3) entity_relations/knowledge_entity_refs require **both** endpoints/parents syncable (a mixed/remote-less endpoint never syncs); (4) reseed param binding correct (relations/refs pass projectId twice), no transaction wrapper, cross_project=0 scoping avoids double-enqueue.

Applied the one substantive follow-up inline: reseed now anchors knowledge_meta/crdt/refs off `knowledge_current` (live-only), matching `seedSelect` rather than the any-version capture gate — so a deleted entry's lingering meta isn't re-enqueued on backfill.

PR #1260, opened 2026-07-11
feat(gateway): publish identity public key to the directory (E-3b, #827)

BYK:
Client half of **E-3** (#827), stacked on E-3a (#1259, merged). Design: `.opencode/plans/sync-E-teams-orgs-impl.md`.

## What
- **`publishIdentityPub(client)`** in `syncOnce` (best-effort, mirroring `reportDeviceProgress`): when encryption is `on`, upsert this device's account identity **public** key into remote `identity_pub` so an admin can later HPKE-wrap the per-scope DEK to this member (E-4). The **private key never leaves the device**.
  - `user_id` fills via the remote `default auth.uid()` (payload carries only the key, like `scope_id` elsewhere).
  - Gated on a **KV hash** of the published key → upserts only when it changes, so it never bumps the remote `updated_at` and churns co-members' pull cursors every cycle.
- **Migration 0026:** recast `identity_pub.public_key` **`bytea` → base64 `text`**. The wire convention for all key material is base64 text (`scope_keys.wrapped_dek`, `account_escrow.wrapped_secret`); publishing a pubkey as `bytea` would be a silent-wrong-bytes footgun for E-4 (which decodes wraps with `Buffer.from(v,'base64')`). The table is empty (E-3b is the first publisher), so the recast is trivial + lossless; size cap adjusted to `octet_length ≤ 512` (base64 of a ≤256-byte key).

## Deferred to E-4
Pulling/consuming **co-members'** pubkeys — E-4 fetches `identity_pub[M]` directly at wrap-time (no local mirror needed).

## Why bytea in 0025 → text now
0025 (E-3a) created the column as `bytea`; on review of the client path I found the whole crypto sync uses base64 text. Rather than let a single column diverge, 0026 aligns it before any client publishes (caught pre-consumption).

## Verification (Tier-1/2, real Postgres)
- New engine test: encryption on → `publishIdentityPub` → remote `identity_pub` holds exactly one row for the user, `public_key` == base64 of the local key; a second publish is a **no-op** (KV-gated, `updated_at` unchanged).
- 0025 identity_pub tests updated to the base64-text contract + a "stored as base64 text, not `\x`-hex" assertion.
- Integration **131/131** (security 93 + engine 14 + orgs 9 + membership 7 + identity-scopekeys 8); unit sync **144/144**; gateway typecheck exit 0; Biome clean.

Adversarial + SECURITY review to follow before merge.

### DIFF
--- a/packages/gateway/src/sync.ts
+++ b/packages/gateway/src/sync.ts
@@ -1038,6 +1038,42 @@ async function reportDeviceProgress(client: SupabaseClient): Promise<void> {
   }
 }
 
+const IDENTITY_PUB_KV = "sync.identityPub"; // base64 of the last-published public key
+
+/**
+ * Publish this device's account identity PUBLIC key to the remote `identity_pub` directory
+ * (E-3, #827) so a scope admin can later HPKE-wrap the per-scope DEK to this member (E-4).
+ * The PRIVATE key never leaves the device. Best-effort: any failure is logged at info and
+ * never breaks sync. Only runs when encryption is "on" (an identity exists + is unlocked).
+ *
+ * `user_id` is filled by the remote column's `default auth.uid()` (like scope_id elsewhere),
+ * so the payload carries only the key. Gated on a KV hash of the published key so we upsert
+ * ONLY when it changes — re-publishing every cycle would bump the remote `updated_at` and
+ * churn co-members' pull cursors for no reason. The key is base64 (the wire convention for
+ * all key material — matches scope_keys.wrapped_dek so E-4 decodes it identically).
+ */
+export async function publishIdentityPub(
+  client: SupabaseClient,
+): Promise<void> {
+  try {
+    if (keystore.encryptionState() !== "on") return; // no unlocked identity to publish
+    const pub = Buffer.from(keystore.getAccountIdentity().publicKey).toString(
+      "base64",
+    );
+    if (getKV(IDENTITY_PUB_KV) === pub) return; // already published this exact key
+    const { error } = await client
+      .from("identity_pub")
+      .upsert({ public_key: pub }, { onConflict: "user_id" });
+    if (error) {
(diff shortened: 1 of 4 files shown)


BYK:
## Adversarial + SECURITY review — SHIP-WITH-FOLLOWUPS, no BLOCKERs, no private-key leak

Read-only review + mutation testing in a throwaway worktree (main tree sha256-verified identical). All 8 points **PASS**: publishes the PUBLIC key only (never `.secretKey`), encryption-state gated, correct `user_id` via default (RLS makes a wrong-user write impossible), KV-gate correct, best-effort isolation, 0026 recast safe/lossless on the empty table, wire-convention consistency, discriminating tests.

Mutations: publish `.secretKey` → **KILLED**; drop KV gate (churn) → **KILLED**; remove 0026 recast → **KILLED**; **drop the `encryptionState` gate → SURVIVED** (coverage gap).

### Fixed the surviving mutation
The reviewer flagged that the `encryptionState() === "on"` gate is load-bearing beyond skipping: in the **"off"** state, `getAccountIdentity()` would **auto-mint + persist a new escrow-less (non-recoverable) identity** as a side effect and publish it. Added a test asserting `publishIdentityPub` is a no-op for **off** (and specifically does **not** mint an `account_identity` row) **and locked** — this kills the surviving mutation.

### Other follow-ups
- NIT (done): noted in 0026 that the `alter column ... type` recast is single-run.
- NICE-TO-HAVE (deferred): namespace the KV gate by `user_id` to close two degenerate stale-skip edges (remote row deleted / account-switch with same identity) — both effectively unreachable in the one-identity-per-account model. Left as a follow-up.

No private-key leak, no wrong-user publish, no sync-abort path. Verification: engine + identity-scopekeys integration **23 passed**, unit sync **68 passed** (reviewer), full local integration **131** + unit sync **144** green (me); typecheck 0; Biome clean.

PR #1263, opened 2026-07-11
feat(supabase): enforce_row_quota membership guard + effective_tier (E-4b, #827)

BYK:
Second half of the deferred **E-4** team gates (stacked on E-4a #1262). Together with 0027 (the write-gate), this is the 🔴 prerequisite that must be in place *before* any cross-user `scope_members` row exists (E-4c).

## What (migration 0028)
`enforce_row_quota` reproduced **verbatim** from 0016 with exactly **two hunks** changed (the function diff is precisely these two + a guard comment):

1. **Oracle-close guard:** `scope_id is distinct from auth.uid()` → `not public.is_member(new.scope_id)`. The old self-guard skipped the quota read for *any* `scope_id ≠ auth.uid()` — so a team member's write to a **shared** scope was **unmetered**. Now a member's write is metered; only a row in a scope the caller does *not* belong to is skipped (that row is an at-cap existence oracle we must not answer — RLS `WITH CHECK` rejects it regardless).
2. **Tier read:** `profiles.tier where id = new.scope_id` → `effective_tier(new.scope_id)`, which resolves a **team** scope's caps via its org (a team scope id has no `profiles` row → the old read would misresolve to `'free'`).

`maintain_usage` has no self-guard (it always counts by `scope_id`), so the guard change alone is sufficient to meter team writes — no `maintain_usage` change needed.

## Behavior-preserving
Every scope today is personal: `is_member(own) = true` (guard never skips, exactly as before) and `effective_tier(own) = profiles.tier`. All 7 integration suites stay green.

## Verification (Tier-1, real Postgres, manufactured team scope)
`sync-quota-membership.integration.test.ts` (3):
- **Guard:** a team `editor`'s 2nd write over a cap of 1 → `23514` (was skipped/unmetered under the self-guard — the discriminating case).
- **Tier read:** a team scope under a **pro** org keeps writing past the tiny *free* cap → both succeed (proves the cap came from `effective_tier`=pro, not a `'free'` misread).
- **Behavior-preserving:** a personal owner's over-cap write still `23514`.

Suites: security 93 + engine 14 + orgs 9 + membership 7 + identity-scopekeys 8 + write-gate 6 + quota-membership 3 = **141/141**. Function diff vs 0016 = the 2 intended hunks only (verified). Gateway typecheck exit 0; Biome clean.

Adversarial + SECURITY review to follow before merge.

### DIFF
--- a/packages/gateway/test/sync-quota-membership.integration.test.ts
+++ b/packages/gateway/test/sync-quota-membership.integration.test.ts
@@ -1,0 +1,171 @@
+/**
+ * Integration tests for migration 0028 — enforce_row_quota membership guard + effective_tier
+ * (E-4b, #827) against a real Postgres. Proves the two targeted changes:
+ *  - GUARD: a team member's write to a SHARED scope (scope_id ≠ auth.uid()) is now METERED.
+ *    Under the old `scope_id is distinct from auth.uid()` guard it was skipped (unmetered).
+ *  - TIER READ: a team scope's caps come from its ORG's tier via effective_tier (not from a
+ *    profiles lookup by the scope id, which for a team is null → 'free').
+ *  - Behavior-preserving: a personal scope is metered exactly as before.
+ *
+ * Gated behind LORE_INTEGRATION=1 (needs Docker). Run:
+ *   LORE_INTEGRATION=1 pnpm exec vitest run packages/gateway/test/sync-quota-membership.integration.test.ts
+ */
+import { execFileSync } from "node:child_process";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
+import { type PgHarness, startPgHarness } from "./helpers/pg-harness";
+
+function dockerReady(): boolean {
+  try {
+    execFileSync("docker", ["info"], { stdio: "ignore", timeout: 10_000 });
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+const RUN = process.env.LORE_INTEGRATION === "1";
+const SKIP = !RUN
+  ? "LORE_INTEGRATION!=1"
+  : !dockerReady()
+    ? "docker unavailable"
+    : false;
+
+let h: PgHarness;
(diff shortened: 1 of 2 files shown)


BYK:
## Adversarial + SECURITY review — SHIP-WITH-FOLLOWUPS, no BLOCKERs

Read-only review + mutation testing in a throwaway worktree (main tree sha256-verified identical). All 7 points **PASS**. The critical byte-identity check: the `enforce_row_quota` fn body diff vs 0016 is **exactly the two claimed hunks** (guard + tier read) plus the guard comment — no dropped `elsif`, no changed table name, no reordered clause, no lost cast; eviction block, advisory lock, all 6 `exists_pk` branches, both `raise` clauses byte-identical.

Mutations all **KILLED**: revert guard → team-metered test RED; revert tier read → tier-read test RED ("tier free: 1 of 1 rows" — the exact old misread); flip eviction ORDER BY → sync-security "evicts LOWEST-confidence" RED. No unmetered-write hole (non-member skip is RLS-rejected *after* the BEFORE-trigger → never lands, `maintain_usage` never fires); no tier-inflation (`profiles.tier`/`orgs.tier` both service-role-only). Baseline 96/96 green.

### Addressed follow-ups
- **Cross-member eviction seam** (CONCERN): documented in the migration header — free-tier eviction selects the victim by `scope_id` not `author_id`, i.e. a shared per-scope quota sheds the least-valuable row team-wide (intended; unreachable until E-4c).
- **Test gap** (NIT): added a test that a **non-member's forged team-scope write** is RLS-rejected (42501) **and** leaves `user_table_usage` untouched — locks point 2(c).

Deferred (NITs): the now-dead `if user_tier is null` guard (retained verbatim for byte-identity); one extra indexed `scope_members` lookup per write (negligible).

Verification: quota-membership **4/4** (incl. the new forged-write test) + security 93 = green; fn-body diff = 2 hunks only (re-verified). Typecheck 0; Biome clean.

PR #1273, opened 2026-07-11
feat(core): client key rotation — epoch-aware DEKs + rotateScopeKey + seal/open (E-4c-3b, #827)

BYK:
The client crypto that makes key **rotation function**, completing E-4c-3 (the schema landed in 3a, #1271). A rotation mints a fresh DEK, wraps it to the remaining members at a new epoch, seals new content at that epoch, and **still opens old-epoch blobs** (each blob's envelope pins its `key_epoch`).

## keystore (`crypto/keystore.ts`)
- **`dekCache` re-keyed by `(scope, epoch)`** — each rotation epoch has its own DEK.
- **`getScopeKey(scope, member, { mint?, epoch? })`** — `epoch` defaults to the scope's **current** (highest) epoch (what encrypt seals at); decrypt passes a blob's pinned epoch. Mints **only at epoch 0** (origination); a missing rotated epoch throws `ScopeKeyUnavailable` (only `rotateScopeKey` creates those).
- **`scopeKeyEpochs` / `currentScopeEpoch`** helpers drive the sync pre-resolve.
- **`rotateScopeKey(scope, newEpoch, members[{userId, pubKey}])`** — fresh `generateDek()`, wrap to each member (caller **must include self**) at `newEpoch`, INSERT (old epochs retained), cache the new DEK. `newEpoch` is server-atomically allocated by `rotate_scope_key(scope)` (0030).
- `putWrappedScopeKey` invalidates only **this** epoch's cache slot (so a pull-adopt of a canonical wrap after a lost mint race re-unwraps).

## sync.ts (epoch-aware — hot path stays SYNC)
- **`EncCtx` → `{ scope, currentEpoch, deks: Map<epoch, dek> }`**; `ctx()` pre-resolves **every** epoch's DEK once per cycle (few epochs; keystore-cached). This avoids propagating `await` through the pull loop.
- **encrypt** seals at `currentEpoch` (was hardcoded 0); fails **closed** (throws, never pushes plaintext) if the DEK is somehow absent.
- **decrypt** dispatches on the blob's `parseHeader().keyEpoch` → that epoch's DEK; a blob under an epoch we hold no wrap for **defers the table** (`EncryptedContentUnavailable`) — never stores ciphertext, never crashes the cycle.

## remote (migration 0031)
`enforce_row_quota`'s `scope_keys` probe now also matches `key_epoch` (reproduces 0028 **verbatim** + that one line — see the diff), so a rotation's new-epoch row is **counted** against the cap. Now reachable since 3b creates epoch>0 rows (deferred from 0030 as documented).

## Verification
- keystore rotation round-trip: fresh DEK ≠ old; **both** epochs readable; a remaining member unwraps the new epoch; `getScopeKey` with no epoch defaults to current.
- Tier-1: a rotation's new-epoch wrap trips the row cap (23514) under the 0031 probe.
- **Behavior-preserving at epoch 0**: core **2791** / 6 skipped; integration **156** (incl. the encrypted device-1→device-2 engine round-trip); `pnpm -r typecheck` **0**; Biome **0 errors**.

Adversarial + SECURITY review to follow. Next: **E-4c-4** (CLI) wires `rotate_scope_key` + `rotateScopeKey` into `lore team remove-member` (enumerating remaining members + their `identity_pub` keys) — the live caller.

### DIFF
--- a/packages/core/src/crypto/keystore.ts
+++ b/packages/core/src/crypto/keystore.ts
@@ -147,19 +147,50 @@ export function getAccountIdentity(): Keypair {
   return kp;
 }
 
+/** dekCache / pendingScopeKey are keyed by (scope, epoch): each rotation epoch has its OWN DEK. */
+const dekKey = (scopeId: string, epoch: number): string =>
+  `${scopeId}\x1f${epoch}`;
+
+/** All rotation epochs this device holds a wrap for, for (scope, member), ascending. */
+export function scopeKeyEpochs(
+  scopeId: string,
+  memberUserId: string = scopeId,
+): number[] {
+  return (
+    db()
+      .query(
+        "SELECT key_epoch AS e FROM scope_keys WHERE scope_id = ? AND member_user_id = ? ORDER BY key_epoch",
+      )
+      .all(scopeId, memberUserId) as { e: number }[]
+  ).map((r) => r.e);
+}
+
+/** The scope's current (highest) epoch this device can encrypt at; 0 if it holds no wrap yet. */
+export function currentScopeEpoch(
+  scopeId: string,
+  memberUserId: string = scopeId,
+): number {
+  const eps = scopeKeyEpochs(scopeId, memberUserId);
+  return eps.length ? eps[eps.length - 1] : 0;
+}
+
 /**
- * The DEK for a scope. Generated + wrapped to the account key on first use (personal
- * v1: `memberUserId` defaults to `scopeId`, so `scope_id = member_user_id = user`).
- * Otherwise unwrapped from the stored `scope_keys` row. Cached in memory per scope.
+ * The DEK for a scope at a given epoch. Minted + wrapped to the account key on first use
+ * (personal v1: `memberUserId` defaults to `scopeId`). Otherwise unwrapped from the stored
+ * `scope_keys` row for that epoch. Cached in memory per (scope, epoch) — each rotation epoch has
+ * its own DEK. `epoch` defaults to the scope's CURRENT (highest) epoch (what encrypt seals at);
+ * decrypt passes the blob's pinned epoch to open past-rotation ciphertext.
  */
 export function getScopeKey(
   scopeId: string,
   memberUserId: string = scopeId,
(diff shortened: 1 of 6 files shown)


BYK:
## Adversarial + SECURITY review — addressed; SHIP

Read-only review + mutation testing (throwaway worktree, main tree sha256-verified). The production crypto passed all 7 points — **no plaintext leak, no wrong-epoch open, no divergence, fail-closed throughout**; envelope AAD binds the epoch so a wrong-epoch DEK can only throw (never yield garbage); 0031 is byte-identical to 0028 except the one probe line.

### Fixed the BLOCKER-class coverage gap
The reviewer proved (via surviving mutations **M1** = decrypt with `currentEpoch` instead of the blob's epoch, and **M3** = seal without the `{keyEpoch}` stamp) that the PR's *central* behavior — live multi-epoch decrypt — had **zero test coverage**. Added a real-engine round-trip (`sync-engine.integration.test.ts`): device-1 pushes `kr0` at epoch 0, **rotates to epoch 1**, pushes `kr1` at epoch 1; device-2 (fresh) pulls **both** epoch wraps, unlocks, and each blob decrypts with **its own** epoch's DEK.
- Verified it **kills M1** (revert decrypt to `currentEpoch` → RED) and **M3** (the `epochOf(kr1)===1` remote assertion catches a missing stamp).

### Both non-blocking follow-ups also applied
- **`rotateScopeKey` atomicity**: wrap-all-then-persist (`Promise.all` the HPKE wraps first, then the puts) so a mid-wrap failure leaves no partial epoch that a retry would collide with at the remote first-write-wins guard.
- **`ctx()` TOCTOU**: derive `currentEpoch` from the single `scopeKeyEpochs` read (`epochs.at(-1)`) instead of a separate `currentScopeEpoch` query — no window if a rotate lands mid-cycle, one fewer query.

Verification: integration **157** (incl. the new multi-epoch round-trip + all E-4 suites), core keystore/sync **110**, `pnpm -r typecheck` **0**, Biome **0 errors**. E-4c-3 (rotation) is now complete pending merge.

Issue #1307, opened 2026-07-12
E-5-F3: re-enqueue the linked entity graph when knowledge is approved into a team (manual policy)


BYK:
Follow-up from the F3-3 (#1304) adversarial review (finding F-1).

## Gap
`approveForTeam` (ltm.ts) sets `approval_status='approved'` on the knowledge row only. On the next push, F3-3's `teamScopeForContent` migrates that knowledge row personal→team. But for **manual policy**, the linked entity graph (`entities` / `entity_aliases` / `entity_relations` / `knowledge_entity_refs`) was pushed under the **personal** scope at creation time and is **not re-enqueued** on approval — so those rows stay personal-scoped while their knowledge sits in the team scope.

- **Auto policy is unaffected**: knowledge lands `approved` at creation, so the whole graph resolves to the team scope on first push (no migration needed).
- **No producer leak/loss**: the resolver (`teamScopeForContent`) already scopes all five tables correctly; nothing *triggers* the migration for a pre-existing manual-policy graph.
- **Impact is confined to the deferred F2 consumer**: a team member pulling the promoted knowledge would see it with RLS-hidden, orphaned entities.

## Why it's non-trivial
`knowledge_entity_refs` is a join table with no `updated_at` and no UPDATE capture trigger (rows are insert/delete-only), so it can't simply be "touched" to re-enqueue. A correct re-enqueue must also reason about cross-scope refs (a personal knowledge entry may reference an entity that is now team-scoped — the single-scope-per-entity v1 limitation).

## Suggested fix (do before/with F2)
On `approveForTeam(logicalId)`, re-enqueue the linked entity graph so the next push migrates it:
- touch linked `entities` (and their `entity_aliases` / `entity_relations`) via a no-op UPDATE to fire their capture triggers;
- re-enqueue the `knowledge_entity_refs` rows for the approved logical_id (delete+reinsert under suppression, or a dedicated enqueue helper);
- decide the cross-scope-ref policy (relation migrates only when both endpoints share the team — already in the resolver).

Prerequisite for F2 (fresh-member onboarding), which is where the team entity graph is actually consumed.

PR #1310, opened 2026-07-13
feat(sync): decrypt each pulled row with its own scope's DEK (E-5-F2, #827)

BYK:
The **consumer half** of team knowledge sharing (F3-3 #1304 was the producer half). On pull, each row is now decrypted with the DEK for **its own scope** (personal OR a team scope), so a team member can actually read knowledge shared into a team.

## Change
- `pullOnce` resolved **one personal ctx per table** and decrypted every row with it. It now resolves the ctx **per row** from `remote.scope_id` (`ctxForScope(row.scope_id ?? personalScope)`), threaded through the primary loop, the constraint handler, and `drainTimestamp`.
- `decryptColumns` already binds AAD to `ctx.scope` and defers the table (`EncryptedContentUnavailable` → cursor frozen) when the DEK is missing — so a team row this device holds **no wrap** for defers cleanly and re-pulls once the admin's group-wrap arrives, **never stored as ciphertext**.
- The per-table device-level **`locked`** check stays (lock is per-device, not per-scope). Key tables (escrow, scope_keys) pull **before** content, so a team wrap is applied before its content is decrypted; personal rows reuse the already-resolved per-table ctx.
- `getScopeKey(teamScope, self, {mint:false})` unwraps the member's **own** wrap — never mints a divergent team DEK.

## Tests (gateway, faithful mock)
- A team-scoped ciphertext row round-trips to plaintext on pull via the team DEK.
- A device holding **no** wrap for the team defers (cursor frozen, content never stored as ciphertext).

Full gateway + core sync suites green, 0 TS errors, lint clean.

## Follow-up
The full **two-user** team round-trip (admin group-wraps to a *different* member who pulls + decrypts under real RLS) lands as a dedicated E2E **Tier-2** test (F2-2). Its constituent links — E-4c-2 group-wrap, scope_keys pull, `getScopeKey` unwrap — are each already covered independently; this PR proves the final per-row decrypt link. Related: #1307 (manual-approval entity-graph re-enqueue).


BYK:
## Adversarial review: SHIP-WITH-FOLLOWUPS (no blockers)

Security-in-scope adversarial + mutation review (throwaway worktree, main tree sha256-verified unmodified). Per-row/per-scope decrypt is correct and fail-closed: AAD-bound, never stores ciphertext, never advances the cursor on a deferrable row, cross-scope isolation holds (DEK and AAD both derive from `remote.scope_id`), `locked` unchanged. Mutations killed: ignore `scope_id`→personal ctx (happy-path fails), swallow no-key (defer test stores ciphertext), swallow decrypt error (cursor advances).

### Fixes applied this round
- **Multi-epoch team decrypt (SHOULD-FIX, made reachable by this PR):** `ctxForScope` enumerated epochs via `scopeKeyEpochs(scope)` — single-arg defaults `member_user_id` to `scope`, but a team member's wraps are keyed by their own uid, so it returned `[]` and collapsed `deks` to only epoch 0. A **rotated** team's content (epoch ≥1) would then permanently defer even though the device holds the wrap. Fixed to `scopeKeyEpochs(scope, me)` (matches the adjacent `getScopeKey(scope, me, …)`). Added a **discriminating** regression test: after a rotation to epoch 1, the producer seals at epoch 1 (proving `[0,1]` enumeration) and the consumer decrypts it — the buggy single-arg form seals at epoch 0 and the `keyEpoch === 1` assertion fails.
- **NIT:** removed dead `personalScopeId` / `?? personalScopeId` (`asString` returns `""` not `undefined`, so the fallback ran via the `!rowScope` guard anyway).

### Tracked follow-ups (before team DEK rotation is user-facing / F2-2)
- Drain-path (`>PAGE` rows sharing one `updated_at`) team-row coverage (Mutation E gap — correct by construction, uncovered).
- A true **two-user** (other-member group-wrap) Tier-2 integration test for consumer decrypt (F2-2).
- #1307 (manual-approval entity-graph re-enqueue).

PR #1311, opened 2026-07-13
feat(sync): re-enqueue the linked entity graph on team approval (#1307)

### LINKED ISSUE #1307: E-5-F3: re-enqueue the linked entity graph when knowledge is approved into a team (manual policy)

BYK:
Closes #1307. Completes the F3-3 entity-graph migration for the **manual-approval** flow.

## Gap
On `approveForTeam`, the knowledge row's own personal→team migration rides its `UPDATE` (the column-agnostic capture trigger). But its linked entity graph (`entities` / `entity_aliases` / `entity_relations` / `knowledge_entity_refs`) has **no** trigger from that UPDATE — so a manual-policy graph, pushed under the **personal** scope at creation, would stay personal while its knowledge moved to the team. (Auto-policy was already correct: the whole graph resolves to the team on first push, no migration.)

## Fix
- **`ltm.ts`**: `approveForTeam` notifies an `onKnowledgeApprovedForTeam` hook with the logical_id. A hook — not a direct import — because `sync-data.ts` already imports `ltm.ts`; the reverse would be a cycle (same pattern as `onProjectRemoteBackfilled`).
- **`sync-data.ts`**: `reenqueueKnowledgeTeamGraph(logicalId)` re-enqueues the approved entry's linked entities, their aliases, the relations touching them, and its refs — gated on the same parent/remote gates as the capture triggers (mirrors `reseedProjectContent`). The push's `teamScopeForContent` resolver decides each row's **actual** scope (a relation migrates only when **both** endpoints share the team); unchanged rows short-circuit (same content_hash AND scope). Registered at module load.

## Test (gateway, faithful mock)
Approving a manual-policy entry with two linked entities (+ an alias + a relation between them) migrates K **and** the full linked graph (entities, alias, relation, refs) from personal → team on the next push, with no personal duplicate left.

Full sync + core sync suites green, 0 TS errors, lint clean. Follow-up from the F3-3 (#1304) review.

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -174,6 +174,9 @@
 <!-- lore:019f27e4-458e-7d6a-bdfc-5ab2f35c5354 -->
 * **Migrated to settings**: migrated to settings.
 
+<!-- lore:019f5b56-d717-7acf-82cd-ecfc75ca4871 -->
+* **Migrated to TG): test failed at sync**: migrating to TG): test failed at sync.
+
 <!-- lore:019f0103-15cb-708f-9cff-f5c8f060d37a -->
 * **Migrated to Vitest in May 2026 — 312ms vs 30s startup time cited as reason**: migrated from Mocha + Chai + ts-node to Vitest in May 2026 — 312ms vs 30s startup time cited as reason.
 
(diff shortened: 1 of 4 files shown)


BYK:
## Adversarial review: SHIP-WITH-FOLLOWUPS (no blockers)

Correctness + mutation review (throwaway worktree, main tree sha256-verified unmodified). Hook wiring is cycle-free and fires only on real approvals; the re-enqueue SQL's row_id/placeholder/gate all match the capture triggers and `reseedProjectContent`; idempotent (unchanged rows short-circuit); resolver-driven single-scope semantics hold. All 5 targeted mutations killed (neuter fn; drop each of the 4 enqueues; corrupt the `char(31)` separator).

### Addressed this round
- **Demotion symmetry (review CONCERN):** my PR made *promotion* reliable but left `rejectForTeam` asymmetric (entity graph stranded in the team scope after approve→reject). Fixed: renamed the hook to `onKnowledgeTeamPromotionChanged` and fire it from **both** `approveForTeam` and `rejectForTeam`, so the graph migrates INTO the team on approve and BACK to personal on reject (the resolver decides — an entity still linked to other team-approved knowledge stays team). Test extended to assert the full round-trip: approve → team, reject → back to personal, no leftovers.

### Tracked follow-ups
- **#1312**: the broader `remove()`/deletion demotion case (deliberately not bundled — `remove()` fires in many non-team contexts and needs its own gating/tests), plus a mixed-scope-relation re-enqueue test and a negative parent-gate test. (The "both endpoints" relation rule and the gates are already covered at the resolver level in `scope-selection.test.ts` and verbatim from `reseedProjectContent`.)

Full sync + core suites green (277), 0 TS, lint clean.

Issue #1345, opened 2026-07-16
Revoke outstanding invite tokens on member removal (E-5-c F-2, #827)


BYK:
Follow-up from the E-5-c-1 adversarial review (F-2).

**Problem:** `accept_scope_invite` tokens are single-use (deleted on accept) and expire in 14 days, but `remove_scope_member` does NOT delete a removed member's outstanding *unredeemed* invite tokens. A removed member holding a live token (or a leaked token) can rejoin the scope as editor/viewer within the 14-day window.

**Severity:** low. Membership grants FETCH only; DECRYPT requires a DEK wrap, and the F-1 fix (migration 0043, `scope_keys_target_member`) now blocks re-wrapping the rotated DEK to a non-member — so a rejoined-via-stale-token user cannot read post-rotation content. The residual is a roster/re-join nuisance, not a content-exposure path.

**Fix options:**
- On `remove_scope_member`, `DELETE FROM pending_invites WHERE scope_id = p_scope` (revokes ALL outstanding tokens for the scope — simplest, safe), or scope to the removed user if invitee identity is known.
- Or document the 14-day capability-token window explicitly as accepted behavior.

Note: `pending_invites` already `ON DELETE CASCADE`s on `scope_id`, so scope deletion clears tokens; only member *removal* is uncovered.

Ref: PR for E-5-c-1 (`feat/827-e5c-email-invite`).

Issue #1294, opened 2026-07-12
Registry mirror: no remote-delete propagation for orgs/scopes/members (E-5 follow-up)


BYK:
Follow-up from PR #1290 (E-5-F1 registry mirror), flagged in adversarial review as a deferred CONCERN.

## Problem

The pull-only registry mirrors (`orgs`/`org_members`/`scopes`/`scope_members`, added in #1290) have **no tombstone / absence reconciliation**. When a membership is removed remotely (`remove_scope_member`, 0029) or a scope/org is deleted, the local mirror row **persists** — pull-only tables never observe deletions (client reconcile-by-absence is forbidden by the eviction model, `sync.ts` ~1080).

## Why it's not a security hole (today)

Server RLS + group-DEK re-wrap (rotation on removal) remain the real access gate: a removed member can neither pull nor decrypt the scope's content. The stale local membership row is cosmetic — it does not grant access.

## Why it still needs handling before F2/F3

Once a consumer (F2 onboarding / F3 scope-selection, or `lore team members`) reads the mirror to decide which scopes to show/operate under, a stale row would surface a scope the user was removed from. Needs a bounded propagation mechanism consistent with the eviction model — e.g. a server-side "membership epoch"/watermark the client can diff, or a targeted pull-refresh of the member's full scope set on each sync (small, RLS-scoped) that replaces rather than upserts.

## Acceptance

- Removing a member remotely eventually removes the row from that member's local mirror (bounded latency), without client reconcile-by-absence.
- Covered by a Tier-2 integration test (remove → next pull → local row gone).

PR #1346, opened 2026-07-16
feat(sync): direct email team invites with zero-follow-up DEK sharing (E-5-c-1, #827)

BYK:
## What
Direct email invites for team scopes (E-5-c-1, epic #827) — onboard a member who isn't yet a `user_id`, with **no manual follow-up by either party**.

- `lore team invite <scope> [--role editor|viewer] [--email <hint>]` → an admin mints a capability token.
- `lore team accept <token>` → the invitee self-joins + publishes their identity key.
- The admin's normal idle-sync loop then **auto-wraps** the team DEK to the new member (`reconcileScopeWraps`, hooked into `syncOnce`). The invitee fetches immediately and decrypts once the admin's next sync lands (seconds–minutes) — nobody runs a follow-up command.

Server-side wrap is impossible in our E2E model (the server never holds the plaintext DEK), so the admin's client does the wrap; the auto-wrap makes it invisible.

## How
**Migration 0042** — `pending_invites` + two SECURITY DEFINER RPCs:
- `create_scope_invite`: admin-only (`scope_role='admin'` → 42501), role ceiling editor/viewer only (never admin → 22023), server-generated unguessable token (two `gen_random_uuid()`s, ~244 bits), 14-day expiry.
- `accept_scope_invite`: self-adds **only** `auth.uid()`, additive (`on conflict do nothing` — never downgrades an existing role), single-use, **identical error for not-found vs expired** (no token/email oracle).
- `pending_invites` is RPC-only: RLS-enabled with **no policies** (deny-all); a registry-contract test asserts it stays out of `SYNCED_TABLES` so it can't wedge the outbox prune floor.

Membership grants **FETCH only** (RLS `is_member`); DECRYPT needs the DEK wrap — a leaked/forged token can only JOIN, never read content.

**Migration 0043** (adversarial-review F-1 fix) — `scope_keys_insert/update` WITH CHECK now also require the **target** `member_user_id` be a current member (`scope_has_member`). Previously only the *writer* was gated as admin, so the auto-wrap reconcile could re-wrap the rotated DEK to a member removed on another admin's device but still lingering in a second admin's stale local mirror (#1294) — defeating removal forward secrecy. Now rejected server-side regardless of mirror staleness.

## Tests
11 discriminating integration tests (real Postgres/PostgREST): happy path, admin-only mint, role ceiling, single-use, expiry (adversarial: expired before accept), no-oracle, additive-not-downgrade, org membership, cross-scope isolation, RLS deny-all, end-to-end auto-wrap reconcile, and the F-1 non-member-rewrap rejection. Each verified discriminating (removing a guard fails its test). Full team/membership regression (33 integration + 224 unit) + typecheck + lint clean.

## Review
Adversarial correctness+security review: **SHIP-WITH-FOLLOWUPS**. Invite path fully sound (no escalation, no oracle, correct definer-auth, true deny-all). F-1 fixed here (0043). F-2 (revoke unredeemed tokens on removal — low severity, FETCH-only now that F-1 closes decrypt) filed as #1345.

## Follow-ups
- #1345 — revoke outstanding invite tokens on member removal (F-2)
- E-5-c-2 — `--offline` ephemeral-key escape hatch + rotate-on-accept (the `eph_pub` column lands here unused, ready)

Part of epic #827.


BYK:
Addressed the Sentry MEDIUM: widened `eph_pub` cap 64→512 chars (matching the `identity_pub` precedent) for HPKE-algorithm headroom. The column is unused today (E-5-c-2 not yet built), so this is a free change that avoids a future migration.

BYK:
Addressed both Sentry MEDIUMs:
- **eph_pub cap**: widened 64→512 chars (HPKE-algorithm headroom, matches identity_pub precedent).
- **personal-scope invites**: `create_scope_invite` now rejects any scope where `kind != 'team'` (22023) — a personal scope is single-user, so inviting a second member would break its isolation. Added a discriminating regression test.

PR #1352, opened 2026-07-16
feat(sync): offline (--offline) team invites with ephemeral key delivery (E-5-c-2, #827)

BYK:
## What
The admin-never-returns escape hatch on top of the E-5-c capability invite (#1346). `lore team invite --offline` additionally mints an **ephemeral X25519 keypair**, wraps the scope DEK to it (an `eph:<pub>` scope_keys row, pushed), and embeds the ephemeral SECRET in the token (`<capability>.<base64url(secret)>`). A **fresh invitee** can then unwrap + adopt the DEK and read immediately — the admin can be offline forever, zero follow-up on either side.

## Retirement (design decision vs. plan's rotate-on-accept)
The invitee joins as **editor/viewer**, and `rotate_scope_key` is **admin-only** (42501 for non-admins; its re-wrap step also needs admin write on co-member wraps). So instead of rotating, the invitee **retires the spent ephemeral wrap**: deletes the `eph:` row locally (capture-suppressed) + remotely (`retire_ephemeral_invite`, a member-callable definer RPC). Since the token is already single-use + short-lived + member-gated read, deleting the spent wrap fully closes the ephemeral secret — no epoch churn, no admin privilege, DEK unchanged.

## Migration 0044
- `scope_key_target_ok()` — scope_keys writes may target a current member **OR** an `eph:%` capability key (still admin-only mint, so grants no new capability).
- `scope_keys_read` widened so a member can read `eph:` rows for their scope (unreadable without the token secret).
- `retire_ephemeral_invite(scope, eph_pub)` — member-callable, deletes **only** the exact `eph:<pub>` row, never a real wrap.

## Core keystore (pure crypto + local DB)
`mintEphemeralInviteWrap` / `adoptEphemeralInviteWrap` / `deleteEphemeralInviteWrap`. Adopt re-wraps the DEK to the invitee's own identity under capture-suppression (self-wrap is admin-only to push → stays local; the admin's reconcile writes the remote copy later).

## Client + CLI
`createTeamInvite({offline})`, `acceptTeamInvite` detects the eph suffix and adopts+retires; `lore team invite --offline` warns the token carries a one-time key.

## Security invariants
- Ephemeral **secret never touches the server** — only the DEK wrapped to the eph *public* key is stored; the secret rides the token only.
- Widened read exposes only ciphertext unreadable without the token secret.
- `retire_ephemeral_invite` is member-gated (`is_member`) and exact-match `eph:`-only — a member can never delete another member's real wrap.
- A capability (non-offline) token has no `.` suffix and mints no eph wrap.

## Tests
- **keystore-group-wrap** (core): mint→adopt→delete round-trip proves B holds the SAME DEK A minted; wrong-secret HPKE failure; absent-wrap error.
- **team / team-cmd** (unit): offline token shape, `p_eph_pub`, `--offline` passthrough + warning.
- **sync-team-invite** (integration): offline adopt delivers the SAME DEK, eph wrap retired both sides, member-callable retire is eph-only, capability token mints no eph wrap.

Full typecheck/lint/format clean; 124 integration + 195 unit tests pass. Follow-up to #1346, part of epic #827.


BYK:
Adversarial correctness+security review complete: **SHIP-WITH-FOLLOWUPS, no BLOCKERs.** All security gates verified (eph secret never reaches server; writes admin+author-gated; eph reads member-gated ciphertext-only; retire is member-gated + exact-`eph:`-only with no injection; invitee self-wrap capture-suppressed so no cursor wedge; HPKE fails closed; migration idempotent; tests discriminating).

Addressed in this branch:
- **SHOULD-FIX:** corrected stale rotation comments (keystore.ts / team.ts) that described accept-side rotation which no longer happens — now accurately document the delete-not-rotate retirement + residual (leaked token only exploitable by an existing member who already holds the DEK).
- **NIT:** added a retire test asserting a real member's uuid passed as `p_eph_pub` retires only `eph:<uuid>`, never the bare real wrap.

Deferred (non-blocking): eph-row lifecycle reaping for unaccepted/orphaned invites → filed as #1353.

BYK:
Addressed the Seer MEDIUM (orphaned eph row on create_scope_invite failure): reordered `createTeamInvite` to be **RPC-first**. The ephemeral keypair is now generated up front (no DB write) so the RPC can record `eph_pub`, but the DEK wrap is stored + pushed **only after `create_scope_invite` succeeds**. An RPC failure short-circuits before any eph `scope_keys` row is created or pushed — no orphan can accumulate. `mintEphemeralInviteWrap` now takes the caller-supplied keypair (returns void). Tests updated; 79 unit + 15 integration green.

BYK:
Addressed the second Seer MEDIUM: `acceptTeamInvite` now throws when a token carries an ephemeral secret but the server returns `out_eph_pub = null` (mismatch/corruption) — instead of silently returning "joined" while never adopting the DEK. Added a unit test. (The first Seer MEDIUM was already marked resolved by the RPC-first reorder.)

PR #1354, opened 2026-07-16
feat(sync): domain auto-join (Model B: verified-admin claim + approve-on-join) (E-5-b, #827)

BYK:
## What
The second E-5 provisioning mechanism (after E-5-a GitHub Teams and E-5-c email invites). **Model B** domain auto-join:
1. An org admin/manager **claims** an auto-join domain, proving control via their **own verified email** at that domain (freemail domains rejected).
2. A user whose **verified email** matches a claimed domain **requests** to join.
3. An admin **approves** — additive (never demotes an existing higher role).

No silent absorption, no public-domain hijack; pure email-match with no control proof is explicitly rejected. DNS-TXT (Model A) zero-touch is a later upgrade.

## Migration 0045
- `org_domains` + `domain_join_requests` — RPC-only writes (deny-all DML), reads RLS-scoped.
- `email_domain(uid)`: self-only oracle guard (cross-user probe → NULL unless service_role, mirrors 0032). `verified_email_domain(uid)`: unguarded helper for the admin-gated approve re-check — **explicitly REVOKEd from public/anon/authenticated** (the 0039 default-PUBLIC-EXECUTE hazard) so it is never client-callable.
- `is_freemail` blocklist; `claim/request/approve/reject_domain_join` RPCs (SECURITY DEFINER, pinned search_path, org-advisory-locked, admin-gated with `coalesce` to avoid the NULL-`not in` SQL trap; approve re-verifies the requester's email at approval time).

## Client + CLI
- `team.ts` thin RPC wrappers (org membership only — **no DEK/crypto**).
- `lore team domain <claim|request|requests|approve|reject>` — **not** gated behind encryption/sync.

## Test harness
Added `email_confirmed_at` to the auth.users shim + a `confirmed` flag on pg-harness `createUser` (default true).

## Tests
- **sync-domain-join integration (11):** admin-gate / freemail / own-domain-proof on claim; request needs a claim + verified email (identical error → no oracle); unverified email can't request; approve admin-only + additive + re-verify; reject; and `verified_email_domain` is not client-callable while `email_domain` pins a cross-user probe to null.
- **team / team-cmd unit (18):** RPC wrappers + CLI dispatch/validation; domain verbs need neither encryption nor sync.
- **sync-registry-contract:** `org_domains`/`domain_join_requests` stay OUT of SYNCED_TABLES (server-internal; would wedge the prune floor, #828).

Full typecheck/lint/format clean; 156 unit + 11 domain integration pass (133 total integration green). Part of epic #827.

### DIFF
--- a/packages/core/test/sync-registry-contract.test.ts
+++ b/packages/core/test/sync-registry-contract.test.ts
@@ -253,4 +253,17 @@ describe("server-internal tables are intentionally NOT synced", () => {
         `pending_invites must NOT be in SYNCED_TABLES.${tier} — it is a server-internal, RPC-only table (deny-all RLS, no local mirror). Syncing it would wedge the outbox prune floor.`,
       ).toBeUndefined();
   });
+
+  // org_domains / domain_join_requests (E-5-b, #827): same rationale as pending_invites — written/read
+  // only via SECURITY DEFINER RPCs (claim/request/approve/reject_domain_join), no local mirror, and
+  // (for writes) deny-all under RLS. They MUST stay out of SYNCED_TABLES or the client-unpushable
+  // rows would pin the outbox prune floor at 0 for every table (#828 wedge).
+  test("domain auto-join tables are absent from every tier's registry", () => {
+    for (const tier of ["basic", "pro", "max"] as const)
+      for (const table of ["org_domains", "domain_join_requests"])
+        expect(
+          SYNCED_TABLES[tier].find((m) => m.table === table),
+          `${table} must NOT be in SYNCED_TABLES.${tier} — server-internal, RPC-only (no local mirror). Syncing it would wedge the outbox prune floor.`,
+        ).toBeUndefined();
+  });
 });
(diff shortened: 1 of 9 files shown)


BYK:
Adversarial correctness+security review complete: **SHIP-WITH-FOLLOWUPS, no BLOCKERs** (5 mutations all killed, including the load-bearing `verified_email_domain` revoke, the additive-no-demote guard, the NULL-not-in coalesce, and the no-oracle identical-error property).

Addressed both non-blocking follow-ups in this branch:
1. **TOCTOU in `approve_domain_join`** — now re-reads `rec` under the org advisory lock (mirrors `accept_scope_invite`) so a concurrent approval can't stamp a stale `decided_by` audit trail. (Membership insert was already additive, so no escalation/demotion was ever possible — this only fixes the audit trail.)
2. **Test regexes** — dropped the `|42501` SQLSTATE alternation so the security assertions match message text only (stronger discrimination).

11 domain integration + 156 unit tests still green; typecheck/lint/format clean.

PR #1364, opened 2026-07-17
feat(cli): `lore admin grant` operator command for tier promotion (#827)

BYK:
## What
An **operator-only** command to promote a personal account or team org to a paid tier — the manual lever behind entitlement, for staff use until billing (Stripe) is wired.

```
lore admin grant <email|orgId> <free|pro|team>
```

- **email** → resolved via `profiles` (service-role read) → flips `profiles.tier` (personal). Refuses on no-match and on **ambiguous** multi-match (never guesses).
- **UUID** → flips `orgs.tier` by org id (team). Refuses on no-match.

`effective_tier()` reads profiles for personal scopes and orgs for team scopes, so this single lever drives entitlement.

## Why service-role
The tier guards (`guard_profile_tier` 0006, `guard_org_tier` 0023) deliberately block clients from self-upgrading — `service_role` is the sole writer. So a manual promotion must go through service-role.

`getServiceRoleClient()` (supabase.ts) builds an **in-memory** client from `SUPABASE_SERVICE_ROLE_KEY` **at call time** — never persisted, never shipped (only the anon key ships; RLS enforces normal users). Returns null when the env var is unset → the command prints a clear "operator-only" message instead of a 401.

`admin` is added to `knownCommands` (recognized, not "did you mean") but kept **out of user-facing help** — a hidden staff command.

## Tests (admin-cmd unit, 10)
usage/arg validation, unknown-tier rejection, missing-key guard, email→profiles promote, UUID→orgs promote, not-found + ambiguous-email + update-error paths.

Typecheck/lint/format clean; 10 unit tests pass. No schema change. Part of epic #827.

### DIFF
--- a/packages/gateway/src/cli/admin-cmd.ts
+++ b/packages/gateway/src/cli/admin-cmd.ts
@@ -1,0 +1,121 @@
+/**
+ * `lore admin` — OPERATOR-ONLY commands for the platform maintainer, gated by the
+ * `SUPABASE_SERVICE_ROLE_KEY` env var (which only staff hold; it is never shipped
+ * or persisted). These perform privileged writes the RLS/tier guards deliberately
+ * block for normal users.
+ *
+ *   lore admin grant <email|orgId> <free|pro|team>
+ *
+ * Personal upgrade (an email) flips `profiles.tier`; a team upgrade (an org UUID)
+ * flips `orgs.tier`. `effective_tier()` reads profiles for personal scopes and
+ * orgs for team scopes, so this is the single lever behind entitlement.
+ */
+import { getServiceRoleClient } from "../supabase";
+
+const USAGE =
+  "Usage: lore admin grant <email> <free|pro>   (personal)\n" +
+  "       lore admin grant <orgId> <free|team>  (team org)";
+// Valid tiers per target type. effective_tier() reads profiles.tier for personal scopes (free/pro)
+// and orgs.tier for team scopes (free/team). Cross-pairing (e.g. an email → 'team') writes a tier the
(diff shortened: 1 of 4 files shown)


BYK:
Adversarial correctness+security review complete: **SHIP-WITH-FOLLOWUPS, no BLOCKERs** (M1/M2/M3/M5 mutations all killed; service-role key verified env-only + never persisted/logged; RLS/tier guards respected; errors surfaced + exitCode set on every failure path; command hidden from user help).

Addressed the follow-ups in-branch:
1. **`.ilike` → `.eq`** for the email lookup — removes wildcard semantics so a `%`/`_` in the operator's input can't silently promote an unintended single-match account.
2. **Tier/target coherence** — now enforces the pairing: an **email (personal)** target accepts `free|pro`, an **org UUID (team)** target accepts `free|team`. This prevents the incoherent `profiles.tier='team'` state (server treats it as unlimited, client maps it to basic).
3. **Anchoring regression test** — an email that *contains* a UUID substring must route to the profiles path (closes the surviving-mutant gap where dropping `^`/`$` would misroute).
4. **Defense-in-depth** — a "new caller = security-review trigger" doc comment on `getServiceRoleClient()` (it bypasses RLS).

13 unit tests (was 10); typecheck/lint/format clean.

BYK:
Addressed the remaining Seer LOW: the profiles update now uses `.eq(...).select("id")` and errors (instead of printing a misleading success) if the update hits no row — guards the silent no-op where the account is deleted between the lookup and the write. Added a regression test. 14 unit tests; typecheck/lint/format clean.

Issue #1343, opened 2026-07-15
Distiller re-ingests huge verbatim user blobs (85K in → 571 out), inflating worker cost


BYK:
## Summary

Discovered while auditing eval cost (#961). The segment distiller feeds **entire user-message content verbatim** to the worker LLM, with no cap on oversized pasted blobs. When a user message contains a large pasted blob (logs, a dumped file, a data export), the distiller reads the whole thing to emit a tiny observation.

## Evidence

From a real Sonnet cross-session eval run's `gateway.log`:

```
distill segment: 1 msgs, 85568→571 tokens
distill segment: 1 msgs, 85404→402 tokens
distill segment: 3 msgs, 85456→174 tokens
```

16 distill calls consumed **769,602 input tokens total** — because each ~85K blob message was sent to the distiller whole. On the answering model (Sonnet, \$3/M input) that's ~\$2.31 of worker spend for one session; it was **32% of the entire Lore bill** for that run (\$1.85 of \$5.86).

## Root cause

`packages/core/src/distillation.ts` `messagesToText()` (~line 379):

```ts
// User text is always signal — never truncate.
const body =
  m.role === "user"
    ? m.content
    : truncateToolOutputsInContent(m.content, cap);
```

Tool outputs are capped via `toolOutputMaxChars`, but **user text is never truncated**. The "user text is always signal" assumption holds for prose/directives but not for large pasted blobs (dumped files, logs, data). Those are bulk, not signal — and `detectAssertions()` already pre-scans and pins the actual user directives, so the raw blob body is largely redundant to the distillation objective.

## Impact

- Inflates background worker cost roughly in proportion to blob size.
- Worse when the worker runs on an expensive answering model, but wasteful on any worker model.
- The emitted observation is tiny (571 tokens from 85K in), confirming almost none of the blob was signal worth distilling.

## Proposed fix

Cap oversized **user** message bodies in `messagesToText` before distillation, analogous to `truncateToolOutputsInContent` but for user blobs:

- Add a `userBlobMaxChars` config (generous default, e.g. keep prose but truncate a single message body beyond ~8-16K chars with a `[… N chars elided …]` annotation).
- Keep `detectAssertions` pinning intact so genuine directives inside a large message are still surfaced verbatim.
- Never truncate short/normal user messages (the "always signal" rule still holds for ordinary prose).

## Follow-up (separate, eval-config)

[...]

PR #1367, opened 2026-07-17
fix(distill): preserve leading user prose during blob reduction (#1343 follow-up)

BYK:
## Summary
Fixes a memory-retention failure discovered while auditing the #961 eval: facts a user states as a casual aside at the **head** of an oversized message are dropped from durable memory before the distiller ever sees them.

## Root cause
The #1351 blob reducer (`reduceBlob`) segments an oversized user message and keeps only the top-N segments by cosine relevance to the distillation query. When a user writes their own words (instruction + asides / stated facts) at the head of a message and pastes a large blob after, that leading prose competes with the blob on relevance and — being low-relevance to the current coding task — gets elided. The fact never reaches distillations/knowledge.

Confirmed empirically: in a Sonnet long-session eval, the turn-1 aside `(…our orders ride the WHOLESALE channel out of EMEA, ship from WH-07, keep status uppercase…)` is segment index 0, non-junk, yet dropped whenever the relevance budget is tight — so the distillations captured the API-spec blob but none of the 4 probe facts, even in a run that happened to score 4/4 (that pass came from the raw conversation window, not memory).

## Fix
Force-keep the leading prose prefix (until the first paste-junk segment or a bounded budget) exactly like a pinned assertion:
- New config `knowledge.distillation.userBlobHeadChars` (default **1500**, 0 disables).
- `reduceBlob` marks leading non-junk segments as head-kept; they bypass relevance scoring, the segment cap, and the keepChars budget.
- Bounded by design, so a pure blob with no prose prefix still reduces normally — #1343's cost intent is preserved.

## Validation
- With a **zero-relevance fake embedder**, all 4 facts survive purely on head preservation (independent of relevance score).
- Mutation test (`headChars=0`) confirms the new regression test is non-vacuous.
- `blob-select` 38/38, `distillation` 122/122, `config` 33/33, tsc clean.

Follow-up to #1343 / part of #961.

### DIFF
--- a/packages/core/src/blob-select.ts
+++ b/packages/core/src/blob-select.ts
@@ -234,6 +234,15 @@ export interface ReduceBlobOptions {
    * Empty/whitespace entries are ignored; a trailing display "…" is tolerated.
    */
   pinnedLines?: string[];
+  /**
+   * Chars of the leading prose prefix always kept verbatim, before any scoring.
+   * A user writes their own words (instruction + casual asides / stated facts)
+   * at the head of a message and pastes the bulk blob after; that prose is
+   * signal even when it scores low against the current objective. Leading
+   * segments are force-kept until the first paste-junk segment or this budget is
+   * reached. 0 (or omitted) disables head preservation. #1343 follow-up.
+   */
+  headChars?: number;
 }
 
 export interface ReduceBlobResult {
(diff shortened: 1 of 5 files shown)


BYK:
Adversarial review pass — findings addressed:

- **SHOULD-FIX (segment-cap defeat under large `headChars`):** added a hard clamp — head segments are now capped at `maxSegments/2` (`maxHeadSegments`), so a misconfigured large `headChars` can never force-embed the whole body and reintroduce the #1343 unbounded-embed cost. At least half the embed budget is always reserved for body relevance sampling.
- **SHOULD-FIX (stale comment):** updated the unconditional-keep comment — it no longer claims `MAX_PINNED_ASSERTIONS` bounds the force-kept set; now correctly notes both pins (MAX_PINNED_ASSERTIONS) and head (maxHeadSegments ≤ maxSegments/2) bounds.
- **Coverage gaps:** added two regression tests — (1) head clamped to maxSegments/2 (cost bound holds under absurd headChars), (2) head walk stops at the first paste-junk segment. Both mutation-verified non-vacuous (each fails when its guard is removed).

Review's finding #8 (gradient.ts in the change set) was a local uncommitted WIP artifact — the pushed commit is and remains the 5 head-keep files only. Verdict was SHIP-WITH-FOLLOWUPS with no default-path bug; the two SHOULD-FIX items are now folded in rather than deferred.

blob-select 40/40, distillation 122/122, config 33/33, tsc + check:docs clean.

PR #1372, opened 2026-07-18
fix(sync): propagate remote registry-mirror deletions via authoritative snapshot (#1294)

### LINKED ISSUE #1294: Registry mirror: no remote-delete propagation for orgs/scopes/members (E-5 follow-up)

BYK:
Closes #1294 (E-5-F1 registry-mirror follow-up).

## Gap
The pull-only registry mirrors (`orgs`/`org_members`/`scopes`/`scope_members`, #1290) had **no remote-delete propagation**. `remove_scope_member` (0029) hard-DELETEs remotely with no tombstone, and the keyset delta pull can never observe an absence — so a removed member's local mirror row **persisted forever**. Cosmetic today (RLS + group-DEK rotation are the real access gate), but now that **F2 onboarding (#1310) reads the mirror**, a stale row would surface a scope the user was removed from.

## Fix — authoritative full-set snapshot replace (issue's Option 2)
These four tables are tiny and fully RLS-scoped to the member, so the server's RLS result **is** the member's complete current set. New `mirrorSnapshot` flag on their `SyncTableMeta`:
- `pullOnce` **skips** them from the keyset delta loop;
- `refreshRegistryMirror(client)` fetches each table's full RLS set and calls `replaceRegistryTable` — **upsert-all + delete-absent, capture-suppressed**;
- runs in `syncOnce` after `pullOnce` and **before `reconcileScopeWraps`** (so an admin never re-wraps a DEK to a just-removed member).

This is **not** the forbidden reconcile-by-absence of an *evicted* content table (where absence is ambiguous with local-cache eviction): mirrorSnapshot tables are never evicted and the RLS result is authoritative. A partial/failed fetch **skips** the table (never deletes) and retries next tick. Local mirror tables carry no cross-table FKs → per-table replacement is FK-safe.

## Tests
- **Tier-2** (sync-engine integration): remove member remotely → next refresh → local row gone, **admin's own row survives** (the acceptance criterion). Existing registry-mirror test rewired `pullOnce`→`refreshRegistryMirror`.
- **Unit** (sync-data): `replaceRegistryTable` deletes-absent/upserts-rest, empty set clears the mirror, capture-suppressed (no outbox), refuses a non-mirrorSnapshot table.
- **registry-contract**: `mirrorSnapshot` implies `pullOnly`.

No schema change. Full suite green (5957); typecheck/lint/format clean. Part of epic #827.

### DIFF
--- a/packages/core/src/sync-data.ts
+++ b/packages/core/src/sync-data.ts
@@ -140,6 +140,17 @@ export interface SyncTableMeta {
    * hits the existing row (unique_violation 23505) is treated as already-synced. Defaults to false.
    */
   insertOnly?: boolean;
+  /**
+   * #1294: this pull-only mirror is refreshed by an AUTHORITATIVE FULL-SET SNAPSHOT each sync
+   * (`refreshRegistryMirror` in sync.ts → `replaceRegistryTable` here), NOT the keyset delta pull
+   * (which is skipped for it). The remote RLS result IS the member's complete current set for these
+   * tiny org/scope registry tables, so a row ABSENT from the snapshot means the membership was
+   * removed remotely (`remove_scope_member` hard-DELETEs, leaving no tombstone) → the local row is
+   * deleted. This is NOT the forbidden "reconcile-by-absence" of an evicted content table (where
+   * absence is ambiguous with local-cache eviction); here the small table is never evicted and the
+   * server returns the authoritative complete set. Implies pullOnly. Defaults to false.
+   */
+  mirrorSnapshot?: boolean;
 }
 
 /** ASCII Unit Separator — joins composite row ids (mirrors the SQL `char(31)`). */
@@ -432,6 +443,7 @@ export const SYNCED_TABLES: Record<SyncTier, SyncTableMeta[]> = {
       ftsTables: [],
       versioned: false,
       pullOnly: true,
+      mirrorSnapshot: true,
       syncColumns: [
         "id",
         "kind",
@@ -448,6 +460,7 @@ export const SYNCED_TABLES: Record<SyncTier, SyncTableMeta[]> = {
       ftsTables: [],
       versioned: false,
       pullOnly: true,
+      mirrorSnapshot: true,
       syncColumns: ["org_id", "user_id", "role", "created_at", "updated_at"],
     },
     {
@@ -456,6 +469,7 @@ export const SYNCED_TABLES: Record<SyncTier, SyncTableMeta[]> = {
       ftsTables: [],
       versioned: false,
       pullOnly: true,
+      mirrorSnapshot: true,
       syncColumns: [
         "id",
         "org_id",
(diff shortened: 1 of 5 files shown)


BYK:
Adversarial correctness review complete: **SHIP-WITH-FOLLOWUPS, no BLOCKERs** — delete-by-absent verified correct + FK-safe + empty/partial-safe + idempotent + ordered before reconcileScopeWraps; mutation-killed on both the unit test and the real-Postgres Tier-2 test.

Addressed both follow-ups in-branch:
1. **Vacuous capture-suppression test fixed** — the reviewer correctly flagged that the original 'never enqueues an outbox row' test passed even with `withApplying` removed (registry tables have no production capture trigger). Rewrote it to install a real-shaped TEMP capture trigger gated on `_sync_applying` (exactly like `installSyncCapture`), so it is now **discriminating**: verified it FAILS (`got 1` outbox row) when `withApplying` is neutered, PASSES when present.
2. **Defensive outer try/catch** on the `refreshRegistryMirror(client)` call site in `syncOnce` (symmetry with `convergeProjectsByRemote`/`reconcileScopeWraps`) — guards a throw from the loop header (tier resolution) from aborting the rest of the cycle.

155 core unit tests + 20 sync-engine Tier-2 tests green; typecheck/lint/format clean.

BYK:
Fixed the Seer finding (real bug): `refreshRegistryMirror` was passing raw PostgREST rows to `replaceRegistryTable`, storing ISO timestamptz **strings** in the local `created_at`/`updated_at` INTEGER columns. The keyset pull path normalizes these to epoch-ms via `stripSyncCols` (sync.ts:1570-1574) before `applyRemoteUpsert`; the snapshot path bypassed it. Now `refreshRegistryMirror` runs each fetched row through `stripSyncCols` (with `SCOPE_MEMBER_KEEP` for scope_members' PK `scope_id`), matching the keyset pull exactly. Added a Tier-2 assertion that `scopes.created_at`/`updated_at` land as epoch-ms integers. All green.

PR #1378, opened 2026-07-18
fix(core): FTS fallback for passive context-sources (#961 Gap 2b)

BYK:
## Problem

The passive context-source path (`loadContextSourceCandidates`, from #1228/#1293) that folds relevant **distillation/temporal facts into system[2]** was **vector-only** and fully gated behind a context embedding vector. Whenever embeddings are unavailable:

- the worker-init retry window,
- the pre-embed race (a distillation created before its async embed completes),
- or a non-embedding runtime,

`embedDistillation()` silently no-ops (`embedding.ts`: `if (!isAvailable()) return`) → `distillation_vec` stays empty → `vectorSearchDistillations` returns nothing every turn → **distillation facts never surface into system[2].**

The **knowledge path already degrades gracefully to FTS** in exactly this situation; context-sources did not. This is the root cause of #961 Gap 2b (facts stored but never surfaced in long/cross sessions when the embedding worker was down).

## Fix

Make `loadContextSourceCandidates` **hybrid**, mirroring the knowledge path. Alongside the vector pass it now runs a **BM25 keyword pass over the existing `distillation_fts` / `temporal_fts` indexes** (project-scoped), merging candidates by id and keeping the stronger normalized signal (max, per Seer #1318). FTS keyword hits are inherently relevant, so they bypass the cosine floor (identical to knowledge). `contextVec` is now **optional** and the call-site guard is relaxed from `&& contextVec` to `&& sessionContext.trim()`, so the block runs (FTS-only) when embeddings are down.

**No schema change** — `distillation_fts` already exists (db v7) and its sync triggers already populate it.

## Tests

- New **FTS-fallback test**: embeddings unavailable → vector helpers are never called, yet both distillation + temporal facts surface via keyword match. **Mutation-verified**: neutering the FTS match fails it (`expected false to be true`).
- New **vacuity test**: a keyword-disjoint context surfaces nothing (the FTS pass doesn't fold in every row).
- Retargeted the two vector-floor tests to a keyword-disjoint hint so they still isolate the cosine floor (they previously shared keywords with the fixtures and would now legitimately match via FTS).
- 3014 core tests green; tsc / lint / oxfmt clean.

## Live validation

A Sonnet session with embeddings disabled generated **22 distillations**; `distillation_fts` indexed all 22 (`distillation_vec` empty), and a BM25 query surfaced the **code-invisible probe fact** (`orders ride the WHOLESALE channel, EMEA region, warehouse WH-07`) as the top hit — nothing surfaced under the old vector-only path.

## Scope note

The embedding-worker init failure that exposed this in the eval is **eval-only infra** (the isolated eval gateway runs under Bun; `onnxruntime-node` throws "protobuf parsing failed" under a Bun worker). Production embeddings are healthy (`/health` reports `available:true, state:ok`). This PR is the durable product fix that makes context-sources resilient to **any** embedding outage regardless of runtime.

### DIFF
--- a/packages/core/src/ltm.ts
+++ b/packages/core/src/ltm.ts
@@ -2505,11 +2505,17 @@ export async function forSession(
   // fairly and inherit the identical stickyIds hysteresis + deterministic
   // packing below — the system[2] cache stays stable. Synthetic entries carry
   // the RECALLED_CONTEXT_CATEGORY tag and recall-id form ids.
-  if (options?.includeContextSources?.length && contextVec) {
+  //
+  // Runs whenever there is a session context to score against — NOT gated on
+  // contextVec. The FTS fallback inside surfaces facts even when embeddings are
+  // unavailable (worker init window, pre-embed race, non-embedding runtime); the
+  // vector pass adds semantic matches when a vector is present. (#961 Gap 2b)
+  if (options?.includeContextSources?.length && sessionContext.trim()) {
     const contextScored = await timer.await(
       loadContextSourceCandidates(
         pid,
         contextVec,
+        sessionContext,
         options.includeContextSources,
         options.contextSourceLimit ?? CONTEXT_SOURCE_LIMIT,
         sessionID,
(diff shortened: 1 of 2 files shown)


BYK:
Ran an adversarial review (SHIP-WITH-FOLLOWUPS — production code verified correct: cross-project scoping holds under mutation, score-merge sound, FTS syntax sanitized via `extractTopTerms`, limits bounded, cache-stability matches the knowledge path). Addressed the follow-ups in the same PR:

1. **Cross-project FTS leak test (the main gap).** All 9 prior tests passed even with the `project_id` filter dropped from either FTS query — the scoping invariant had zero direct coverage. Added `SECURITY: FTS fallback never surfaces distillation/temporal facts from ANOTHER project`: plants a foreign-project distillation + temporal message sharing keywords with the query, runs with embeddings down (FTS-only path), asserts PROJ's own facts surface but the foreign ones do not. **Mutation-verified**: dropping the `project_id` filter from the distillation FTS query OR the temporal FTS query fails it (the temporal case was previously uncovered).

2. **Defense-in-depth on temporal hydration.** Added `WHERE project_id = ?` to the temporal hydration query to mirror distillation hydration, so temporal isn't solely reliant on upstream FTS/vector scoping.

3. **Stale comment fixed.** The `vectorSearchDistillations is NOT project-scoped` comment was inaccurate for the vec0 backend (it partitions by project); reworded to note it's the legacy blob backend that needs the hydration re-filter.

Not changed (intentional): the per-turn FTS cost noted in review (point 4) is hybrid-by-design — the FTS pass adds keyword matches alongside the vector pass. It's bounded (`LIMIT limit*4` = 48 rows), offloaded to the read-worker pool, and consistent with the unconditional `knowledge_fts` scan `forSession` already runs. 10 tests green; tsc/lint/oxfmt clean.

PR #1380, opened 2026-07-18
fix(supabase): GitHub-provision role upgrade-on-resync + role validation (#1314 parts 2-4)

### LINKED ISSUE #1314: Team-org owner_user_id cascade + provisioning follow-ups (E-5-a)
Follow-ups from the #1313 (E-5-a-1) adversarial review. None blocking; the first is a real durability footgun to fix **before team sync is GA**.

1. **(durability — main)** `orgs.owner_user_id` is `NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE`. For a **team** org this is not an authz axis (all gates use `is_org_member`/`org_role`/`scope_role`), but the cascade means: if the recorded `owner_user_id` (the arbitrary first provisioner — possibly a plain GitHub *member* for a GitHub-mirrored org, or the creator for `create_team`) **deletes their account**, the entire team org → all scopes → all team content cascade-delete **for every other member**. Harmless today (no content lands in team scopes pre-GA). Fix before GA: for `kind='team'`, detach/transfer ownership (nullable + `ON DELETE SET NULL`, or reassign to another admin) instead of cascade. Applies to `create_team` (0029) as well as `provision_github_membership` (0035).

2. **(limitation)** `provision_github_membership` is strictly additive (never-demote AND never-upgrade). A sole team member first seen as `editor` who is later promoted to maintainer on GitHub stays `editor` → that team can remain permanently admin-less/unshareable. Tie off with a deliberate role-**upgrade**-on-resync path (upgrade-only, still never demote).

3. **(NIT)** `provision_github_membership` passes `rec->>'role'` through unvalidated (CHECK constraints are the backstop → whole-provision rollback on a bad role). Consider explicit per-record role validation for clearer errors (like `add_scope_member` 0029).

4. **(coverage)** The concurrent-race re-read branches in `provision_github_membership` (`0035:65-67, 88-90`) are unexercised by the sequential tests (mirrors the proven `provision_personal_scope` pattern). Add a two-transaction race/duplicate-key regression if feasible.

BYK:
Closes #1314 (parts 2-4; part 1 landed in #1374). Follow-ups to `provision_github_membership` (0035) from the E-5-a review. Migration 0048.

## Part 2 — role UPGRADE on re-sync (still never demote)
0035 used `ON CONFLICT DO NOTHING`, so a member promoted on GitHub (member→owner/admin, or team editor→maintainer) kept their **first-seen role forever** — a team could stay permanently admin-less/unshareable. Now a re-sync **ratchets the role UP** the GitHub-mappable ladder and never down:
- org `member(1) < manager(2)`, scope `viewer(1) < editor(2) < admin(3)`.

The manual-only, off-ladder org roles `owner`/`billing` are **never touched** (per the resolved design question): the UPGRADE is gated on the *existing* role being on the GitHub ladder AND the incoming role strictly out-ranking it. A manual `billing`/`owner` grant is preserved — the union-not-demote invariant holds. Two `IMMUTABLE` rank helpers (`org_role_rank`/`scope_role_rank`); EXECUTE revoked from all external roles (only the definer function calls them).

## Part 3 — explicit per-record role validation
0035 passed `rec->>'role'` straight through; a bad value only failed via the CHECK constraint, rolling back the **whole** provision with an opaque error. Now each record's role is validated up front (`errcode 22023`, clear message: org expects member/manager, team expects editor/admin).

## Part 4 — concurrent-race regression test
A new test provisions the **same** GitHub org/team from two users concurrently (`Promise.all`, separate service connections) and asserts convergence to exactly **one** org + **one** scope — exercising the lost-race re-read branches (0035:65-67 / 88-91) that the sequential tests never hit.

## Tests (Tier-1, real Postgres)
upgrade member→manager / editor→admin; never upgrade over a manual `billing`; invalid-role rejection (22023); concurrent-race convergence. Existing never-demote + idempotency + service-role-only tests still green (10/10 in the file). Lint/typecheck/format clean.

### DIFF
--- a/packages/gateway/test/sync-github-provision.integration.test.ts
+++ b/packages/gateway/test/sync-github-provision.integration.test.ts
@@ -1,5 +1,5 @@
 /**
- * Integration tests for migration 0035 — GitHub Teams provisioning (E-5-a, #827) against a real
+ * Integration tests for migration 0035 + 0048 — GitHub Teams provisioning (E-5-a, #827) against a real
  * Postgres. Proves the SERVICE-ROLE-ONLY provisioning RPC that mirrors a user's GitHub org/team
  * memberships into Lore orgs/scopes:
  *  - creates the org + scope keyed by GitHub numeric id, with the mapped roles;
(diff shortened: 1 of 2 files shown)


BYK:
Adversarial correctness + security review: **SHIP, no BLOCKERs** — never-demote (org+scope), off-ladder owner/billing protection, part-3 validation, migration ordering, definer/search_path, grant boundary, and byte-identical race branches all verified; 4/4 mutations killed, 10/10 integration tests green against real Postgres.

Notable finding from the review: **part-3 org role validation is a genuine privilege-escalation guard**, not just a nicety — `owner` is a CHECK-legal `org_members.role`, so there is **no CHECK backstop** at the org level. Without validation, a GitHub sync mapping a user to `owner` would *silently grant org ownership* (mutation M3 confirmed the insert succeeds when validation is removed). The validation closes that.

Applied the one cosmetic follow-up (stale `0035` labels/comments in the test file → now reference 0035+0048).

PR #1381, opened 2026-07-18
fix(supabase): revoke outstanding invite tokens on scope member removal (#1345, E-5-c F-2)

### LINKED ISSUE #1345: Revoke outstanding invite tokens on member removal (E-5-c F-2, #827)

BYK:
Closes #1345 (E-5-c F-2, follow-up from the E-5-c-1 adversarial review).

## Problem
`accept_scope_invite` tokens are single-use and expire in 14 days, but `remove_scope_member` did **not** revoke a scope's outstanding *unredeemed* invite tokens. A removed member holding a live (or leaked) token could **re-join** the scope as editor/viewer within the 14-day window.

## Fix — migration 0049
Invite tokens are **not invitee-bound** (any holder can redeem one), so there is no per-user token to revoke — the only sound fix is to revoke **all** of the scope's outstanding tokens on any member removal (an admin can re-issue). `remove_scope_member` (reproduced verbatim from 0029) gains one line:
```sql
delete from public.pending_invites where scope_id = p_scope;
```

## Severity
Low — membership grants **FETCH only**; DECRYPT needs a DEK wrap, and F-1 / migration 0043 (`scope_keys_target_member`) already blocks re-wrapping the rotated DEK to a non-member, so a rejoined-via-stale-token user cannot read post-rotation content. This closes the roster **re-join nuisance**. (`pending_invites` already `ON DELETE CASCADE`s on `scope_id`, so scope *deletion* cleared tokens; only member *removal* was uncovered.)

## Test (Tier-2, real Postgres)
Admin mints a live invite, removes a member → all the scope's tokens are revoked, the token can no longer be redeemed (`invalid or expired invite`), and the removed member is not a scope member.

Typecheck/lint/format clean. (Note: a pre-existing, environment-sensitive failure in the *"auto-wraps the DEK"* test in this file reproduces on unmodified main locally and is unrelated to this change; CI's fresh harness is the arbiter.)

### DIFF
--- a/packages/gateway/test/sync-team-invite.integration.test.ts
+++ b/packages/gateway/test/sync-team-invite.integration.test.ts
@@ -373,6 +373,44 @@ describe.skipIf(SKIP)("lore team — direct email invite (E-5-c)", () => {
     });
     expect(selfIns.error).toBeNull();
   });
+
+  it("revokes the scope's outstanding invite tokens on member removal (F-2, #1345)", async () => {
+    // A removed member (or anyone holding a live/leaked token) must not be able to re-join via a
+    // still-valid invite. remove_scope_member now wipes ALL of the scope's pending_invites.
+    const scope = await freshAdminTeam("Revoke-On-Remove");
+    await addTeamMember(clientFor(admin), scope, invitee, "editor");
+    // An admin mints an outstanding invite (for some future member) that is still live.
+    const token = await createTeamInvite(clientFor(admin), scope, "editor");
+    expect(
+      await h.client
+        .query(
+          "select count(*)::int n from public.pending_invites where scope_id=$1",
+          [scope],
+        )
+        .then((r) => r.rows[0].n),
+    ).toBe(1);
+    // Admin removes the member → the removal revokes ALL of the scope's outstanding tokens.
+    const { error: rmErr } = await clientFor(admin).rpc("remove_scope_member", {
+      p_scope: scope,
+      p_user: invitee,
+    });
+    expect(rmErr).toBeNull();
+    expect(
+      await h.client
+        .query(
+          "select count(*)::int n from public.pending_invites where scope_id=$1",
+          [scope],
+        )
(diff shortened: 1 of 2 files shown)


BYK:
Adversarial correctness + security review: **SHIP, no BLOCKERs**. Verbatim reproduction (only the added DELETE), correctly placed after the admin/last-admin guards, fully closes the re-join vector (including ephemeral/offline invites which also live in `pending_invites`), concurrency-safe via the shared per-scope advisory lock (remove/create/accept all serialize on the same key), idempotent, definer/search_path safe. Mutation-verified: dropping the DELETE fails the test (token survives removal). The `auto-wraps the DEK` flake in this file is pre-existing/unrelated (reproduces on unmodified base).

PR #1382, opened 2026-07-19
fix(sync): migrate an orphaned entity graph back to personal on team knowledge deletion (#1312, E-5-F3)

### LINKED ISSUE #1312: E-5-F3: demote/clean up a team knowledge entry's entity graph on deletion (remove())
Follow-up from the #1311 (#1307) adversarial review.

#1311 made team-promotion scope migration **symmetric** for the review flow: `approveForTeam` re-enqueues the linked entity graph (migrate → team) and `rejectForTeam` does too (migrate → personal). The remaining asymmetry is **`remove()`** (append-only death-cert deletion):

- When a team-scoped knowledge entry is deleted, its own delete/tombstone propagates, but its linked entities/aliases/relations/refs are **not** re-enqueued — so entity-graph metadata may linger in the team scope on the remote until some unrelated mutation re-enqueues those rows.
- Deeper question: after deletion the entry leaves `knowledge_current`, so `knowledgeTeamScope`/`entityTeamScope` resolve it to personal — the interaction between a deleted-knowledge tombstone push and its last-pushed team `scope_id` (F3-3 migration) needs verifying (does the delete target the correct — team — scope?).

Deliberately **not** bundled into #1311 because `remove()` is called in many non-team contexts (decay, eviction, curator deletes), so firing a graph re-enqueue there needs care (gating, cost) and its own tests. Bounded severity: only entity-graph metadata (names/types/aliases/relations) is affected — encrypted knowledge content is not exposed.

Suggested: on `remove()` of a team-scoped entry, re-enqueue the (now-orphaned-of-this-entry) linked graph so the push re-resolves each row's scope; add a test for delete-of-team-knowledge scope targeting. Also add: (a) a mixed-scope relation re-enqueue test, (b) a negative parent/remote-gate test (a linked entity in a remote-less project must not be enqueued).

BYK:
Closes #1312 (follow-up from the #1311 / #1307 review).

## Problem
#1311 made team-promotion scope migration symmetric for the **review** flow: `approveForTeam`/`rejectForTeam` re-enqueue the linked entity graph so the push re-resolves each row's scope. The remaining asymmetry was **`remove()`** (append-only death-cert deletion).

When a team-scoped knowledge entry is deleted, its own death-cert propagates and its `knowledge_entity_refs` are dropped (both ride capture triggers), but the linked **entities/aliases/relations** were **not** re-enqueued — so an entity that was team-scoped *only* because of the now-deleted entry lingered in the team scope on the remote until some unrelated mutation happened to re-enqueue it.

## Fix
`remove()` now fires a new `onKnowledgeDeleted` hook with the entity ids that just lost their last ref (already captured pre-delete for `recomputeEntityRank`). `sync-data` registers `reenqueueDeletedKnowledgeGraph`, which re-enqueues those entities + their aliases + relations so the push re-resolves each row's scope (`teamScopeForContent`) — migrating back to personal what no longer belongs to any team, while an entity still linked to **another** team-approved entry stays team-scoped.

**Cost / the many-call-sites concern:** gated on sync being enabled (no-op otherwise) and on `entitiesLosingRef` being non-empty, so the non-team `remove()` call sites (decay, eviction, curator) pay nothing when sync is off and only a cheap, idempotent re-enqueue when on. The hook (not a direct import) mirrors the existing `onKnowledgeTeamPromotionChanged` pattern (sync-data already imports ltm → reverse would be a cycle).

**Note:** a deleted team entry's **own** death-cert stays team-scoped — a tombstone is not a scope migration; team peers must learn of the deletion. Only the orphaned entity graph migrates.

## Test (full engine)
A team entry K links `e1` (only K) + `e2` (K **and** a second approved K2); after `remove(K)`, `e1` + its alias migrate back to personal while `e2` stays team. Mutation-verified: without the re-enqueue, `e1` stays team and the test fails. Typecheck/lint/format clean; 213 tests across the affected suites green.

### DIFF
--- a/packages/core/src/ltm.ts
+++ b/packages/core/src/ltm.ts
@@ -654,6 +654,20 @@ export function onKnowledgeTeamPromotionChanged(
   teamPromotionChangedCb = cb;
 }
 
+// E-5-F3 (#1312): notified with the entity ids that LOST their last ref to a knowledge entry when
+// that entry is deleted (remove()). The entity graph's team-scope is derived from the APPROVED
+// knowledge that references each entity (entityTeamScope); once the referencing team-knowledge is
+// tombstoned + its refs dropped, a formerly team-scoped entity may no longer belong to any team and
+// must migrate BACK to personal. remove() deletes the refs (no trigger re-resolves the entity), so
+// the entities/aliases/relations are re-enqueued explicitly for the push to re-resolve their scope.
+// A hook (not a direct import) because sync-data.ts already imports ltm.ts (reverse would be a cycle).
+let knowledgeDeletedCb: ((lostEntityIds: string[]) => void) | null = null;
+export function onKnowledgeDeleted(
+  cb: (lostEntityIds: string[]) => void,
+): void {
+  knowledgeDeletedCb = cb;
+}
+
 /**
  * Approve a knowledge entry for team promotion (manual review). Sets 'approved' + reviewer/time on
  * the current version, in place. Idempotent; returns false if no current row matched the id.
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial correctness review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — verified across all 10 points (pre-delete `entitiesLosingRef` capture, resolver re-resolution after death-cert + ref delete, gate/binding parity with `reenqueueKnowledgeTeamGraph`, no import cycle, sync-OFF true no-op, correct relation both-endpoints gating). Confirmed the knowledge tombstone correctly **stays team-scoped** (a tombstone is not a scope migration — peers must learn of the deletion); only the orphaned entity graph migrates.

Addressed the one SHOULD-FIX the review found: the test had **no `entity_relations` row**, so the delete-path relation re-enqueue clause was untested (dropping it left all tests green). Added a relation `r1` between the orphaning entity (e1) and the staying entity (e2) and asserted it migrates to personal — mutation-verified the clause is now load-bearing (dropping it now fails the test). Also added a comment noting the entity enqueue is trigger-redundant and the load-bearing part is aliases/relations.

PR #1385, opened 2026-07-19
fix(core): drain in-flight distillation embeds on shutdown + warm-start worker (#1331)

### LINKED ISSUE #3: publish: BYK/opencode-lore@main
Requested by: @BYK

Quick links:
- [View changes](https://github.com/BYK/opencode-lore/compare/0.3.0...release/main)

Assign the **accepted** label to this issue to approve the release.

### Targets

 - [ ] npm
 - [ ] github

Checked targets will be skipped.

### LINKED ISSUE #1331: Fast/short sessions tear down before fire-and-forget distillation embeds finish → silent recall degradation
## Symptom
On short or fast sessions, distillations are created but never embedded (`distillation_vec` stays empty), so recall silently degrades to FTS/temporal-only. No error is logged — the only signal is a `worker not running` warning on some sessions. Observed repeatedly in the eval harness (#961): single-session Lore runs scored 0/4 on fact retention despite distillations existing; the same task scored 4/4 once a settle wait was added before the probe.

## Root cause
`embedDistillation()` (packages/core/src/embedding.ts:2168) is fire-and-forget: `trackDocEmbed(embed(...).then(store))`. The **first** `embed()` call pays a ~21s cold model load (measured). If the session ends (gateway torn down) — or reaches the point where recall is needed — before that promise resolves, the vector is never written. Distillations exist in the base table but have no `distillation_vec` row, so vector recall returns nothing.

This is invisible because:
- `isAvailable()` returns true (provider constructed optimistically; init is lazy in the worker).
- The embed rejects/aborts on teardown without an error path that surfaces (`LocalProviderUnavailableError` is swallowed).
- Startup log shows `coverage: distillations 0/0` and there is no 'embedding scheduled' line for in-session distillations.

## Impact
Any real user with short sessions (or a machine where the model cold-loads slowly) gets degraded recall on recently-distilled facts until a later session's backfill catches up. The facts ARE stored; they're just not retrievable via vector search when needed.

## Possible fixes
1. **Drain-on-shutdown**: on graceful gateway shutdown, await outstanding `trackDocEmbed` promises (bounded) so in-flight embeds complete before exit.
2. **Warm-start**: pre-warm the embedding worker at gateway startup (throwaway embed) so the cold-load is paid before the first real distillation, keeping incremental embeds fast enough to finish within a turn.
3. **Startup backfill already re-indexes** missing `distillation_vec` rows on next boot — but that only helps the *next* session, not the one that created them.

## Repro
Single-session task, cheap fast model, buried fact in turn 1, probe in a later turn of the same session. Without a settle wait, `distillation_vec` is empty at probe time and recall misses the fact. Found via #961 eval harness (driver added an in-session `waitForMemoryReady` before the probe turn as a harness-side workaround; the product-side fix is tracked here).

cc @onurtemizkan

BYK:
Fixes #1331.

## Problem

On short/fast sessions, distillations are created but never embedded
(`distillation_vec` stays empty), so recall silently degrades to FTS/temporal-only.
`embedDistillation()` is fire-and-forget and the **first** `embed()` pays a ~21s
ONNX cold-load. If the session ends (gateway torn down) — or reaches the point
where recall is needed — before that promise resolves, the vector is never
written.

## Fixes

Fix #3 from the issue (next-boot re-index of missing `distillation_vec`) was
already implemented in `runStartupBackfill`. This PR adds the two remaining
product-side fixes; they are complementary — warm-start makes the common case
finish within the turn, the bounded drain is the safety net.

### 1. Bounded drain-on-shutdown
- `settleDocumentEmbeds()` gains an optional `timeoutMs`. No arg → unbounded
  (existing test behavior); with a bound → races the in-flight set against a
  deadline and resolves either way.
- Wired into the production `shutdown` closure in `start.ts`, **between**
  `resetPipelineState({ fast: true })` and `embedding.resetProvider()` (must run
  while the worker is still alive). Bounded by
  `EMBED_DRAIN_DEADLINE_MS = 0.6 x SHUTDOWN_DEADLINE_MS` so it can't reintroduce
  the Ctrl+C hang that `fast: true` exists to avoid. Anything left pending is
  re-indexed on next boot.

### 2. Warm-start
- New `warmupEmbedding()`: fires one throwaway local `embed(["warmup"], "document")`
  at gateway startup to pay the ~21s cold-load up front.
- Local-only (remote providers have no cold-load and a warmup would burn API
  quota); `document` inputType so it is never counted as a recall embed and
  cannot self-gate the temporal backfill; fire-and-forget (never blocks startup).
- Wired into pipeline startup beside `runStartupBackfill` (same
  `NODE_ENV !== "test"` gate), so the model loads during the backfill delay.

## Tests
- `background-drain.test.ts`: bounded drain resolves at the deadline with a
  pending embed / returns fast on completion; warmup fires exactly one `document`
  embed locally; warmup does not touch the recall-in-flight counter; warmup is a
  no-op for remote providers (verified via `fetch` spy — no network call).
- `shutdown-embed-drain.test.ts` (new): asserts the shutdown closure calls the
  bounded drain **before** `resetProvider`, with a finite positive deadline.

### Non-vacuousness (mutation-verified)
- remove drain call → shutdown-order test fails
- drain after reset → shutdown-order test fails
- remove warmup body → local-fires test fails
- `document` → `query` inputType → recall-counter test fails
- remove local-only guard → remote fetch-spy test fails
- bounded drain ignores timeout (unbounded) → deadline test hangs (killed)

## Verification
- `pnpm run typecheck`, `lint`, `format` clean.
- Full suite: 6124 passed, 0 failed.

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -310,9 +310,6 @@
 <!-- lore:019f147e-7b0d-7ed0-aef5-bb2b8c0bc065 -->
 * **standard.site publish script: safety invariants for orphan cleanup**: Three invariants in \`publish-standard-site.mjs\` orphan cleanup (PR #1043, also ported to byk.github.io PR #58): 1. \*\*Upserts-first ordering\*\*: all \`putRecord\` calls complete before any \`deleteRecord\` — a mid-run failure never strands the blog without live records. 2. \*\*Empty-manifest guard\*\*: if \`manifest.documents.length === 0\`, skip orphan cleanup entirely with \`console.warn\` — a broken/empty build cannot wipe the whole collection. 3. \*\*Pagination termination\*\*: \`listAllRecords\` uses dual-condition stop — (a) stop when batch length is 0 even if PDS returns a cursor (PDS returns trailing cursor on last page — confirmed live), (b) stop when cursor is null/undefined. 4. \*\*DID cross-check\*\*: if \`session.did !== recordDid\`, script refuses to publish. 5. \*\*\`validate: false\`\*\* on \`putRecord\` — PDS doesn't host \`site.standard.\*\` lexicons. 6. \*\*Blob uploads\*\*: blobs are content-addressed (CID) → idempotent re-runs; best-effort; size cap \`MAX\_BLOB\_BYTES = 1\_000\_000\`; Astro-optimized webp variants always under 1MB.
 
-<!-- lore:019f1cc8-b694-721d-9b2b-10ff1ed2d390 -->
-* **storeTurnTemporal: user-side store must run BEFORE resolveToolResults (ordering constraint)**: Pattern: storeTurnTemporal: user-side store must run BEFORE resolveToolResults (ordering constraint).
-
 <!-- lore:019ef387-4afe-7137-a151-88dff1dd0ab7 -->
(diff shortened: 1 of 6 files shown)


BYK:
## Adversarial review — resolution

Ran an adversarial correctness review (verdict: SHIP-WITH-FOLLOWUPS, no BLOCKER). Addressed the findings in `5663c93d`:

- **SHOULD-FIX #1 (comment overstates the drain's guarantee):** The shutdown drain only covers document embeds *already dispatched* — a distillation whose LLM call is still in flight hasn't called `embedDistillation()` yet, so its embed isn't enqueued. Fast-path shutdown deliberately skips the background drain (that's the whole point of `fast: true` — avoiding the Ctrl+C hang), so I did **not** add a background drain into the fast path. Instead the comment now states the scope honestly; that sub-case is covered by `runStartupBackfill` on next boot (plus warm-start makes it rare).
- **NIT #2 (sub-ms busy-spin at the deadline boundary):** the bounded drain now breaks the instant the deadline wins the race (sentinel value) instead of re-polling `Date.now()`.

Findings #3 (archived gen-0 excluded from next-boot backfill) and #4 (`.lore.md` churn) are pre-existing / by-convention and out of scope here.

Re-verified after the fix: targeted suites green (9 passed), mutation on the new `break` still kills the deadline test (loop hangs → test times out), typecheck/lint/format clean.

PR #1386, opened 2026-07-19
fix(gateway): surface reasoning-only worker completions from the OpenAI SSE path (#1334)

### LINKED ISSUE #1334: MiniMax-M3 via OpenRouter (openai protocol) returns empty worker completions → distillation/curation parse-error, 0 knowledge extracted
## Symptom
When MiniMax-M3 is used as the Lore **worker model** via OpenRouter (`openrouter/minimax/minimax-m3`, openai protocol), background distillation and curation silently produce **0 knowledge entries**. Logs show:
```
WARN: [worker-health] lore-distill failed (parse-error) for session=… — N in window
/lore:curate: Curation complete: 11 segments distilled, 0 entries created
```
Cross-session recall then returns nothing (Lore scores 0/N) because no knowledge was ever extracted.

## Root cause (suspected)
`parseDistillationResult` (distillation.ts) is lenient — it returns null only when the completion is **empty** after trim. So `parse-error` here means the worker returned an **empty completion**. M3-via-OpenRouter (openai protocol) appears to return empty `content` for the distill/curate worker prompt — possibly all output goes to a reasoning/thinking field, or the openai-protocol worker path doesn't extract M3's content shape correctly.

Notably, the SAME model on the MiniMax **coding-plan** endpoint (anthropic protocol) previously produced knowledge fine (earlier #961 eval showed ~83% cross-session retention). The failure is specific to the **openai-protocol / OpenRouter** worker path.

## Contrast
- **DeepSeek V4 Flash** as worker (also openai protocol, via OpenRouter): curation works, 8 knowledge entries extracted, cross-session 4/4.
- **M3 via OpenRouter**: 0 entries, parse-error.

So it's not openai-protocol in general — it's M3's response shape on that path.

## Impact
Any user running M3 via an openai-protocol route as their worker model gets silently degraded memory (distillations created, but no knowledge, and possibly no embeddings depending on timing). No hard error surfaces beyond the worker-health parse-error warning.

## Repro
`openrouter/minimax/minimax-m3` as answering+worker model, run a cross-session task, force `/lore:curate`, observe `0 entries created` + `parse-error` warnings. Found via #961 eval harness.

## Questions
- Is M3 returning content in a non-standard field on the openai path that the worker LLM adapter should map?
- Should the distiller log the raw (empty) completion at debug to confirm empty-vs-malformed?

cc @onurtemizkan

BYK:
Closes #1334.

## Symptom
MiniMax-M3 via OpenRouter (`openrouter/minimax/minimax-m3`, openai protocol) as the Lore **worker model** produced **0 knowledge entries** — background distillation/curation logged `parse-error` ("empty completion") and cross-session recall returned nothing. DeepSeek V4 Flash (also openai/OpenRouter) worked fine.

## Root cause
An asymmetry between the **non-streaming** and **streaming** worker parse paths. Workers send `stream:false`, but OpenRouter (and DeepSeek etc.) return an SSE body anyway, so the worker read routes through `accumulateOpenAISSEStream` → `gatewayResponseToWorkerResult`. That streaming path **only accumulated `delta.content`** and **only projected `type:"text"` blocks** — it dropped reasoning deltas entirely. M3 streams its whole answer as reasoning deltas and leaves `content` empty → the worker saw an empty completion.

The non-streaming `parseOpenAIResponse` already fell back to `reasoning_content`/`reasoning`; the streaming path did not. DeepSeek worked because it emits real `content`.

## Fix (mirrors the non-streaming fallback)
1. **`stream/openai.ts`** — accumulate `delta.reasoning` (OpenRouter) / `delta.reasoning_content` (DeepSeek/Qwen) into a `thinking` block (emitted before text, Anthropic ordering). Reasoning was previously dropped on this path entirely, so this is purely additive.
2. **`gatewayResponseToWorkerResult`** — fall back to `thinking` blocks when there is no visible text, mirroring `parseOpenAIResponse` (content→reasoning) and `parseAnthropicResponse` (text→thinking). A `tool_use`-only response stays empty (workers consume text, not tool calls).

This also makes a reasoning-only response correctly non-empty on the **conversation** path (the pipeline already handles `thinking` blocks), aligning the OpenAI stream path with the Anthropic path.

## Tests
- `stream-openai.test.ts`: `reasoning`/`reasoning_content` deltas → thinking block; thinking-before-text ordering; no-reasoning path unchanged.
- `worker-reasoning-parse.test.ts`: `gatewayResponseToWorkerResult` thinking fallback, prefer-text, multi-block join, `tool_use`-only stays null.
- Mutation-verified: neutering either half fails the tests. 201 tests across streaming/worker/pipeline suites green; typecheck/lint/format clean.

### DIFF
--- a/packages/gateway/src/llm-adapter.ts
+++ b/packages/gateway/src/llm-adapter.ts
@@ -1385,6 +1385,19 @@ export function gatewayResponseToWorkerResult(resp: GatewayResponse): {
     )
     .map((b) => b.text)
     .join("");
+  // Reasoning models (e.g. MiniMax-M3 via OpenRouter) can emit their entire answer
+  // as reasoning/thinking with an empty text block. Never treat a present thinking
+  // body as no-response — fall back to it when there is no visible text, mirroring
+  // parseOpenAIResponse (content→reasoning) and parseAnthropicResponse (text→thinking). (#1334)
+  const workerText =
+    text ||
+    resp.content
+      .filter(
+        (b): b is Extract<GatewayContentBlock, { type: "thinking" }> =>
+          b.type === "thinking",
+      )
+      .map((b) => b.thinking)
+      .join("");
   const usage: AnthropicUsage | null = resp.usage
     ? {
         input_tokens: resp.usage.inputTokens,
@@ -1393,7 +1406,7 @@ export function gatewayResponseToWorkerResult(resp: GatewayResponse): {
         cache_creation_input_tokens: resp.usage.cacheCreationInputTokens,
       }
     : null;
-  return { text: text || null, usage, model: resp.model || null };
+  return { text: workerText || null, usage, model: resp.model || null };
 }
 
 /**
--- a/packages/gateway/src/stream/openai.ts
+++ b/packages/gateway/src/stream/openai.ts
@@ -369,6 +369,11 @@ export async function accumulateOpenAISSEStream(
   let model = "";
   let stopReason = "end_turn";
   let textContent = "";
+  // Some reasoning models (e.g. MiniMax-M3 via OpenRouter) stream their entire
+  // answer as reasoning deltas and leave `content` empty. Capture it so a
+  // reasoning-only response is not mistaken for an empty completion (#1334) —
+  // mirrors the non-streaming parseOpenAIResponse content→reasoning fallback.
+  let reasoningContent = "";
   const toolCalls = new Map<
     number,
     { id: string; name: string; args: string }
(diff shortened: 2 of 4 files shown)


BYK:
Adversarial correctness review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs — verified all 8 points including the critical conversation-path safety (OpenAI client egress silently drops the thinking block → zero client-visible change; Anthropic egress surfaces it well-formed with signature correctly omitted; the `isEmptyCompletion` de-noise is the intended fix, and no retry loop is wired to it). Both mutations killed; 155+ tests green.

Addressed the reviewer's follow-ups:
1. **Added the end-to-end seam test** the reviewer flagged as missing — drives a reasoning-only (mislabeled `application/json`) SSE body through the real worker wire `isSSE → looksLikeSSE → accumulateWorkerSSE(openai) → gatewayResponseToWorkerResult → client.prompt` and asserts non-null text. Mutation-verified: pre-fix it returns `null` (the exact #1334 production failure).
2. **NIT** — added a comment noting the streaming `reasoning`-vs-`reasoning_content` precedence is per-provider identical to `parseOpenAIResponse` (diverges only in the impossible both-set-in-one-delta case).
3. The whitespace-only truthiness quirk is pre-existing (not introduced here) and left as-is.

Issue #630, opened 2026-06-09
feat: GitHub App + Folk Lore team integration (GitHub Teams, repo collaborators, network effects)


BYK:
## Summary

Build a GitHub App as the primary integration surface for Folk Lore, leveraging GitHub Teams, repo collaborators, and org structure to provide team memory management that surpasses Copilot Memory's enterprise controls — while creating network effects among OSS and individual developers.

## Motivation

GitHub Copilot Memory offers basic enterprise controls (view/delete per user or bulk, billing-entity scoping). Folk Lore can leapfrog this by importing the team structure developers already maintain on GitHub, rather than building a parallel management system.

Additionally, by importing **repo collaborator/access configuration**, we enable a lightweight "add this repo to Lore" flow that creates network effects: when one collaborator on an OSS project uses Lore, all collaborators see the `.lore.md` in the repo — and a one-click GitHub App install brings them into the Folk Lore ecosystem.

### GitHub Teams Integration

| Capability | How it works |
|---|---|
| Team → knowledge scope | "Backend Team" gets backend architectural decisions; "Frontend Team" gets frontend patterns. Knowledge visibility maps to existing team boundaries. |
| Org admin → Folk Lore admin | Natural role mapping from GitHub org roles to Folk Lore admin permissions. |
| Team membership changes → knowledge access | Automatic sync — no manual management of who sees what. |
| Cross-team knowledge promotion | Entries auto-promoted across teams when semantically relevant to 3+ team-scoped repos. Approval workflows gate cross-team visibility. |

### Repo Collaborators / Access Configuration

| Capability | How it works |
|---|---|
| "Add this repo to Lore" | GitHub App reads repo collaborators. One collaborator installs → all collaborators can be onboarded with one click. |
| OSS network effect | When a contributor to an OSS project installs Lore, the `.lore.md` is already in the repo. The GitHub App surfaces this: "This repo has Lore knowledge — install to use it." Other contributors see the value and join. |
| Individual developer onboarding | GitHub Login → App reads your repos + collaborator lists → auto-discovers which repos already have `.lore.md` → instant setup for all of them. |
| Access control inheritance | Repo read access → can consume knowledge. Repo write access → can contribute knowledge. Repo admin → can manage knowledge settings. |

### Comparison with Copilot Memory Enterprise

| Capability | Copilot Memory | Folk Lore + GitHub App |
|---|---|---|
[...]

PR #1387, opened 2026-07-19
fix(core): fall back to WASM ONNX runtime when native can't load the model (#1379)

### LINKED ISSUE #1379: Eval iso-gateway: local ONNX embeddings fail to init under Bun (protobuf parsing failed)
## Summary

The eval harness's isolated Lore gateway runs under **Bun** (`spawn("bun", [launcher])`, `dist/index.bun.js`). Under Bun, the local ONNX embedding worker fails to initialize with:

```
ERROR_CODE: 7, ERROR_MESSAGE: Failed to load model because protobuf parsing failed
```

As a result `embedding.isAvailable()` returns `false` for the entire run, and:
- `embedDistillation()` silently no-ops (`embedding.ts`: `if (!isAvailable()) return`)
- `distillation_vec` / `knowledge_vec` stay empty
- vector recall is fully degraded in eval arms

## Not a production bug

The **production gateway is healthy** — it runs the embedding worker under **Node**, and `/health` reports:

```json
{ "embeddings": { "available": true, "state": "ok", "provider": "local", "detail": "local ONNX embeddings active (vector recall enabled)" } }
```

The model file hash in the eval workspace is **identical** to the working production copy, so this is a **Bun + onnxruntime-node runtime incompatibility**, not a corrupted model.

## Impact

Eval-only. It previously masked #961 Gap 2b (distillation facts never surfaced into system[2] because they were never embedded). Gap 2b is now fixed durably by the FTS fallback (PR #1378), so context-source recall works in the eval regardless — but the eval's **vector** path is still dark under Bun, which:
- understates Lore's semantic-recall quality in eval measurements, and
- means the eval never exercises the vec0 storage path end-to-end.

## Options to investigate

1. Run the eval iso-gateway under **Node** instead of Bun (the production runtime) — most faithful; needs the CJS worker bundle + node launcher path in `driver.mjs`.
2. Root-cause the Bun + `onnxruntime-node` "protobuf parsing failed" (Bun worker-thread native-addon loading, ORT version, or WASM-vs-native ORT selection under Bun).
3. Fall back to WASM ORT under Bun for the eval (`ort-wasm-simd-threaded`), accepting slower embeds.

## Repro

Launch any eval task with the Bun iso-gateway and grep `gateway.log` for `protobuf parsing failed`; confirm `distillation_vec` count is 0 despite N distillations, and startup log shows `coverage: distillations 0/0` even after distillations are created.

BYK:
Fixes #1379.

## Problem

The eval iso-gateway runs the npm bundle under **Bun** (`dist/index.bun.js`). The embedding worker's `loadPipeline()` prefers native ONNX: `resolveNativeOrtBindingPath()` uses `createRequire().resolve()`, which **works under Bun** and finds the installed `@loreai/onnxruntime-<target>` addon → native path is selected. But the native `onnxruntime-node` NAPI addon then **fails to load the model under Bun** with `Failed to load model because protobuf parsing failed` (a Bun ↔ onnxruntime-node worker-thread native-addon incompatibility).

Because native *resolution* succeeds, the already-Bun-hardened WASM fallback is never reached. Worse, the failure matches `isCorruptModelError`, so the worker's corrupt-model self-heal would (wrongly) treat an intact model as corrupt. Result: `isAvailable()` is false for the whole run, `embedDistillation()` no-ops, `distillation_vec`/`knowledge_vec` stay empty, and vector recall is dark. **Production (Node) is unaffected.**

## Fix — probe native, fall back to WASM keyed on the actual error

The backend choice is committed at the first `require("onnxruntime-node")` (the bundle's runtime shim reads `globalThis.__LORE_ORT_BINDING_PATH__` once and caches the module), so an in-process retry cannot switch to WASM. The fix mirrors the existing OOM→respawn pattern:

1. **Worker** — when pipeline init fails with an `isCorruptModelError` parse failure **while native was in use** and the on-disk model **looks structurally intact** (`looksLikeIntactOnnxFile`), it posts a new `init-needs-wasm` message instead of purging (the model isn't corrupt) or retrying in-process (same backend).
2. **Main thread** (`LocalProvider`) — on `init-needs-wasm`, terminates the native worker and respawns a **fresh** worker with `WorkerInitData.forceWasm=true` (a new module graph that skips native → uses the bundled, Bun-hardened WASM runtime), re-submitting any in-flight request. `forceWasm` is **sticky** across later respawns (incl. OOM backoff) so it never bounces back to native.
3. **One-shot latch** — a second `init-needs-wasm` (WASM also failing, i.e. genuine corruption on both backends) falls through to the normal `init-error` path; no respawn loop.

This is runtime-agnostic (keyed on the real signal, not a Bun sniff), so it also self-heals any future native-incompat while keeping Node/production and the SEA binary on fast native.

## Tests (all non-vacuous, mutation-verified)

- `embedding-wasm-fallback.test.ts` — respawn-forcing-WASM + resubmit; `forceWasm` sticky across a later OOM respawn; second `init-needs-wasm` latches (no loop).
- `embedding-worker-types.test.ts` — new pure predicate `shouldRequestWasmRespawn` (native/WASM, vendored, intact/truncated, error-kind).

Mutations killed: don't-set-forceWasm (2 tests), remove-one-shot-latch (1 test), drop-usedNative-clause in predicate (1 test).

Full suite: 6143 passed / 0 failed. typecheck + lint + format clean.

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -90,9 +90,6 @@
 <!-- lore:019f3974-7c6a-7091-8602-b4ec0bb763a4 -->
 * **Chose next task: knowledge\_meta seed-order alignment (self-contained, low-risk, completes value-ranked seed coherence work started with knowledge #1189 and entities #1194) — over #1191b (continuous top-N eviction)**: Completed value-ranked seed coherence work in sequence: #1189 (knowledge seed value-ranking) → PR #1194 (entity-graph seed: entities ranked by knowledge\_entity\_refs count DESC then recency; entity\_relations by recency; entity\_aliases by created\_at DESC) → PR #1196 (knowledge\_meta seed aligned to knowledge's ORDER BY: confidence DESC, COALESCE(last\_reinforced\_at, updated\_at) DESC, logical\_id tiebreak). Deferred #1191b (continuous top-N eviction for remote) as a separate larger/riskier task needing a design doc first (remote eviction, cap discovery, thrash guards, concurrency) — not suitable for autonomous same-session implementation. Follow-up issue #1197 filed for a known non-blocking edge case: a deleted high-confidence entry's orphaned knowledge\_meta row can outrank a live entry's meta row under the cap (self-healing, not data loss).
 
-<!-- lore:019f5211-7cb6-7f6e-b1ee-cc442b8878ad -->
-* **Chose to proceed with team lifecycle epic over verifying main's release state**: chose to proceed with team lifecycle epic over verifying main's release state.
-
 <!-- lore:019f3d0e-5602-7457-b4d3-a36993bd7b9b -->
 * **Chose to run real A/B eval scenario cm-1-early-detail (context dimension, real recall questions with cumulativeTokens for rot curve) over cost-2**: chose to run real A/B eval scenario cm-1-early-detail (context dimension, real recall questions with cumulativeTokens for rot curve) over cost-2
 
@@ -105,9 +102,6 @@
 <!-- lore:019efe86-ad54-7ad3-868b-c3a76af9d306 -->
 * **Lore license: FSL-1.1-Apache-2.0 (Fair Source), NOT open source**: Lore/loreai is Fair Source, specifically FSL-1.1-Apache-2.0 (Functional Source License v1.1, Apache 2.0 Future License). Competing Use (using code to compete against Lore) is prohibited. Converts to Apache 2.0 later. Copyright 2026 Burak Yigit Kaya. Implication: competitors legally cannot lift Lore's context-management code. This kills any 'more open than ByteRover's Elastic License' positioning angle — both are source-available, not OSS. Licensing framing belongs on positioning pages, not in vendor-neutral blog posts.
 
-<!-- lore:019ef58f-9ff2-7edc-a84c-e7b3ba2622a4 -->
-* **Migrated to register (no assertion values changed)**: migrated to register (no assertion values changed).
-
 <!-- lore:019f5b56-d717-7acf-82cd-ecfc75ca4871 -->
 * **Migrated to TG): test failed at sync**: migrating to TG): test failed at sync.
 
@@ -119,12 +113,6 @@
 
 <!-- lore:019efa1b-96f8-7702-8ddb-ccbc2fece37e -->
 * **sqlite-vec: native extension chosen over WASM; vec0 virtual table deferred**: Chose native extension + JS fallback over WASM-first for sqlite-vec. WASM build targets browser sqlite-wasm engine only — using it in node:sqlite/bun:sqlite would require replacing the entire SQLite engine, slowing ALL queries. Drop-in \`vec\_distance\_cosine()\` over existing BLOB columns chosen for PR #967 scope; \`vec0\` virtual table migration rejected as first step. Portability strategy: keep embeddings as F32 BLOB columns (byte-compatible with libSQL's \`F32\_BLOB\`); when hosted gateway becomes real, add \`driver.libsql.ts\` behind existing \`#db/driver\` map backed by \`vector\_top\_k\` — no data migration, no caller changes.
-
-<!-- lore:019f718f-9d3f-7637-bd1e-f56a97912261 -->
-* **Switched from claude-sonnet-4.6 to claude-sonnet-5 for consistency with published results**: Switched from claude-sonnet-4.6 to claude-sonnet-5 for consistency with published results.
-
-<!-- lore:019f05e3-1abc-7c77-944b-bc9e7ae25bb1 -->
-* **Switched from plan mode to build mode (read-write**: switched from plan to build mode (read-only lifted; file changes,
 
 ### Gotcha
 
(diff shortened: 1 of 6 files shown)


BYK:
## Adversarial review round 1 — resolution

First adversarial review returned **NEEDS-WORK** with two deterministic BLOCKERs on the happy path (both invisible to the original tests because the fake worker didn't model real `terminate()→exit` semantics or the worker's two-message output). Fixed in `ca143009`:

- **B1 — stale native worker's `exit(1)` latched the provider broken.** `terminate()` emits an async `exit(1)`; the unconditional `on("exit")`/`on("error")`/`on("message")` handlers then nulled the fresh worker, rejected the resubmitted request, and set `localProviderKnownBroken` → `isAvailable()` false (the #1379 symptom). **Fix:** each handler is now bound to the specific `spawned` worker and early-returns when `this.worker !== spawned`, so a superseded worker's late events are ignored.
- **B2 — trailing per-request error dropped the in-flight embed.** After posting `init-needs-wasm`, the worker's `processEmbed` catch posted `{type:"error", id}` (since `initFailed` was false), which the main thread used to reject+delete the pending before `respawnForWasm` could resubmit. **Fix:** the worker suppresses the per-request error when `wasmRespawnRequested` is set.
- **S1 — test fidelity.** `FakeWorker.terminate()` now emits `exit(1)` (real semantics), and a new protocol-faithful test drives the full sequence (`init-needs-wasm` + trailing `error` + terminate `exit`) and asserts the caller still resolves and `isAvailable()` stays true. Reverting either B1 or the guards fails this test.
- **S2 — dead predicate / drift risk.** Added a source-parse drift guard asserting the worker's inline `isCorruptModelError` (the load-bearing component of the respawn decision) matches the canonical, mirroring the existing `isTransformersInferenceDumpLine` guard. Mutation-verified.
- **N1** — added a comment at the `terminate()` call documenting that it emits `exit(1)` and why the per-worker guard makes that safe.

Re-verified: full suite **6135 passed / 0 failed**; typecheck + lint + format clean. Mutation checks confirm B1 fix (kills 3 tests when reverted), the drift guard (kills on regex change), and the original guards all non-vacuous.

Running a second adversarial review to confirm the fixes hold before merge.

BYK:
## Scope cleanup + review round 2

**Second adversarial review: SHIP-WITH-FOLLOWUPS.** It confirmed the #1379 fix is correct — B1 (per-worker event guard) is mutation-verified and all `respawnForWasm`/`terminate()`/`ensureWorker` race windows are handled (no hang, double-spawn, false latch, or dropped legitimate event); B2 is correct defense-in-depth.

It also caught that the pushed commit had picked up **unrelated files from concurrent workstreams** sharing this working copy (earlier: #1334 reasoning-stream files; then the #630 github-discover feature + a remote Supabase migration `0050_github_discover_lookup.sql`). A stray remote migration in an ONNX-fallback PR is a real deploy hazard. **Fixed:** rebuilt the branch off the latest `main` with ONLY the six #1379 files — `git diff --name-only origin/main <head>` is now exactly:
- `packages/core/src/embedding.ts`, `embedding-worker.ts`, `embedding-worker-types.ts`
- `packages/core/test/embedding-wasm-fallback.test.ts`, `embedding-worker-types.test.ts`
- `.lore.md`

### Follow-ups (non-blocking, from review #2)
- **B2 independent coverage**: reverting the worker-side `!wasmRespawnRequested` guard doesn't fail a test because the fake-worker harness never runs the real `embedding-worker.ts`, and B1's main-thread guard already drops the trailing error. B2 is correct/safe (redundant with B1); a direct worker-level unit test would close the gap.
- **`shouldRequestWasmRespawn`** is exercised by unit tests but the worker inlines the equivalent nested `if`s; the load-bearing `isCorruptModelError` component is drift-guarded.

BYK:
## Seer review triage (3 findings)

- **[Resolved] LOW — exit handler overrides transient retry latch.** Seer already marked this resolved in `a24b756`: the per-worker `if (this.worker !== spawned) return` guard on the exit handler prevents a superseded/terminated worker's `exit` from latching `localProviderKnownBroken`.
- **[Applied] HIGH — `workerReady` race in WASM fallback.** Real-world it's mitigated (the handler sets `workerReady=false` synchronously and there's no `await` before `respawnForWasm`→`ensureWorker` reads it), but I applied Seer's suggestion anyway as defense-in-depth: `respawnForWasm` now sets `this.workerReady = false` locally right before `ensureWorker()`, so the "must spawn a fresh worker" invariant holds for any caller and survives refactors. Pushed in `0e954dd2`.
- **[Not a bug] MEDIUM — double-processing via concurrent map mutation.** The re-post loop over `pendingRequests` is fully synchronous (no `await` in the loop body; `postMessage` is sync), so the map cannot be mutated mid-iteration and no request is posted twice. This is the same long-shipped pattern as the OOM `resubmitPending` path. New embeds that arrive during the preceding `await ensureWorker()` are added before the loop and posted exactly once.

Branch also cleaned of unrelated concurrent-workstream files — `git diff --name-only origin/main HEAD` is exactly the six #1379 files.

PR #1388, opened 2026-07-19
feat(gateway): GitHub repo-collaborator discovery (#630, E-5-d Slice 1)

BYK:
Part of #630 (re-scoped — **no GitHub App required**; see the issue comment). First slice of the collaborator discovery/invite flow, reusing the E-5-a OAuth + server-verified Edge Function pattern.

## What
`lore team discover [repo...]` shows which of your GitHub repo collaborators already have a Lore account (and which don't), so you can invite the rest to a team you admin.

## How (mirrors `github-provision`)
- **Client** acquires a **fresh** GitHub `provider_token` via a one-shot loopback OAuth (`acquireGitHubProviderToken()` in login.ts) — the token is short-lived and not persisted, so it's re-acquired on demand — then calls the new **`github-discover` Edge Function**.
- **Edge Function** verifies the JWT, **binds the provider_token** to the caller's linked GitHub identity (`fetchGitHubUser` + `isTokenOwnerBound`, fail-closed), reads the caller's repos + each repo's collaborators **from GitHub with the caller's own token** (GitHub authorizes on their access — an unreadable repo 403s and is skipped), then resolves which collaborator github ids are on Lore via the **service-role-only `lore_users_for_github_ids` RPC** (migration 0050).

## Security / privacy 🔴
- `provider_token` always owner-bound (a leaked/foreign token can't enumerate as this user).
- Lore-membership disclosed **only** for collaborators of repos the caller can already read.
- **Never** returns Lore user_ids — only an `onLore` boolean per collaborator.
- The lookup RPC is **service-role-only** → not an open "is X on Lore" oracle. Keyed on `auth.users.raw_user_meta_data->>'provider_id'` (works in prod + the test shim).
- `parseRepoRef` rejects traversal-y input; `MAX_REPOS` caps GitHub fan-out per call.

## Scope / limitations
- Private-repo collaborator reads need broader than `read:org`; v1 covers repos the caller can read and skips the rest.
- The actual **invite wiring** (`--invite <scope>` → `create_scope_invite`) and **repo→team correlation** are **E-5-d-2** — the server has no repo→scope binding (`git_remote` is encrypted on the wire), so that correlation is client-side.
- The `github-discover` Edge Function is not auto-deployed (like `github-provision`): `supabase functions deploy github-discover`.

## Tests
- `github-discover.test.ts`: pure helpers (`parseRepoRef`, `fetchUserRepos`, `fetchRepoCollaborators` incl. 403/404-skip, `collectGithubIds`/`annotateOnLore`) with a mocked fetch — **12**.
- `sync-github-discover.integration.test.ts` (Tier-1, real Postgres): the 0050 lookup RPC — membership match, github-id-only output, ignores no/invalid provider_id, **service-role-only** (42501 on authenticated/anon) — **6**.
- Typecheck clean; 18 tests pass; format clean.

### DIFF
--- a/packages/core/src/embedding-worker-types.ts
+++ b/packages/core/src/embedding-worker-types.ts
@@ -246,6 +246,30 @@ export function shouldRequestWasmRespawn(
 }
 
 /**
+ * Whether `processEmbed`'s catch block should post a per-request `error` back to
+ * the main thread for a failed embed request (#1379/#1387-B2).
+ *
+ * Two cases must stay SILENT:
+ *  - `initFailed` — `ensurePipeline()` already posted `init-error`; re-posting a
+ *    per-request error would double-report the same failure.
+ *  - `wasmRespawnRequested` — the worker asked the main thread to respawn it
+ *    forcing WASM, so `ensurePipeline()` rejected this request with "awaiting
+ *    WASM respawn". Posting a per-request `error` here would make the main thread
+ *    reject+drop the pending request BEFORE `respawnForWasm()` re-submits it to
+ *    the fresh WASM worker — silently losing the caller's embed. The respawn
+ *    re-submits it instead, so this worker must not report it.
+ *
+ * Pure predicate, extracted so the B2 suppression is directly unit-testable
+ * without importing the worker module (which self-executes on import).
+ */
+export function shouldPostPerRequestError(
+  initFailed: boolean,
+  wasmRespawnRequested: boolean,
+): boolean {
+  return !initFailed && !wasmRespawnRequested;
+}
+
+/**
  * The two leading strings transformers.js' `sessionRun()` (models.js) passes to
  * `console.error` when `session.run` throws, before it re-throws: the error
  * message, and a dump of every input tensor's metadata AND `.data`. For our
--- a/packages/core/src/embedding-worker.ts
+++ b/packages/core/src/embedding-worker.ts
@@ -744,6 +744,8 @@ async function processEmbed(req: EmbedRequest): Promise<void> {
     // "awaiting WASM respawn", but posting a per-request `error` here would make
     // the main thread reject+drop the pending BEFORE respawnForWasm() can
     // re-submit it to the fresh WASM worker. The respawn re-submits it instead.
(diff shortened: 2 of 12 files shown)


BYK:
Adversarial + security review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs (done in an isolated worktree; main tree verified untouched).

Verified all 10 points, security core solid:
- **Enumeration oracle bound (primary)**: `lore_users_for_github_ids` is genuinely service-role-only (`revoke ... from public, anon, authenticated` neutralizes PG's default PUBLIC grant; mutation-verified — granting to `authenticated` fails the isolation tests). No client path reaches it; the only enumeration input is collaborators of repos the caller has **push access** to (GitHub 403s the rest). Privacy bound holds.
- **Token binding**: provider_token bound to the JWT's linked GitHub id *before* any GitHub read; fails closed on email-only accounts.
- **RPC**: github-ids-only output (no user_id/email), `~ '^[0-9]+$'` guard before `::bigint`, SECURITY DEFINER + pinned search_path.
- All 3 requested mutations killed; typecheck clean; PR files lint-clean.

**Addressed the one SHOULD-FIX inline:** `parseRepoRef` accepted bare `.`/`..` segments (charset allowed dots), contradicting its 'path-traversal defense' comment. Not exploitable (URL collapses to invalid GitHub endpoints, always the caller's own token), but the invariant should hold — now rejects exact `.`/`..` owner/name segments, with new unit cases (`owner/..`, `../x`, `owner/.`, `./x`, `../..` → null). Unit tests 12→13.

Deferred (non-blocking, noted for later): assert refreshed-OAuth user matches the logged-in account in `acquireGitHubProviderToken`; optional expression index on `provider_id` at scale. The invite wiring + repo→team correlation are the planned E-5-d-2 follow-up.

BYK:
Addressed the Seer MEDIUM finding (stale client): `discover` now uses the **fresh** `client` returned by `acquireGitHubProviderToken()` for the `github-discover` call, instead of the older `getAuthedClient()` instance whose token could have expired during the interactive OAuth flow.

PR #1391, opened 2026-07-19
fix(core): record stated facts neutrally, strip turn-scoped 'don't act on this' framing (#1383)

### LINKED ISSUE #1383: Cheaper worker models poison distillations by editorializing 'throwaway aside' facts as 'do not act on'
## Finding (from #961 core A/B suite, post-merge validation)

In the `pref-long` single-session eval, a code-invisible fact is planted once as a **throwaway aside**: *"nothing to act on or write down — but: orders ride the WHOLESALE channel, EMEA region, warehouse WH-07, status SUBMITTED uppercase."* The final probe turn later asks the agent to "build it the way I like things done on this project" (without restating the values), so the agent must recall and **apply** the aside.

### Two distinct failure modes observed on MiniMax-M3

**1. Worker-distiller poisoning (Lore-side, actionable).** With M3 as the Lore background worker, the distiller frequently records the aside with a suppression flag baked into durable memory:
- run1: `"...aside should NOT be acted on or written down (orders ride WHOLESALE channel, EMEA region...)"`
- run2: `"(do not act on or write down): orders ride the WHOLESALE channel..."`
- run4: `"as not-to-be-recorded: orders ride the WHOLESALE channel..."`

The values ARE captured, but the surrounding text tells any future reader **not to use them**. When this distillation is later recalled into system[2], the memory actively discourages application — Lore is surfacing an anti-instruction.

Contrast: on `pref-combined` (different phrasing), the same M3 distiller recorded clean conventions — `"every order carries channel='WHOLESALE', region='EMEA', warehouse='WH-07'"` — and the agent scored 100%. Frontier Sonnet's distiller recorded `pref-long` cleanly too (`"User's orders use channel WHOLESALE... stated as an aside, no action requested"`).

**2. Fact-application capability (model-side, not fixable in Lore).** In run5 the M3 distiller recorded the aside *cleanly* (no poison flag) — yet M3 still produced `channel=""`, `region=""`, `warehouse=""` in the probe file. So even with clean recall available, M3 doesn't reliably apply incidental recalled facts on the single-long task. Sonnet applied them 5/5.

### Data (N=5 each)
| cell | pass% |
|---|---|
| pref-long lore-m3 | 0% |
| pref-long lore-sonnet | 100% |
| pref-combined lore-m3 | 100% |
| pref-combined lore-sonnet | 100% |

### Proposed Lore-side action (mode 1 only)

The distiller prompt should capture stated facts/conventions as **neutral latent context** even when the user says "don't act on this now" — record the *fact*, not a durable "never use this" directive. "Nothing to act on right now" is a turn-scoped instruction, not a permanent property of the fact. Cheaper workers over-literally propagate the suppression into memory, which is strictly worse than recording nothing.

Mode 2 (application capability) is a model limitation and out of scope — documented for the benchmark, not a Lore bug.

Repro: eval `task-pref-long.json`, `--arm lore --model minimax-coding-plan/MiniMax-M3`; inspect `distillations.observations` for the WHOLESALE aside framing.

BYK:
Fixes #1383 (mode 1 — Lore-side distiller poisoning).

## Problem

In the `pref-long` eval, a fact is planted as a throwaway aside: *"nothing to act on or write down — but: orders ride the WHOLESALE channel, EMEA region, warehouse WH-07, status SUBMITTED uppercase."* Cheaper worker distillers (MiniMax-M3) capture the values but **bake the turn-scoped suppression into durable memory**:

- `"...aside should NOT be acted on or written down (orders ride WHOLESALE channel...)"`
- `"(do not act on or write down): orders ride the WHOLESALE channel..."`
- `"as not-to-be-recorded: orders ride the WHOLESALE channel..."`

When that distillation is later recalled into `system[2]`, the memory actively tells a future agent **not to use the fact** — Lore surfaces an anti-instruction. Frontier Sonnet's distiller recorded it cleanly (`"...stated as an aside, no action requested"`), and M3 on differently-phrased `pref-combined` also recorded clean conventions and scored 100%. So the framing is the lever.

## Fix

New section in `DISTILLATION_SYSTEM` (the observer prompt): **record the fact itself as neutral latent context**, and treat "nothing to act on / don't write this down / just FYI" as a **turn-scoped instruction about the current turn, not a permanent property of the fact**. Never bake a "do not use / not to be recorded" directive into a durable observation; a future reader must be able to apply the fact. An optional "stated as an aside; no action requested this turn" note preserves that it was incidental without telling a future reader to discard it.

Includes the exact WHOLESALE case as BAD (poisoned framings) vs GOOD (neutral values + incidental note) examples, matching the prompt's existing BAD/GOOD teaching style.

Scope: mode 2 (M3's fact-*application* capability on the single-long task, even with clean recall) is a model limitation, explicitly out of scope per the issue.

## Tests

`distillation.test.ts` — new `DISTILLATION_SYSTEM — neutral fact recording (#1383)` block asserts the guidance, the named suppression phrases, the "recording nothing beats a 'never use this' fact" rule, and that the WHOLESALE example is GOOD (neutral, no suppression directive) with the poisoned framing shown as BAD. Non-vacuous: removing the prompt section fails all 4 tests. Full `distillation.test.ts`: 126 passed; typecheck + format clean.

### DIFF
--- a/packages/core/src/prompt.ts
+++ b/packages/core/src/prompt.ts
@@ -17,6 +17,23 @@ When the user ASKS about something, mark it as a question (🟡):
 
 User assertions are AUTHORITATIVE — the user is the source of truth about their own life.
 
+RECORD FACTS NEUTRALLY — STRIP TURN-SCOPED "DON'T ACT ON THIS" FRAMING:
+
+When the user states a fact, convention, or preference, record the FACT ITSELF as neutral latent context — even when they wrap it in "no need to do anything with this", "nothing to act on", "don't write this down", "just FYI", or "for later". Those phrases are TURN-SCOPED instructions to the assistant about the current turn; they are NOT a permanent property of the fact. A future reader must be able to APPLY the fact, so never bake a "do not use / do not act on / not to be recorded" directive into the durable observation.
+
+Record the stated values plainly. Do NOT editorialize them as suppressed, ignorable, or off-limits.
+
+BAD: 🔴 (14:30) User mentioned an aside NOT to act on or write down: orders ride the WHOLESALE channel, EMEA region, warehouse WH-07, status SUBMITTED (uppercase)
+BAD: 🔴 (14:30) User noted (do not act on or write down) that orders use channel WHOLESALE...
+BAD: 🔴 (14:30) User gave a not-to-be-recorded aside about order conventions
+GOOD: 🔴 (14:30) User stated their orders use channel=WHOLESALE, region=EMEA, warehouse=WH-07, status=SUBMITTED (uppercase). Stated as an aside; no action requested this turn.
+
+The optional "stated as an aside / no action requested this turn" note preserves that it was incidental WITHOUT telling a future reader to discard it. When in doubt, record the fact — recording nothing is better than recording a fact tagged "never use this", which is strictly worse than silence.
+
+TWO EXCEPTIONS — cases where suppression IS a permanent property, not turn-scoped framing:
+- SECRETS: genuine secrets (passwords, API keys, tokens, private keys) and anything the user asks to be forgotten/scrubbed — here "don't write this down" is real. OMIT the value; a bare reference is fine ("User provided a prod DB password" — never the password itself).
+- CORRECTIONS / HYPOTHETICALS: "ignore that, it's wrong", "I misspoke", "that's hypothetical" are NOT the incidental-aside case above — they are handled by STATE CHANGES. Keep the corrected/final value, not the mistaken or hypothetical one; do not record the retracted value as a live fact.
+
 TEMPORAL ANCHORING — CRITICAL FOR TEMPORAL REASONING:
 
 Each observation has up to two timestamps:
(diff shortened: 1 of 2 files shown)


BYK:
## Adversarial review — SHIP-WITH-FOLLOWUPS, both follow-ups applied

Review confirmed no BLOCKERs and — the decisive question — **no secret-recording regression**: the baseline prompt already records secret *references* (`CODECOV_TOKEN` etc.) and has no scrub layer, so this change doesn't move that needle. It landed both SHOULD-FIX hardenings in this PR (each one clause + tests):

- **Secrets carve-out** — genuine secrets (passwords, API keys, tokens, private keys) and content the user asks to be forgotten are the case where "don't write this down" IS permanent: omit the value, keep a bare reference ("User provided a prod DB password" — never the value). Directly closes the reviewer's highest-value concern.
- **Correction/hypothetical guard** — "ignore that, it's wrong" / "I misspoke" / "hypothetically" are supersession (STATE CHANGES), NOT incidental-aside framing to strip: keep the corrected/final value, not the retracted one. Closes the over-reach risk.

Tests extended (now 6 in the block); both new tests are non-vacuous (removing the exceptions block fails them). Full `distillation.test.ts`: 128 passed; typecheck + format clean.

PR #1392, opened 2026-07-19
feat(gateway): email team invites via SMTP2GO (E-5-e, #630/#827)

BYK:
Builds on the invite infra so `lore team invite --email <addr>` **actually emails** the invitee a join link, instead of only printing a token for the admin to hand-deliver. This is the shared email foundation the collaborator-discovery invite path (E-5-d-2) will reuse.

## What
- New **`send-invite-email` Edge Function** (SMTP2GO HTTP API, sender `keeper@withlore.ai`): pure `send.ts` core (`buildInviteEmail` + `sendViaSmtp2go`, injectable fetch) + thin Deno `index.ts` glue (mirrors `github-discover`).
- `sendInviteEmail` + `isEmailAddress` in team.ts; the CLI `invite` case emails when `--email` is a real address and **always still prints the token link** (a send failure never loses the invite).

## Security / abuse 🔴
- The Edge Function authorizes on the invite **token**, not client input: resolves `pending_invites` service-role, then requires the caller be an **admin** of the invite's scope (`scope_role`) **and** its **creator** (`invited_by = auth.uid()`). → not an open email-spam relay: you can only email an invite you created for a team you admin, one recipient per call. Expired/absent → generic 404 (no oracle).
- Online (auto-wrap) invite emails carry **no** decryption-capable material; offline (`eph:`) tokens are decryption-capable by design (same posture as today, just emailed — copy warns to treat like a password).
- `SMTP2GO_API_KEY` is a server-only Function secret, never shipped to clients / never logged.

## Scope / notes
- **No migration** (reads existing `pending_invites`/`scopes` only).
- GitHub's `/collaborators` API returns no email → GitHub-discovered collaborators (E-5-d-2) can't be auto-emailed; that path prints links and emails only where the admin supplies an address.
- Not auto-deployed: `supabase functions deploy send-invite-email` + set `SMTP2GO_API_KEY`.

## Tests
- `send-invite-email.test.ts`: `buildInviteEmail` (accept cmd/role/expiry, unknown-role→editor, offline-only warning, HTML escaping) + `sendViaSmtp2go` (key header + recipient, non-ok throw, API-error throw, no-recipient throw) — **9**, mocked fetch.
- `sync-team-invite.integration.test.ts`: Tier-1 test asserting the two authorization signals the Edge Function reads (admin role + `invited_by` creator binding).
- Typecheck clean; format clean. (Pre-existing `reconcileScopeWraps` integration test fails in my local Docker env but is green in CI on main @38b44a31 — unrelated.)

### DIFF
--- a/packages/gateway/src/cli/team-cmd.ts
+++ b/packages/gateway/src/cli/team-cmd.ts
@@ -23,11 +23,13 @@ import {
   createTeam,
   createTeamInvite,
   discoverGitHubCollaborators,
+  isEmailAddress,
   listDomainJoinRequests,
   listTeams,
   rejectDomainJoin,
   removeTeamMember,
   requestDomainJoin,
+  sendInviteEmail,
   setTeamRole,
   teamMembers,
 } from "../team";
@@ -194,11 +196,22 @@ export async function commandTeam(
         const token = await createTeamInvite(client, scope, role, hint, {
           offline,
         });
+        // E-5-e: if --email is a real address, email the join link. A send failure NEVER loses the
+        // invite — the token/link is always still printed.
+        const willEmail = !!hint && isEmailAddress(hint);
+        const emailed = willEmail
+          ? await sendInviteEmail(client, token, hint)
+          : false;
         console.log(
           `Invite created (${role}${hint ? `, for ${hint}` : ""}${offline ? ", offline" : ""}). It expires in 14 days.\n` +
             (offline
               ? "This token carries a one-time decryption key — treat it like a password and prefer a private channel.\n"
               : "") +
+            (willEmail && emailed
+              ? `Emailed the join link to ${hint}.\n`
+              : willEmail
+                ? "Could not send the email — share the link manually below.\n"
+                : "") +
             `Share this with the invitee — they run:\n\n  lore team accept ${token}\n`,
         );
         break;
--- a/packages/gateway/src/team.ts
+++ b/packages/gateway/src/team.ts
@@ -297,6 +297,31 @@ export async function createTeamInvite(
   return `${data as string}${ephSecretSeg}`;
 }
 
+/** Conservative "looks like an email" check — decides whether `--email` is a real address to send to. */
+export function isEmailAddress(s: string): boolean {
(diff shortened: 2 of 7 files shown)


BYK:
Adversarial + security review: **SHIP-WITH-FOLLOWUPS** (isolated worktree; main tree verified untouched).

Security core verified: token-anchored **triple gate** (token exists+unexpired **AND** caller is scope admin **AND** caller is the invite creator) all enforced before the SMTP2GO send with correct `!isAdmin || !isCreator` logic and no missing awaits; no relay amplification (caller already holds the token); generic 404 for absent/expired (no oracle); API key never logged; fail-closed on missing config; HTML escaping + role/token safety — all mutation-verified (3/3 kills: missing key header, no-recipient guard, teamName escaping). The one integration failure (`reconcileScopeWraps`) is confirmed **pre-existing on green-CI main** (@38b44a31), not from this PR.

**Fixed the BLOCKER** (2 new oxlint `correctness` errors that would red the lint step):
- `team-cmd.ts:203` `no-unnecessary-type-assertion` — dropped the redundant `hint as string` (narrowed by the `willEmail` guard).
- `send-invite-email.test.ts:63` `no-base-to-string` — narrowed the mock's `url` param to `string` (matches how `sendViaSmtp2go` calls fetch). oxlint now clean on all PR files; typecheck + format clean; 9 unit tests pass.

Deferred (non-blocking): treat a missing `data.succeeded` on a 200 as failure (currently resolves ok — acceptable per SMTP2GO contract, best-effort UX); extract the index.ts authz decision into a pure testable helper in a later slice.

BYK:
Fixed the Seer MEDIUM (offline-invite lookup mismatch): the Edge Function now splits the token and looks up `pending_invites` by the **capability portion** only (via a new pure `capabilityOf` helper in send.ts, mirroring `acceptTeamInvite`), while still emailing the **full** token so an offline invitee gets the ephemeral secret suffix needed to unwrap the DEK. Added `capabilityOf` unit tests (capability-only unchanged; `cap.SECRET`→`cap`). Was a real bug — offline `--email` invites would have 404'd on the DB lookup. 11 unit tests pass; lint/typecheck/format clean.

Issue #1178, opened 2026-07-05
N+1 sweep follow-ups: per-candidate FTS (3E) and per-tool-call tool_calls insert (3G)


BYK:
Follow-ups from the Sentry N+1 sweep since 0.34.0. The clean, high-value one (LOREAI-GATEWAY-3D, `outcomeImpact` on the consolidation path) is fixed in #1177. The two below are lower value and left for a focused follow-up.

### LOREAI-GATEWAY-3E — per-candidate distillation FTS (`lore.curator`)
- **Where:** `instruction-detect.ts` → `findRepeatedInstructions` loops over candidate instructions; each iteration runs `searchDistillationsFTS` (`instruction-detect.ts:242`):
  ```sql
  SELECT d.id, d.session_id
    FROM distillation_fts f CROSS JOIN distillations d ON d.rowid = f.rowid
   WHERE distillation_fts MATCH ? AND d.project_id = ?
   ORDER BY rank LIMIT 30
  ```
- **Why not folded into #1177:** genuinely per-candidate — each candidate has a distinct `MATCH` expression and its own `LIMIT 30`/`rank`, so results can't be merged without losing per-candidate attribution. The same loop *also* does a per-candidate `vectorSearchAllDistillations` (`instruction-detect.ts:191`), so batching only the FTS wouldn't fully flatten the transaction's span profile.
- **Options:** (a) single `UNION ALL` of per-candidate tagged sub-selects (1 statement, N internal indexed FTS scans → collapses the Sentry N+1 span); (b) leave as-is — both paths are **local, indexed** SQLite on the background idle/curation path, so real cost is low.

### LOREAI-GATEWAY-3G — per-tool-call `tool_calls` upsert (`POST /v1/chat/completions`)
- **Where:** `temporal.ts` → `recordToolCalls` loops `seedStmt.run(...)` / `updateStmt.run(...)` per tool part (`temporal.ts:191`).
- **Notes:** already runs inside a single savepoint; 1 event observed. Could become a multi-row `INSERT ... VALUES` but the per-part INSERT-vs-UPDATE branching makes it awkward for marginal benefit.

Both run on background/local-SQLite paths with indexed queries, so impact is minor. Filing to track rather than rush a riskier refactor.

BYK:
Both N+1 follow-ups from the Sentry sweep are now resolved:

[...]

Issue #796, opened 2026-06-16
fix(gradient): inject LTM on turn 1 for already-large resumed sessions (skip the turn-2 defer)


BYK:
## Problem
Context-bound LTM (the system-prompt knowledge block) is intentionally **deferred to turn 2** of a newly-tracked session. For genuinely *fresh* small sessions this is correct (avoids paying for LTM before we know the conversation needs it). But for a session that is **already large on first sight** (a resumed conversation Lore hasn't tracked in this process — `new=true` with a big message count), deferring LTM wastes the turn-1 cold-cache write: turn 1 writes the entire body *without* LTM, then turn 2 re-injects LTM into the system prefix and **rewrites the whole body again** → a full front-bust.

## Production evidence
Session `ses_156668952ffeDIrhFHaU8Frjdq` (internal `12na8hXrpthm52Hl`, project witty-garden, opus-4-8, **761 messages**), 2026-06-16 ~22:00:

| turn | hit | divergence | event |
|---|---|---|---|
| 1 (21:59, `new=true`) | — | cold | first sight of 761-msg convo → full write (`write=622099 read=0`) |
| 2 (22:00) | 8% | `system[1]` | LTM knowledge block injected → system prefix shifts → **bust** |
| 3 (22:03) | 0% | `messages[0]` | size threshold crossed → Layer 0→1 re-window → **bust** |
| 4 (22:04) | 99% | `messages[646]` | stabilized |

Divergence snippet at turn 2 confirms it: `prev` had no `## Long-term Knowledge` block, `curr` inserted it.

## Proposed fix
When a session is seen for the first time in-process (`new=true`) **and** the incoming conversation is already large (e.g. above the layer-0 cap / a message-count or token threshold), inject context-bound LTM on **turn 1** instead of deferring to turn 2.

**Key insight (per @ben):** if LTM is injected on turn 1 for an already-large session, the gradient would *immediately* see the conversation is too large and kick in compression on turn 1 — collapsing all three stabilization busts into the single, unavoidable cold-start write. The turn-2 `system[1]` bust and the turn-3 `messages[0]` layer-transition bust both disappear because LTM + compression are decided together up front.

Keep the turn-2 defer for genuinely fresh/small sessions (the original rationale stands).

## Notes
- Discovered while investigating a live "unsustainable conversation" warning. The warning was transient (recovered by turn 4) but the two extra full-body rewrites (~600K–1.2MB each) are real, avoidable cost on every large resumed session's cold start.
[...]

PR #1395, opened 2026-07-19
perf(core): batch recordToolCalls tool_use seeds into one multi-row upsert (#1178 / 3G)

BYK:
Closes the remaining item in #1178. **3E (per-candidate FTS N+1) was already fixed and merged in PR #1265** (`searchDistillationsFTSBatch`, single `UNION ALL`), so only 3G remained.

## 3G — per-tool-call `tool_calls` upsert (`recordToolCalls`)

`temporal.ts` looped `seedStmt.run(...)` once per `tool_use` part, so Sentry saw an N+1 (one `db.query.run` span per part) under `POST /v1/chat/completions`.

### Change
- Split the parts up front: `tool_use` **seeds** vs `tool_result` **outcome updates**.
- **Seed phase** is now ONE multi-row `INSERT ... VALUES (…),(…) ON CONFLICT(call_id) DO UPDATE …`, chunked at 81 rows (11 cols/row) to stay under SQLite's conservative ~999 bound-variable ceiling — following the existing #796 chunking convention in `db.ts`. All original upsert semantics preserved verbatim: the pending-guard (`CASE WHEN status='pending'`), `duration_ms`/`verifier` `COALESCE`, and name preservation.
- **Result phase** stays per-part `UPDATE`s — each carries distinct outcome values keyed by its own `call_id`, so batching them buys little and complicates the INSERT-vs-UPDATE branching (per the issue's own note).
- Still runs inside the caller's single `withSavepoint("post_response_temporal", …)`; no transaction-nesting change.

SQLite applies multi-row VALUES in order and fires `ON CONFLICT` row-by-row, so a duplicate `call_id` within one batch resolves identically to the old sequential loop.

### Tests (`temporal.test.ts`, all 33 pass; core suite 3065 pass)
New coverage on top of the existing 6:
- batches many seeds into distinct rows;
- mixed `tool_use` + `result` parts in a single call;
- re-seed does NOT revert a resolved row to pending (pending-guard, across the batch);
- inserts every row when the batch spans multiple chunks (chunk-loop correctness);
- duplicate `call_id` within one batch behaves like the sequential loop.

Non-vacuous: reverting the pending-guard fails the guard test; treating results as seeds fails the orphan-no-op test; an off-by-one in the chunk loop fails the multi-chunk test. typecheck + oxlint + format clean.

### DIFF
--- a/packages/core/src/temporal.ts
+++ b/packages/core/src/temporal.ts
@@ -163,37 +163,86 @@ export function recordToolCalls(input: {
   const pid = ensureProject(input.projectPath);
   const createdAt = input.info.time.created;
 
-  // Phase A: assistant tool_use parts — seed name + status. May already be
-  // resolved (e.g. host delivers completed state directly). On conflict, only
-  // overwrite outcome fields when the existing row is still `pending` — never
-  // revert a resolved row back to pending on a re-seed (retry / re-delivery).
-  const seedStmt = db().query(
-    `INSERT INTO tool_calls
-       (call_id, message_id, project_id, session_id, tool, status, error_type, error_message, duration_ms, created_at, verifier)
-     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
-     ON CONFLICT(call_id) DO UPDATE SET
-       status = CASE WHEN tool_calls.status = 'pending' THEN excluded.status ELSE tool_calls.status END,
-       error_type = CASE WHEN tool_calls.status = 'pending' THEN excluded.error_type ELSE tool_calls.error_type END,
-       error_message = CASE WHEN tool_calls.status = 'pending' THEN excluded.error_message ELSE tool_calls.error_message END,
-       duration_ms = COALESCE(excluded.duration_ms, tool_calls.duration_ms),
-       -- verifier is derived from the tool_use input (stable across re-seeds);
-       -- keep the first non-null classification.
-       verifier = COALESCE(tool_calls.verifier, excluded.verifier)`,
-  );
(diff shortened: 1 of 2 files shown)


BYK:
## Adversarial review — SHIP-WITH-FOLLOWUPS, all three NITs applied

Independent review proved behavior-equivalence with the old per-part loop on the real `node:sqlite` driver: a 6-case battery (plain multi-seed, intra-batch duplicate call_id pending & differing, completed-then-pending guard, verifier COALESCE first-non-null, error seeding) produced **byte-identical** DB state; intra-statement `ON CONFLICT` confirmed row-by-row (duplicate call_id within one VALUES resolves against the earlier row, matching two sequential runs); all three mutation claims reproduced RED. No BLOCKERs/SHOULD-FIX.

The reordering-safety point: the seed/result split relies on the invariant that a single `parts` array never mixes a tool_result and the tool_use seed for the *same* call_id — verified true for both production call sites (tool_use only on assistant messages, tool_result only on user messages; each caller passes one message's parts).

Landed all three NITs in this PR:
- **NIT-1**: corrected the stale "999 ceiling" comment — the shipped driver's limit is 32766; the ~900-param chunk budget is the portable #796 convention, well under both.
- **NIT-2**: documented the phase-split invariant in `recordToolCalls` (with a "restore in-order processing if a future caller passes a mixed array" note).
- **NIT-3**: `rows()` test helper now selects `verifier` + `message_id`; added two tests — per-row verifier classification survives batching (non-vacuous: hardcoding verifier=0 fails it), and a result UPDATE preserves the batched seed's `message_id`.

Now 35 temporal tests pass (was 33); core suite green; typecheck + oxlint + format clean.

PR #1396, opened 2026-07-19
feat(gateway): wire discover→invite + auto-deploy Edge Functions (E-5-d-2, #630)

BYK:
Closes the loop from E-5-d Slice 1 (#1388): `lore team discover [repo...] --invite <scope>` now **mints team invites** for the discovered collaborators. Also adds **auto-deploy** for the Supabase Edge Functions (so we stop relying on manual `supabase functions deploy`).

## discover → invite
- **Invite-only for everyone** (on-Lore or not): mints one `create_scope_invite` per **distinct** collaborator (deduped by GitHub login via `distinctCollaborators`) and prints a `lore team accept <token>` link each. We never expose Lore user_ids → no direct auto-add; an on-Lore invitee just gets a frictionless accept.
- **Scope resolution**: `--invite <team-name-or-id>` via `resolveWritableScope`; if that doesn't resolve and the cwd is a project linked with `lore team link`, falls back to that linked scope (**repo→scope correlation**). `--role editor|viewer` (default editor). `invite` declared in the global OPTIONS map (undeclared string flags silently become booleans under `strict:false` — #696).
- GitHub's `/collaborators` returns no email → links are printed; use `lore team invite <scope> --email <addr>` (E-5-e) to email a known address.

## Auto-deploy Edge Functions 🔧 (sustainability)
- New **`.github/workflows/deploy-functions.yml`**: on push to `main` touching `supabase/functions/**` or `config.toml`, runs `supabase functions deploy` for the project. Migrations already deploy via the Supabase↔GitHub integration; **functions did not** — they relied on a human remembering to deploy from a local branch (unsafe/unsustainable).
  - **⚠️ One-time setup required:** add a `SUPABASE_ACCESS_TOKEN` repo secret (Supabase → Account → Access Tokens; `gh secret set SUPABASE_ACCESS_TOKEN`). The job fails fast with a clear message if absent. Function *secrets* (e.g. `SMTP2GO_API_KEY`) are still set via `supabase secrets set` — not managed by CI.
- Added `[functions.github-discover]` + `[functions.send-invite-email]` to `config.toml` (verify_jwt=true) so all three functions are declared (matching github-provision).

## Tests
- `distinctCollaborators` unit (dedupe across repos / first-occurrence wins / sort / empty).
- CLI `discover --invite`: one invite per distinct person (bob on 2 repos → 1 invite), `--role viewer`, scope-resolution failure → exit 1, and no-`--invite` lists only.
- Typecheck / lint / format clean; 205 gateway unit tests green across the touched suites.

### DIFF
--- a/.github/workflows/deploy-functions.yml
+++ b/.github/workflows/deploy-functions.yml
@@ -1,0 +1,49 @@
+name: Deploy Supabase Functions
+
+# Auto-deploy the Edge Functions on merge to main so we never rely on someone remembering to run
+# `supabase functions deploy` from a local branch (unsafe + unsustainable). Migrations already deploy
+# via the Supabase↔GitHub integration; this covers the functions, which that integration does NOT.
+#
+# One-time setup (required): add a repo secret SUPABASE_ACCESS_TOKEN
+#   (Supabase dashboard → Account → Access Tokens → generate; then
+#    `gh secret set SUPABASE_ACCESS_TOKEN`). Function SECRETS (e.g. SMTP2GO_API_KEY) are set
+#   separately via `supabase secrets set` and are NOT managed here.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "supabase/functions/**"
+      - "supabase/config.toml"
+      - ".github/workflows/deploy-functions.yml"
+  # Manual re-deploy escape hatch.
+  workflow_dispatch:
+
+# Never let two deploys race; the newest merge wins.
+concurrency:
+  group: deploy-functions
+  cancel-in-progress: false
+
+jobs:
+  deploy:
+    name: Deploy Edge Functions
+    runs-on: ubuntu-latest
+    # The project ref is not a secret (it's in supabase/config.toml); the access token is.
+    env:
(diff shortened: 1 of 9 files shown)


BYK:
Adversarial + security review: **SHIP-WITH-FOLLOWUPS**, no BLOCKERs (isolated worktree; main tree untouched). Verified the security boundary is server-side (`create_scope_invite` requires scope admin); client resolution is a UX gate; no user-id leak; invite-only model intact; batch resilience + capability-only tokens correct; actionlint clean on the deploy workflow. All 3 mutations killed.

Folded in all 3 non-blocking nits:
1. **Trailing `--invite` (no value)** now rejected with "`--invite` requires a team name or id" (was a cryptic SQLite bind error under strict:false).
2. **Editor-caller pre-check**: added `scopeMemberRole(scopeId, userId)` to core; the CLI now checks admin **once** before the loop and errors with "You must be an admin of …" instead of firing N per-collaborator RPCs that each get rejected server-side.
3. Added the **editor-rejected** regression test (asserts exit 1 + no RPC fired).

typecheck (core + gateway) / lint / format clean; team-cmd tests 56→57. Rebased onto latest main.

PR #1416, opened 2026-07-21
feat(pattern-extract): gate noisy titles so seeds stay discoverable

BYK:
## What

Follow-up to #1414 (D1c in the Modem *"[how coding agents read your code](https://modem.dev/blog/how-coding-agents-read-your-code)"* plan). #1414 taught the **LLM curator** to write discoverable titles; this fixes the **non-LLM `pattern-extract`** seed path, which was minting the exact junk titles the article warns against.

`pattern-extract.ts` builds a knowledge title as `prefix + raw capture group`, where the capture is bounded only by `(.+?)(?:\.|,|$)`. A capture with no early comma/period runs into multi-clause prose, producing titles like:

- `Always the same: the agent never called the save step in the first`
- `Chose checkout under different path, NULL git_remote); 4) converge…`

These are the opposite of a discoverable title (a short, specific, *searchable* term) — and several such entries already exist in this repo's own `.lore.md`.

## Change

Adds `isNoisyTitle(title)`: a minted title is rejected when it

- exceeds a word/char cap (**10 words / 72 chars**), or
- carries **clause punctuation** (`;` `:` `(` `)` or an `N)` enumerator) that only appears when prose leaked past the intended `,`/`.` boundary.

Rejecting is safe — `pattern-extract` is a best-effort seed, and **skipping a bad mint is strictly better than creating a low-discoverability entry** the curator later has to clean up (and whose delete busts the prompt cache).

### Design notes
- **Gate is on the FINAL title, not per-capture.** Some patterns capture a trailing clause they never put in the title (the `because Y` tail of `going with X because Y`); gating that tail would wrongly drop a clean `Going with X`. Verified by a dedicated test.
- The only `titleFn` prefix that legitimately carries punctuation, `Convention: `, is stripped before the noise scan so its colon isn't mistaken for leaked prose.

## Tests

`pattern-extract.test.ts` — new `noisy-title rejection` block: rejects colon / paren+enumerator / semicolon / over-long-prose titles; and two no-over-reject cases (a clean title still mints; a trailing `because` clause doesn't drop `Going with X`). All 79 pattern-extract tests pass.

Mutation-verified in a throwaway copy: removing the `isNoisyTitle` gate fails exactly the 4 rejection tests (the 2 positive tests correctly still pass). Real tree confirmed unchanged (sha256).

Full `@loreai/core` suite green (147 files / 3121 tests). Typecheck, format, lint clean.

## Not in this PR (remaining plan follow-ups)
- D1b (re-titleable `update` op — touches append-only versioning + the `LOWER(title)` dedup key; needs its own review)
- D2c (entry↔file association), D3 (blog post)

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -90,8 +90,14 @@
 <!-- lore:019f3974-7c6a-7091-8602-b4ec0bb763a4 -->
 * **Chose next task: knowledge\_meta seed-order alignment (self-contained, low-risk, completes value-ranked seed coherence work started with knowledge #1189 and entities #1194) — over #1191b (continuous top-N eviction)**: Completed value-ranked seed coherence work in sequence: #1189 (knowledge seed value-ranking) → PR #1194 (entity-graph seed: entities ranked by knowledge\_entity\_refs count DESC then recency; entity\_relations by recency; entity\_aliases by created\_at DESC) → PR #1196 (knowledge\_meta seed aligned to knowledge's ORDER BY: confidence DESC, COALESCE(last\_reinforced\_at, updated\_at) DESC, logical\_id tiebreak). Deferred #1191b (continuous top-N eviction for remote) as a separate larger/riskier task needing a design doc first (remote eviction, cap discovery, thrash guards, concurrency) — not suitable for autonomous same-session implementation. Follow-up issue #1197 filed for a known non-blocking edge case: a deleted high-confidence entry's orphaned knowledge\_meta row can outrank a live entry's meta row under the cap (self-healing, not data loss).
 
+<!-- lore:019f821b-dbd6-74c8-b138-a378a7d3c1e3 -->
+* **Chose PostgreSQL over MySQL for JSONB support**: Chose PostgreSQL over MySQL for JSONB support.
+
 <!-- lore:019f4629-d9d0-7c59-924c-dc7341e57c77 -->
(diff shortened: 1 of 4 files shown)


BYK:
Adversarial review pass (no BLOCKERs). Addressed the findings:

- **SHOULD-FIX (test-coverage gap):** the two length caps weren't independently covered — the single over-long test (124 chars / 19 words) tripped both `MAX_TITLE_CHARS` and `MAX_TITLE_WORDS` at once, so each masked the other's mutant. Added two cap-isolating tests: one title >72 chars but ≤10 words (isolates the char cap), one >10 words but ≤72 chars with no clause punctuation (isolates the word cap). Mutation-verified in a throwaway copy: disabling each cap alone now fails exactly its isolating test.
- **NIT (redundant regex):** dropped the dead `\b\d+\)` alternative from `TITLE_NOISE_RE` — an `N)` enumerator is already caught by the bare `)`. Now `/[;:()]/` with a comment noting enumerators are covered.
- **NIT (by-design over-rejection):** parens / `::` / ratio-colons in a legit title are dropped. Accepted per the file's stated contract — pattern-extract is a best-effort seed and skipping a bad mint is safer than minting junk the curator must clean up. The LLM curator path (#1414) is the real title source.

Full core suite green (152 files / 3173 tests), typecheck/format/lint clean.

PR #1415, opened 2026-07-20
feat(import): structured-memory import lane for migrating Engram memory

BYK:
## What & why

Closes the ask in #1398 (migrating existing memory-tool memories into Lore). Adds a **second `lore import` lane** that maps **already-curated** memory directly into the knowledge store, bypassing the curator LLM (the entries are already structured, so mapping is fast and free). Ships **Engram**; **mem0** is a documented follow-up.

```bash
lore import --source engram                    # runs `engram export` for you
lore import --source engram --file dump.json   # or an explicit export
lore import --source engram --global           # as cross-project knowledge
lore import --source engram --dry-run          # preview counts
```

## Design

- **Single trust boundary** (`import/schema.ts`): `LoreImportDoc`/`LoreImportEntry` zod, `.strict()`, version-literal gate, `content` bounded at 64K (defense-in-depth; importer truncates to 1200). Adapter output **and** `--file` input are validated here.
- **Importer** (`import/structured.ts`): exact-title → fuzzy dedup (update-on-content-change, else skip), by-title **tombstone-resurrection guard**, 1200-char truncation, confidence clamp, `--global`, `dryRun`. No worker attribution (user-authored).
- **Engram adapter** (`import/sources/engram.ts`): `type`→category map (`bugfix`/`discovery`→`gotcha`, `config`→`architecture`, …), project from session `directory`, truthy `deleted_at` skip, generic pass-through.
- **Registry** (`import/structured-sources.ts`): `execFileSync` (no shell) to run `engram export` or read `--file`.
- **Endpoint** `POST /api/v1/import/structured`: re-validates server-side; **403 in hosted mode** (writes knowledge + triggers a DB-wide embedding backfill → operator-gateway only, mirrors `entities/rebuild`); supports `dry_run`.
- **CLI routing**: `--source`/`--file`/`--global`/`--agent`. A bare `lore import` **never** auto-routes to a structured source, so a user who merely has the `engram` binary on PATH isn't surprised.

## Review rigor

Two independent review passes per `quality/REVIEW.md §6`:

- **Adversarial correctness** → MERGE. Two surviving mutants from the first pass are now closed with regression tests (fuzzy-update branch; `findTombstonedByTitle` live-precedence). Four guards (truncation, fuzzy, `deleted_at`, live-precedence) each **mutation-verified** — reverting the guard turns exactly one test red.
[...]

Issue #1398, opened 2026-07-20
question: how will the migration be with existing memory layer


Kjaer:
Hey there @BYK,

I am using opencode for a while with a pair of engram memory MCP (https://github.com/Gentleman-Programming/engram) If I want to use lore, should I feed the lore with my existing engram memories? would it work along with engram or breaks it?

BYK:
Hey @Kjaer — great question. Short version: **you don't need to migrate anything, and the two can coexist without breaking each other**, but you probably won't want to run both long-term because they overlap.

The reason is that Lore and Engram work in fundamentally different ways:

**Engram is agent-driven, explicit memory (MCP).** Your agent has to decide to call `mem_save` / `mem_search`, and it writes to its own SQLite DB at `~/.engram/engram.db`. Memory only exists when the model remembers to save it.

**Lore is a transparent proxy + plugin.** It sits in the request path, watches the whole conversation automatically, distills it into summaries, extracts long-term knowledge, and injects the relevant bits back into your context on its own. Nothing depends on the model calling a tool. Its data lives in its own DB (`~/.local/share/lore/lore.db`) plus a version-controlled `.lore.md` in each repo.

So to your specific questions:

**Will they break each other?** No. Different storage, different mechanisms. Lore intercepts LLM API calls and manages context; Engram exposes MCP tools. They don't touch the same files or the same code path, so running both is safe — you'd just have two memory systems active at once, and your agent would still see Engram's `mem_*` tools alongside Lore's automatic context injection.

**Should you feed your existing Engram memories into Lore?** There's no import path today, and you mostly don't need one. Lore builds its knowledge automatically from your ongoing sessions, so it'll re-learn your project's patterns/decisions on its own within a few sessions of normal use. If you have important curated memories in Engram that you want Lore to know immediately, the simplest approach is to `engram export` them and drop the salient facts into your repo's `.lore.md` (or just paste them into a session once) — Lore will pick them up. But honestly, letting Lore observe and rebuild is usually less effort than a manual migration.

**Running both together:** it works, but it's redundant and a bit noisy — you'd be paying for Engram's explicit tool calls while Lore is already doing automatic capture. My suggestion: try Lore alongside Engram for a bit, and once you're confident Lore is capturing what you care about, drop Engram to avoid two overlapping memory layers competing for context budget.

TL;DR: no migration required, they won't break each other, but they solve the same problem two different ways — so pick one once you've tried Lore.

BYK:
[...]

PR #1417, opened 2026-07-21
feat(import): native mem0 connector for the structured-import lane

BYK:
## What & why

Completes #1398's migration story: adds a **native, Python-free mem0 connector** to the structured-import lane shipped in #1415. `lore import --source mem0` auto-detects the deployment shape and reads memories straight into the knowledge store (no curator LLM — mem0 memories are already curated).

```bash
lore import --source mem0                    # auto-detect (Qdrant / server / embedded)
lore import --source mem0 --file mem0.json   # explicit dump
lore import --source mem0 --global           # cross-project
lore import --source mem0 --dry-run          # preview
```

## Deployment shapes (`packages/core/src/import/sources/mem0.ts`)

| Shape | Read path |
| --- | --- |
| **Qdrant server** (OpenMemory / raw Qdrant) | HTTP `points/scroll` via `fetch`, `next_page_offset` pagination, reads `payload.data` |
| **mem0 self-hosted server** (FastAPI) | `GET /memories?user_id=…`, `X-API-Key` auth |
| **Embedded default** (`Memory()`) | `collection/<name>/storage.sqlite` opened read-only via `node:sqlite`; each pickled Qdrant `PointStruct` decoded with **`pickleparser`** (pure-TS, MIT, lazy-imported). No Python. |
| **`--file <dump.json>`** | generic `LoreImportDoc` pass-through, or raw mem0 export (`{results:[…]}` / `{memories:[…]}` / bare array) |

`resolveMem0Doc` probes in order (file → Qdrant → server → embedded) with an actionable last-resort message (manual `pip`/`python` one-liner + `--file`) when nothing is found. New `--mem0-qdrant/-collection/-server/-token/-path/-user` overrides.

## Key decisions / correctness notes

- **No `@qdrant/js-client-rest`** — it has no embedded mode and its `scroll()` is a 1:1 HTTP wrapper; a hand-rolled `fetch` loop is dependency-free and equivalent (consistent with Lore's no-dep HTTP style).
- **On-store payload text key is `data`, not `memory`** (`memory` is only the SDK's *returned* key). Reader handles both.
- **Pickle shapes**: `pickleparser` puts instance attributes either directly on the object OR under `__dict__` (pydantic `__setstate__`); guarded with `obj.__dict__ ?? obj`. Non-primitive payload values (e.g. a pickled `datetime`) decode to a wrapper and are coerced away.
- **produceDoc is now async** (`LoreImportDoc | Promise<LoreImportDoc>`) since mem0 probes the network; Engram stays synchronous (existing tests updated to `await`).
- mem0 OSS has no category taxonomy → default **`pattern`**; `metadata.repo` → project; point id (or `hash`) kept as `external_id`.
- New dep: **`pickleparser@0.2.1`** (pure-TS, zero-dep, MIT) on `@loreai/core`, lazy-imported so detection-only paths stay light.

## Tests / gate

- New `packages/core/test/import/mem0.test.ts`: record mapping, Qdrant scroll pagination (mock `fetch`), server REST + auth-hint, **embedded pickle reader against baked `storage.sqlite` fixtures (no Python in CI)**, `datetime`-value coercion, resolver auto-detect/override, `--file` shapes.
- Guards **mutation-verified**: empty-content skip, content truncation, Qdrant empty-page stop.
- Full suite green (6385 passed; the only failure is the environmental `agents.test.ts` git-remote check — no `.git` remote in a jj workspace, passes in CI's real checkout). typecheck / lint / format:check clean.

## Follow-ups (not blocking)

- Optional: LLM categorization pass for mem0 memories (deferred — default `pattern` per plan).
- Optional: mem0 hosted-cloud (`api.mem0.ai`) REST with `m0-…` key (out of scope for a local-migration tool).

Refs #1398

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,7 @@ lore import --source engram --global
 
 Engram observation types map onto Lore categories (`bugfix`/`discovery` → `gotcha`, `config` → `architecture`, etc.), and each observation's project is recovered from its session directory. Idempotent — re-running dedups by title and only updates entries whose content changed.
 
-> mem0 migration is coming in a follow-up release.
+**mem0** is imported natively too — `lore import --source mem0` auto-detects your deployment (Qdrant/OpenMemory server, mem0 self-hosted server, or the embedded default store) and reads it with no Python required. Imported memories default to the `pattern` category. Use `--file <dump.json>` or the `--mem0-*` flags to point at a specific store.
 
 ### Web dashboard
 
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -31,6 +31,7 @@
     "@noble/hashes": "^2.2.0",
     "micromark": "^4.0.0",
     "p-limit": "7",
+    "pickleparser": "0.2.1",
     "remark": "^15.0.1",
     "sqlite-vec": "0.1.9",
     "uuidv7": "^1.1.0",
--- a/packages/core/src/import/index.ts
+++ b/packages/core/src/import/index.ts
@@ -59,12 +59,26 @@ export {
 } from "./structured";
 export { parseEngramExport } from "./sources/engram";
 export {
+  resolveMem0Doc,
+  parseMem0File,
+  mem0RecordsToDoc,
+  mem0RecordToEntry,
+  scrollQdrantCollection,
+  fetchMem0ServerMemories,
+  readEmbeddedStorage,
+  embeddedStorageCandidates,
+  type Mem0Record,
+  type Mem0ResolveOptions,
+} from "./sources/mem0";
+export {
   engramSource,
+  mem0Source,
   getStructuredSources,
   getStructuredSource,
   detectStructuredSources,
   type StructuredSource,
   type StructuredSourceName,
+  type ProduceDocOptions,
 } from "./structured-sources";
 
 // Register built-in providers on first import.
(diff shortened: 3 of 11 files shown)


BYK:
## Validated against a real mem0 2.0.12 store — found & fixed a real bug

Stood up an actual mem0 2.0.12 embedded store (real library, real Qdrant `storage.sqlite` + pickle format) and ran the real `lore import --source mem0` CLI end-to-end:

- dry-run → 3 found; import → **3 created**; re-run → **3 skipped** (idempotent). Verified all 3 landed with correct title/content/`pattern` category/project.

**Bug the live run caught (fixtures missed it):** real mem0 stores the `points.id` **column** as `base64(pickle(uuid))`, not the plain id — the real UUID lives inside the decoded PointStruct's `__dict__.id`. The reader was passing the column through verbatim, so `external_id` came out as a garbage pickled blob. Fixed `readEmbeddedStorage` to read `state.id` (fallback to the column). Added a **regression test using the real captured blob + real id column**, mutation-verified.

## Adversarial review: MERGE — both flagged test gaps now closed

Review verdict was MERGE (5/7 mutants killed; 2 survivors were test-coverage gaps, not bugs). Both closed with mutation-verified tests:
1. **undecodable-blob swallow** — a corrupt point no longer aborts the read; good rows still returned.
2. **mem0-server resolver branch** — missing-`--mem0-user` throw, `serverUrl || userId` false-trigger guard, and the successful server-via-resolver path are all now tested.

## Gate
- Rebased onto latest main (#1416). Full suite: **6390 passed** (only failure is the environmental `agents.test.ts` git-remote check — no `.git` remote in a jj workspace; passes in CI). typecheck / lint / format:check clean. +5 tests vs the first push.

PR #1418, opened 2026-07-21
fix(gateway): reasoning-headroom floor for OpenAI-protocol workers (no explicit effort)

BYK:
## Problem (found post-deploy of #1413)

After deploying #1413, the retry-on-length fix **works** — production logs show:
```
worker empty response was a budget truncation (finish_reason=length) — retrying once with max_tokens 1035 → 4140 …
```
…but sessions **still degraded** (`lore-distill failed (no-response)`). Root cause: distillation/curation workers pass **tiny raw budgets** (`workerTokenBudget` → 1–8K) with **no reasoning effort**, and OpenRouter routes reasoning models (`anthropic/claude-sonnet-5`) that **reason regardless** — burning the whole budget on hidden reasoning. A single 4× retry from ~1K (→4140) still isn't enough headroom.

The #1413 headroom only applied when an **explicit** `reasoning_effort` was set. A reasoning-on-by-default model reached with **no effort** over the OpenAI protocol got no floor. (The Anthropic/Vertex builders don't need one — they suppress thinking via `thinking:{type:"disabled"}`; the OpenAI protocol has no such lever.)

## Fix

- New `workerReasoningHeadroomFloor(model, effort)`:
  - explicit effort → `anthropicThinkingBudget(effort) + THINKING_OUTPUT_HEADROOM`
  - else if reasoning-on-by-default (`workerThinkingOnByDefault` — models.dev `reasoning_options` toggle, Claude fallback) → `DEFAULT_REASONING_MODEL_BUDGET(8192) + HEADROOM(8192) = 16384`
  - else `0` (non-reasoning model keeps its budget)
- Applied at **loop entry**, gated on `target.protocol === "openai"`, clamped to the model's output limit. Flooring the **loop** `maxTokens` (not just the builder) means the length-retry multiplies from the effective floored baseline (16384→64000) instead of the tiny raw budget (1035→4140).
- Retry simplified back to a clean multiply from the (now effective) `maxTokens`.

## Tests

4 new/updated, all **mutation-verified**:
- reasoning model + small raw budget + no effort → floored to 16384, `reasoning_effort` NOT sent (drop floor → RED)
- NON-reasoning model → keeps raw 1035 (unconditional floor → RED)
- retry multiplies from floored baseline → 64000 (drop loop floor → RED)
- no-shrink guard test uses a non-reasoning id so it isolates the ceiling clamp

Full suite green (6374 passed); typecheck + lint + format clean.

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -37,6 +37,9 @@
 <!-- lore:019ef387-4ae0-7676-b7b5-140c85ad2627 -->
 * **knowledge\_meta table: confidence/last\_reinforced\_at moved off knowledge table (v55)**: Migration v55 moves \`confidence\` and \`last\_reinforced\_at\` off the \`knowledge\` table into a new convergent register table: \`knowledge\_meta (logical\_id TEXT PRIMARY KEY, confidence REAL NOT NULL DEFAULT 1.0, last\_reinforced\_at INTEGER, updated\_at INTEGER NOT NULL DEFAULT 0)\`. \`knowledge\_current\` view exposes both via LEFT JOIN: \`SELECT k.\*, COALESCE(m.confidence, 1.0) AS confidence, m.last\_reinforced\_at FROM knowledge k LEFT JOIN knowledge\_meta m ON m.logical\_id = k.logical\_id WHERE k.is\_current=1 AND k.is\_deleted=0\`. Reading/writing confidence on the base \`knowledge\` table now errors with 'no such column'. For v1 rows: logical\_id == id. \`ltm.create()\` creates matching knowledge\_meta row automatically. Schema version bumped 54→55. \`markInjected()\` updates \`knowledge\_meta.last\_reinforced\_at\` but does NOT bump \`updated\_at\` — injection stays sync-silent. Content-only \`update()\` also must NOT bump \`meta.updated\_at\`.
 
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial review addressed in c6807ce7 (no blockers were found; both SHOULD-FIX + the NIT are fixed):

- **SHOULD-FIX #2 (gemini gap)**: extended the reasoning-headroom floor to the native `gemini` protocol — Gemini 2.5 reasons by default and counts thinking against `maxOutputTokens`, and `buildGeminiWorkerRequest` has no thinking-disable lever, so it's the same class as OpenAI. Gate is now `protocol === "openai" || protocol === "gemini"`, with a mutation-verified gemini floor test.
- **SHOULD-FIX #1 (untested gate, M3 survived)**: added a test asserting an anthropic reasoning worker gets loop floor 0 (raw budget passed through since the anthropic builder suppresses thinking). Dropping the gate now → RED.
- **NIT #1**: corrected the no-shrink test comment to match the shipped guard.

**Accepted residual (not changed):** `DEFAULT_REASONING_MODEL_BUDGET` stays 8192 (floor 16384). A heavy reasoner may still truncate attempt 1 and recover via the retry-to-64000; raising the default would over-provision every worker call. The codex-responses path sends no output budget (ChatGPT rejects it) so it correctly needs no floor.

Full suite green (6376 passed); typecheck + lint + format clean.

PR #1422, opened 2026-07-21
feat(knowledge): let the curator re-title an entry whose scope broadened

BYK:
## What

D1b of the Modem *"[how coding agents read your code](https://modem.dev/blog/how-coding-agents-read-your-code)"* plan (discoverability = the words you write). #1414 taught the curator to **write** discoverable titles on `create`; #1416 fixed the non-LLM `pattern-extract` seed path. This closes the last gap: a **merged/broadened entry could never be re-titled**.

The curator `update` op had no `title` field, so consolidation kept a survivor's original narrow title even after folding in other entries' scope. Since a title is the top-weighted search key (BM25 `title` 6.0 vs `content` 2.0) *and* the dedup identity, a stale narrow title actively hides the merged knowledge.

## Change

Adds an optional `title` to `ltm.update()` and the curator `update` op.

- **Versioning:** a title change (like a content change) appends a new immutable version sharing the same `logical_id`; FTS re-indexes (old title stops matching, new one starts) and the entry is re-embedded. Every ref/anchor/meta/injection table keys on `logical_id`, so they survive a re-title untouched. (`appendVersion` already supported title overrides — this wires it to the update path.)
- **Collision guard (`titleCollides`)** — the load-bearing safety piece. Re-title is the **only** mutation that changes the dedup key *without* going through `create()`'s dedup guard, and there is **no DB unique constraint** on `(project, LOWER(title))`. So a careless re-title could silently create two live entries with the same title in one scope. `titleCollides()` mirrors `create()`'s scope logic (project + cross pool, or global/cross pool), excludes the entry's own `logical_id`, and **drops** a colliding re-title (keeping the entry's prior valid title + still applying any content change). Dropping is safe; a silent duplicate is the exact state dedup exists to prevent.
- **Curator plumbing:** `title` is type-guarded (non-string ignored), trimmed, and length-capped at `MAX_ENTRY_TITLE_LENGTH`. `changedEntries` reports the *effective* (post-guard) title. The `CURATOR_SYSTEM` update-op schema + both consolidation prompts now instruct re-titling a survivor **only** when its scope genuinely broadened.

## Safety assessment (verified before implementing)

Re-titling is structurally safe at storage/sync/FTS/refs layers — title lives on the immutable per-version row, sync coalesces versions by `logical_id` to one remote row (content-hash-aware, no divergence), and no cross-reference table is title-keyed. The single real hazard is the dedup collision, addressed by the code guard above.

## Tests

- `ltm-retitle.test.ts`: version append on title-only change, FTS reindex (new title matches / old doesn't), **collision-drop (no duplicate)**, case-only re-title of self is not a collision, content+title together, idempotent re-title (no version growth).
- `curator-retitle.test.ts`: `title` threads through `applyOps`, collision dropped via the curator path, `update` without `title` is backward-compatible.

Collision guard **mutation-verified**: neutering `titleCollides` fails both collision tests. Full `@loreai/core` suite green (154 files / 3176 tests). Typecheck, format, lint clean. Mutation experiments run in throwaway `/tmp` copies with the real tree verified unchanged.

## Not in this PR
- D2c (entry↔file association), D3 (blog post) — remaining plan follow-ups.


BYK:
Adversarial review pass (no BLOCKERs). Addressed the findings:

- **SHOULD-FIX 1 — `titleCollides()` promoted-entry scope gap (silent duplicate):** a promoted entry (project_id set, `cross_project=1`) fell into the global/cross branch, which omitted `project_id = ?` — so re-titling a promoted entry onto a plain same-project sibling's exact title wasn't detected → two live entries sharing `LOWER(title)` in the home project. Rewrote the guard to build the scope pool additively: any entry with a non-null project_id now includes its **own project** plus the cross/global pool (covers plain, promoted, and global entries in one query). Added a regression test — 'a PROMOTED entry re-titled onto a same-project sibling's title is DROPPED' — **mutation-verified**: reverting to the old promoted branch fails exactly that test.
- **NIT 1 — redundant sync push on a dropped title-only re-title:** the `idsToSync` gate used the *requested* title; a dropped collision still enqueued a no-op push. Now gated on the **effective** title change (`after.title !== entry.title`), read back after `ltm.update`.
- **NIT 2 — consolidation prompt churn:** guidance is conservative ('ONLY when scope genuinely broadened') and idempotent re-titles are no-ops; left as-is.

Full `@loreai/core` suite green (155 files / 3210 tests) after a `pnpm install` — the 5 mem0-connector failures I initially saw were a local missing-`pickleparser` dep from #1417, not this PR (my branch touches no import/mem0 files). Format/typecheck clean.

PR #1424, opened 2026-07-21
feat(provenance): capture tool_calls.input_path (file-provenance foundation)

BYK:
## What

**D2c PR-1** of the Modem *"[how coding agents read your code](https://modem.dev/blog/how-coding-agents-read-your-code)"* plan — this is #627 **Step 1**, the foundation for associating knowledge entries with the source files that produced them.

This is a **different** signal from #1414's code anchors. #1414 surfaces the file:line that a knowledge entry's *prose cites*. This records which files a session *actually touched* — session→file **provenance** — so a later PR can attribute an entry to the files behind it (and, optionally, flag it stale when they change).

**No behavior change:** the column is written but not yet read by anything. This PR is deliberately self-contained and reviewable on its own.

## Change

- **Migration v75:** adds nullable `tool_calls.input_path`, plus the matching column-recovery in `recoverMissingObjects` (a `CREATE TABLE IF NOT EXISTS` can't add a column to a pre-existing table — same recovery pattern as the `verifier` column).
- **`extractFilePath(input: unknown)`** in `tool-trace.ts` (beside its sibling `extractCommand`): pulls the path from the common file-tool shapes (`{path}` / `{filePath}` / `{file}` — the keys Read/Edit/Write/read_file/edit_file/write_to_file use across hosts), a JSON string of the same, and a plain-text path fallback. Returns `undefined` for bash/grep/task/etc. Provenance-only — best-effort, never load-bearing, so a miss is always safe.
- **`recordToolCalls`** populates it in the seed INSERT via `input_path = COALESCE(tool_calls.input_path, excluded.input_path)`, so a re-seed (retry / stream re-delivery) never clobbers a captured path. `COLS_PER_ROW` 11→12; chunk recomputed to 75 rows (75×12 = 900, still under the portable ~900-param budget).

## Tests

`tool-file-provenance.test.ts`:
- `extractFilePath` units: object (`{path}`/`{filePath}`/`{file}` + precedence), JSON string, plain-text fallback, non-file inputs → `undefined`, empty-path → `undefined`, non-string/null/number.
- `recordToolCalls` integration: path stored for `read`/`edit`, NULL for `bash`/`grep`; a re-seed with no path keeps the first captured path (COALESCE).

Population **mutation-verified** (neutering `extractFilePath` fails both integration tests). Bumped the three "current schema version" assertions 74→75 (expected for a new migration). Full `@loreai/core` suite green (156 files / 3213 tests); typecheck/format/lint clean.

## Next (D2c PR-2/3, not in this PR)
- **PR-2:** `knowledge_file_refs` table + curator "files touched this session" context block + create-path population + recall rendering. (Open design decision: heuristic all-touched vs curator-selected subset.)
- **PR-3 (optional):** git-diff `gitHead..HEAD` file-change staleness → confidence decay (#627 Step 3).

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -387,6 +387,9 @@
 <!-- lore:019f83d2-e6cb-753d-8d08-8622859ecfa8 -->
 * **Explicit investigation/research requests are read-only — no file modification**: When Burak Yigit Kaya explicitly frames a task as investigation/research (e.g. "do NOT modify any files"), treat it as strictly read-only: answer with file:line evidence, don't make edits even if a fix seems obvious. This is distinct from the adversarial PR-review workflow \[\[019f8241-1d89-7a92-b3bc-32e34741c308]] — it applies to any ad-hoc "investigate X and tell me" request, seen recurring across sessions (requested-investigation pattern, 10 prior sessions).
 
+<!-- lore:019f8457-b131-7aaa-9397-f7f1ef853bd6 -->
+* **Follow strict TDD/verification/PR workflow for every feature implementation**: When implementing a feature (e.g., new DB column, migration, extraction helper), the user expects the assistant to: (1) research existing code/conventions before writing new code; (2) write tests matching existing test-file harness conventions (helpers, describe blocks); (3) run the specific new tests, then run full package test suite, lint, format:check, and typecheck; (4) perform mutation testing in a throwaway copy/worktree to prove tests are non-vacuous, verifying the real repo file hash is unchanged; (5) fix any incidental failures uncovered (e.g., stale hardcoded schema-version assertions) by updating them to match the new state; (6) update the relevant plan file in quality/plans; (7) create a dedicated feature branch off a clean/up-to-date main, commit with conventional commit messages, push, and open a PR; (8) initiate an adversarial subagent review of the PR before merging. Never skip verification steps or merge without review.
+
 <!-- lore:019f6fde-efd1-7f32-82db-76eb9454a990 -->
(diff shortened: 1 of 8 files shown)


BYK:
Adversarial review pass — one BLOCKER found and fixed:

**B1 (BLOCKER) — ReDoS in the plaintext fallback of `extractFilePath`.** The reviewer correctly flagged that `PATH_FALLBACK_RE`'s nested quantifiers backtrack super-linearly on a long slash-run with no matching extension, and that this runs synchronously inside a SQLite savepoint on the hot path — reachable when a tool call's arguments fail to JSON-parse and the host stores the raw string in `state.input`. Notably, the mitigation the reviewer suggested (mirror gradient.ts' `indexOf('/')` skip + 64KB cap) is **insufficient** — I verified a crafted 64KB `a/a/a/…` blob is one pathological token entirely within the cap and still stalls ~8s (gradient.ts has the same latent exposure). 

Fix: after the cheap `indexOf('/')` skip and a tighter scan cap, **split on whitespace and only run the backtracking regex on individual short tokens** (`length <= 260`). A pathological no-whitespace blob is one giant token that exceeds the cap and is skipped, so the regex never sees adversarial input; a real path is a short whitespace-bounded token and still matches. Worst case now ~1ms. Added a ReDoS regression test (50K-segment blob must return `undefined` in <1s) — **mutation-verified**: reverting to the naive `input.match` fails it at 19s.

**S1 (SHOULD-FIX) — vacuous COALESCE-direction test.** The re-seed test only proved 'NULL doesn't clobber' (passed even with the COALESCE reversed). Hardened it to re-seed with a *different non-null* path (`src/second.ts`) and assert the first (`src/first.ts`) still wins — **mutation-verified**: reversing the COALESCE now fails it.

Everything else in the review was confirmed correct (migration/recovery pattern, param counts, `?? null` binding, tool_calls not in SYNCED_TABLES → no sync ripple, schema-version bumps complete). Full `@loreai/core` suite green (156 files / 3215 tests); typecheck/format/lint clean.

PR #1426, opened 2026-07-21
fix(gateway): reasoning floor keys off any reasoning capability, not toggle-only

BYK:
## Problem (found in production AFTER #1418 deploy)

After deploying #1418, the curator **still** truncated:
```
worker empty response (HTTP 200) … finish_reason=length native_finish_reason=max_tokens
  model=openrouter/anthropic/claude-sonnet-5 worker=lore-curator
worker empty response was a budget truncation — retrying once with max_tokens 2048 → 8192
```
The retry recovered it (no `no-response`, no degradation), but the **first attempt was 2048 (unfloored)** — the #1418 reasoning-headroom floor never fired for this model. If it had, the first attempt would have been 16384 and the retry 64000.

## Root cause

The floor gated on `workerThinkingOnByDefault`, which returns true **only** when `reasoning_options` contains a `toggle` type. models.dev lists OpenRouter's `anthropic/claude-sonnet-5` as:
```json
"reasoning_options": [{ "type": "effort", "values": ["low","medium","high","xhigh","max"] }]
```
No `toggle` → `workerThinkingOnByDefault` returns false; and because the array is non-empty, the Claude-id fallback is never reached. The model reasons by default in practice (it burned the whole budget) but was classified as non-reasoning, so the floor was skipped.

`workerThinkingOnByDefault` is correctly **narrow** for its own job ("must we send `thinking:{type:disabled}`?" — only `toggle` is on-by-default on the Anthropic wire). The budget floor needs a **broader** question: "can this model reason at all by default?" — which is also true for `effort`- and `budget_tokens`-typed models.

## Fix

- New `workerModelReasons(model)`: true for **any** non-empty `reasoning_options`, else the Claude-id fallback (offline-safe).
- `workerReasoningHeadroomFloor` now gates on `workerModelReasons` instead of `workerThinkingOnByDefault`. The two predicates deliberately diverge on `effort`-typed models: not thinking-on-by-default (no opt-out param needed) but DOES reason (needs the budget floor).

## Tests (mutation-verified)

- **Regression**: an `effort`-typed reasoning model with **no toggle** and a **non-"claude" id** gets floored to 16384. Reverting the gate to `workerThinkingOnByDefault` → raw 2048 → **RED** (reproduces the exact production bug).
- `workerModelReasons` unit test: toggle/effort/budget_tokens → true; none → false; explicit divergence from `workerThinkingOnByDefault` on effort-typed; offline Claude-id fallback.

Full suite green (6427 passed); typecheck + lint + format clean.

## Note on deploy verification
This was caught precisely because I verified the logs after your restart rather than trusting the merge. The retry backstop masked it as a recovered warning, but the first-attempt budget was wrong. This closes that gap.

### DIFF
--- a/.lore.md
+++ b/.lore.md
@@ -170,6 +170,9 @@
 <!-- lore:019ef389-a501-7db4-a808-4e0c671bb2f0 -->
 * **effectiveMetaThreshold: internal Date.now() makes tests flaky — inject nowMs parameter**: Trap: \`effectiveMetaThreshold(lastTurnAtMs)\` calling \`Date.now()\` internally looks correct — it needs the current time to compute \`gapMs\`. Fix: any test that captures \`const now = Date.now()\` then passes \`now - (DEEP\_IDLE\_MS - 1)\` as \`lastTurnAtMs\` is flaky — a ≥1ms scheduling delay pushes \`gapMs ≥ DEEP\_IDLE\_MS\`, flipping the return value. Fix (PR #926): add injectable \`nowMs\` parameter; tests pass a fixed \`NOW\` constant. Sentinel: \`lastTurnAtMs === 0\` means 'never set (first turn)' → treated as deep-idle (gapMs = Infinity). All 8 \`effectiveMetaThreshold\` tests rewritten to use fixed \`NOW\` — fully deterministic.
 
+<!-- lore:019f8468-759c-7969-9344-d6be41f6a9b4 -->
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial review complete — **MERGE, no blockers, no SHOULD-FIX**. Confirmed:

- `workerThinkingOnByDefault` is unchanged and still gates the Anthropic-wire `effectiveDisableThinking` (a mutation swapping it to `workerModelReasons` → RED via the existing effort-only opus test), so the thinking-suppression path is untouched.
- The broadened floor cannot cause a new attempt-1 truncation — it's still clamped to `workerLengthRetryCeiling` (min of model output limit, 64000) and applied via `max(raw, floor)`.
- The effort-typed regression test uses a non-"claude" id so only the `reasoning_options` signal is in play; reverting the gate → RED at 2048, reproducing the exact production curator truncation.

On the review's NITs:
- **NIT-2 (gemini floor branch untested)** — already covered: the `applies the reasoning floor on the native GEMINI protocol` test (added in #1418) goes RED when `|| target.protocol === "gemini"` is dropped (mutation-verified just now). No action needed.
- **NIT-1 (shared `reasoningOptions` lookup helper)** — cosmetic; skipping to avoid churn. The two predicates are intentionally distinct (toggle-only for thinking-suppression vs any-reasoning for budget headroom) and each is directly tested.
- **NIT-3** — `workerModelReasons` is test-only-exported with a single production caller; acceptable.

**Accepted residual:** an offline + non-"claude" + no-models.dev reasoning model still gets floor=0 (no signal available); the #1413 length-retry backstop recovers it on attempt 2.

Full suite green (6427 passed); typecheck + lint + format clean.

Issue #1429, opened 2026-07-21
Flaky test: anthropic-client-openai-upstream-stream 'bad port' fetch under full-suite parallel load


BYK:
## Symptom

`packages/gateway/test/anthropic-client-openai-upstream-stream.test.ts > … > re-emits a buffered OpenAI upstream as an Anthropic SSE stream (not JSON)` fails intermittently under the **full parallel suite** (`pnpm test`), but **passes reliably in isolation** (`pnpm exec vitest run packages/gateway/test/anthropic-client-openai-upstream-stream.test.ts` → 1 passed).

Observed 2026-07-21 during an unrelated one-line change (worker reasoning budget constant); re-run and isolated run both green, confirming it is a pre-existing flake, not a regression.

## Error

```
TypeError: fetch failed
Caused by: Error: bad port
 ❯ globalThis.fetch packages/core/test/setup.ts:80:10
 ❯ packages/gateway/test/anthropic-client-openai-upstream-stream.test.ts:154:24
```

## Root cause (likely)

The test builds `baseURL = \`http://127.0.0.1:${server.port}\`` (line ~133) from `await startServer(config)` and immediately `fetch`es `${baseURL}/v1/messages` (line 154). `Error: bad port` means `server.port` was falsy/invalid (e.g. `0` or `undefined`) at fetch time — i.e. `startServer` resolved before the listener's port was actually bound/populated. `setup.ts:80` is an innocent passthrough to `realFetch`; the bad URL originates in the test.

Under full-suite parallel load (many workers, `resetPipelineState()` + DB open/close churn in `beforeEach`), the bind either races or the returned handle's `port` isn't yet set when the test reads it. In isolation there's no contention, so the port is always ready → passes.

## Suggested fix

Make `startServer` not resolve until the socket is actually listening and `port` is populated (await the `listening` event / `server.address()`), OR have the test assert `server.port` is a valid non-zero port before constructing `baseURL`. A retry/backoff around the first fetch would mask it but not fix the race.

## Notes
- Not related to the ephemeral-invite reaper flake (#1397).
- Non-blocking: does not reproduce in isolation or on re-run; CI generally passes.

PR #1430, opened 2026-07-21
fix(gateway): raise default reasoning budget 8192 → 16384 (land curator on attempt 1)

BYK:
## Context

Follow-up to #1426 (which fixed the effort-typed misclassification so the reasoning-headroom floor actually fires). After #1426 deployed, I verified production logs: the floor works, but the value was still too small for a heavy `claude-sonnet-5` curator pass.

Production logs (2026-07-21, session `1eMRchBV7Ajs0dds`), one truncation in ~11 min:
```
worker empty response … finish_reason=length model=openrouter/anthropic/claude-sonnet-5 worker=lore-curator
worker empty response was a budget truncation — retrying once with max_tokens 16384 → 64000
```
It recovered via the #1413 retry (no `no-response`, no degradation) — **correct** behavior — but it means the curator burned a wasted first call + retry latency because the floored 16384 (8192 reasoning + 8192 answer) wasn't enough reasoning room.

## Change

Raise `DEFAULT_REASONING_MODEL_BUDGET` **8192 → 16384** (= `anthropicThinkingBudget("high")`). The no-effort reasoning floor becomes **16384 + 8192 = 24576**, so the common heavy curator/distill pass lands in **one round-trip**. The length-retry (to the model's full 64000 output limit) stays as the backstop for the rare pass that still exceeds it.

"A floor, never a charge" — a higher ceiling costs nothing for models that emit fewer tokens; it only prevents premature `finish_reason=length` truncation.

## Tests

Updated all no-effort floor assertions to 24576 (openai builder, effort-typed model, gemini protocol, retry-from-floored, retries-once). The explicit-`effort=high` path is unchanged (already 24576). **Mutation-verified**: reverting the constant to 8192 turns 5 floor tests RED.

Full suite green (6428 passed); typecheck + lint + format clean.

## Note
A pre-existing full-suite flake (#1429, `startServer` port race in `anthropic-client-openai-upstream-stream.test.ts`) is unrelated — passes in isolation and on re-run.


BYK:
Adversarial review complete — **MERGE, no blockers, no SHOULD-FIX**. Confirmed:

- All 5 changed `16384 → 24576` assertions are backed by models with `limit.output = 64000` (≥ 24576), so the clamp `min(24576, ceiling)` emits 24576 — not a clamped-down value.
- `anthropicThinkingBudget("high") === 16384` — the constant's comment is accurate.
- No small-output-reasoner regression: the floor is clamped to `min(24576, model output limit)` and applied via `max(raw, floor)`, so it never exceeds a model's limit nor shrinks a larger raw budget. Unknown/fallback models (output 8192) clamp to 8192 exactly as before.
- Mutation-verified: reverting the constant to 8192 → exactly the 5 updated tests RED (checked in a throwaway worktree; main tree byte-identical after).
- No other test file references the constant or asserts 24576.

Addressed the one NIT (stale mutation-hint comment in the effort=high test) in 4625e9eb.

**Accepted residual:** an offline + non-"claude" + no-models.dev reasoning model still gets floor=0 (pre-existing, out of scope); the #1413 length-retry recovers it.

Issue #1403, opened 2026-07-20
#961 Workstream A (PR2): decision + negative-control recall scenarios, de-justified memory grader


BYK:
Part of #961. Plan Workstream A / PR2. PR1 (objective signal engine) shipped as #1185.

## Deliverable
1. Add a `decision` + `negative-control` scenario set (extend the `multi-session-recall.ts` MSR pattern) with explicit ground-truth "should-recall" and "must-not-resurface" events.
2. Memory-recall scoring WITHOUT a justifier: deterministic checks first (required-fact present? stale/superseded value surfaced? = deterministic fail); only a constrained rubric where unavoidable, and the grader must NEVER see or synthesize from the retrieved context — only judge presence/absence of ground-truth facts.
3. Replace the leniency-biased `JUDGE_SYSTEM` framing (`judge.ts:329-333` tells the grader the reference is a floor and extra content scores 5 — the exact leniency #961 forbids) for memory questions.

## Acceptance
- Ground-truth-anchored deterministic scorer for recall/negative-control.
- No grader path that reads the retrieved context.
- Reported separately from end-task/behavioral signals.

PR #1433, opened 2026-07-21
fix(eval): decision-recall negative-control baseline must be "raw", not "no-memory"

BYK:
## What

Fixes the Seer bug prediction on #1431 ([review r3622842287](https://github.com/BYK/loreai/pull/1431#discussion_r3622842287)) — **verified real**.

`decision-recall.ts` (the DEC-1 decision-evolution scenario, #1403) listed `"no-memory"` in `APPLICABLE_BASELINES`, but `"no-memory"` is **not** a member of the `BaselineMode` union (`types.ts:31`). The no-memory control is `"raw"` — *"full conversation, no memory layer"* (`baselines.ts:264`). The harness intersects a scenario's `applicableBaselines` against the configured baselines via `.includes(b)` (`harness.ts:886`), so `"no-memory"` was silently filtered out and the negative-control arm — the scenario's entire reason for existing (measure a *"memory does nothing"* baseline) — **never ran**.

## Why it wasn't caught

`packages/core/eval` is not in the core `tsconfig.json` `include` (`["src","test","script"]`), so `BaselineMode[]` type-checking never covered the eval scenarios — a `"no-memory"` literal in a `BaselineMode[]` array compiles clean. Widening eval typecheck coverage surfaces ~60 pre-existing errors (ImportMeta, `bun:sqlite`, etc.), so that's tracked separately and out of scope here.

The existing test was also **vacuous** on this: `expect(scenario.applicableBaselines).toContain("no-memory")` only checked the string was *present*, so it passed while the arm silently didn't execute.

## Change

- `decision-recall.ts`: `"no-memory"` → `"raw"` in `APPLICABLE_BASELINES`; doc comment clarified (`` `raw` (full history, no memory layer) ``).
- `decision-recall.test.ts`: now asserts `"raw"` is present **and** that *every* listed baseline is a real `BaselineMode` (`ALL_BASELINES` membership) — a regression guard against any future invalid baseline the harness would silently drop.

**Mutation-verified:** forcing `"no-memory"` back (`as BaselineMode`) fails the new membership assertion. Full `@loreai/core` eval suite green (3 files / 32 tests); format clean.

### DIFF
--- a/packages/core/eval/decision-recall.test.ts
+++ b/packages/core/eval/decision-recall.test.ts
@@ -1,16 +1,26 @@
 import { describe, expect, test } from "vitest";
 import { scoreRetrieval } from "./recall-score";
 import { scenarios } from "./scenarios/decision-recall";
+import { ALL_BASELINES } from "./types";
 
 const scenario = scenarios[0];
 const byId = new Map(scenario.questions.map((q) => [q.id, q]));
 
 describe("DEC-1 decision-recall scenario (#1403)", () => {
-  test("registers one scenario with 3 evolving sessions and the no-memory control arm", () => {
+  test("registers one scenario with 3 evolving sessions and the raw no-memory control arm", () => {
     expect(scenario.id).toBe("dec-1-decision-evolution");
     expect(scenario.sessions).toHaveLength(3);
-    expect(scenario.applicableBaselines).toContain("no-memory");
+    // The negative-control arm must be `raw` (full history, no memory layer):
+    // "no-memory" is NOT a valid BaselineMode, so the harness would silently
+    // filter it out and the control would never run (regression guard for the
+    // Seer-flagged bug in #1431).
+    expect(scenario.applicableBaselines).toContain("raw");
     expect(scenario.applicableBaselines).toContain("lore");
+    // Every listed baseline must be a REAL BaselineMode — else harness.ts's
+    // `.includes(b)` intersection silently drops it and the arm never executes.
+    for (const b of scenario.applicableBaselines) {
+      expect(ALL_BASELINES).toContain(b);
+    }
   });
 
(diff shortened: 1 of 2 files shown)


BYK:
Adversarial review passed — **safe to merge**, no BLOCKER/SHOULD-FIX. One accuracy NIT worth correcting in the record:

**Precision on runtime impact.** The harness runs the *intersection* `config.baselines.filter(b => scenario.applicableBaselines.includes(b))` (harness.ts:885-887), and the default configured baselines are `["lore","compaction"]` (run.ts:111). So `raw` is **selectable** but not run by default — it executes only when the caller passes it (e.g. `--baselines ...,raw`). The precise before/after: previously, *requesting* `raw` still dropped the arm (the scenario listed the phantom `"no-memory"`); now, requesting it finally executes the negative control. The bug was real (the arm was **unreachable**, even on request); calling it "never ran → now runs" slightly overstates the default behavior.

**Verified correct by review:** `raw` is the right control (only no-memory `BaselineMode`; `rawBaseline` = full transcript, no memory layer — the right "memory does nothing" arm for a 3-session decision-evolution scenario). The hardened test's `ALL_BASELINES` membership loop is the assertion that guards the bug *class* (invalid literal mixed with valid ones) — mutation-verified RED both ways. No shared `APPLICABLE_BASELINES` symbol with other scenarios; zero other `no-memory` refs.

**Follow-up (out of scope):** `packages/core/tsconfig.json` includes only `["src","test","script"]` — `eval/` is never typechecked in CI (`pnpm -r run typecheck`), which is exactly why an invalid `BaselineMode` literal reached runtime as a silent filter-drop instead of failing the build. Filing an issue to wire `packages/*/eval` into a CI typecheck step so this bug class fails at compile time.

PR #1435, opened 2026-07-21
fix(gateway): true-streaming passthrough for codex/openai-responses main turn

BYK:
## Problem

GPT-5.6 / any ChatGPT-backend (codex) model was **completely unusable**: the OpenCode client showed \"waiting for response headers\" and retried every ~10s forever.

**Root cause** (`pipeline.ts:8530`): the codex/\`openai-responses\` main-turn path fully **buffered the entire upstream SSE response** via \`accumulateResponsesSSEStream\` *before* constructing the client \`Response\`/headers, with **no keepalive**. For a codex turn — where ChatGPT reasoning generation is slow-to-first-token — the client sees zero bytes and no headers for the full generation window, so its header timeout fires → retry loop.

Only the Anthropic path streamed live (event-by-event) with a 30s keepalive ping. `moonshotai/kimi-k3` worked in the same session because it takes the Anthropic streaming path.

This is a **pre-existing architectural gap** (last real change to the codex stream path was #1322), not caused by the recent worker PRs (#1407/#1413/#1418/#1426/#1430). The buffering exists so the recall-tool interception loop can rewrite the response — but that only matters when a recall tool is actually in the request.

## Fix

Add `streamResponsesPassthrough()` in `stream/openai-responses.ts`: forwards each upstream Responses SSE event to the client **verbatim as it arrives** while accumulating a complete `GatewayResponse` in parallel (so `postResponse` cost/calibration/temporal runs identically), with:
- 30s SSE-comment (`: keepalive`) keepalive (Responses API has no first-class `ping`),
- `response.failed` emitted on upstream error (client never hangs on a missing terminal event) + `onComplete` still fires,
- upstream-reader cancel on client disconnect.

**Gated fast path** in `pipeline.ts`: use passthrough only when
- upstream AND client are both `openai-responses`,
- **no recall tool present** (recall interception requires buffering so an injected `tool_use` never leaks — the property is known up-front via `modifiedReq.tools`), and
- no warning to inject.

Otherwise the existing buffered `accumulate + finalizeWithRecall` path is unchanged.

Refactored the buffered accumulator into shared per-event helpers (`makeResponsesAccState`/`applyResponsesEvent`/`finalizeResponsesAcc`) so both paths derive an identical `GatewayResponse` from one implementation (no drift).

## Tests

New `streamResponsesPassthrough` block in `openai-responses-stream.test.ts`:
- **true streaming** — client reads `response.created`/deltas *before* the upstream emits its terminal event (mutation-verified: withholding the in-loop forward makes this test **time out at 30s** — the exact hang it guards against);
- **verbatim field preservation** — `reasoning_summary_text.delta` (not accumulated) still reaches the client byte-for-byte;
- **upstream error** → `response.failed` emitted + `onComplete` still fires;
- **client cancel** → upstream reader cancelled.

All existing accumulator tests pass unchanged (buffered path now delegates to shared helpers). Recall-codex-stream (buffered+recall path) unaffected.

- typecheck: clean
- lint: clean (only pre-existing unrelated warnings)
- full suite: 334 files passed | 15 skipped; 6438 tests passed | 226 skipped

## Follow-up (separate PR, per sequencing)

Worker cross-model misroute for ChatGPT-backend sessions (`gpt-5.4-mini` → `chatgpt.com/backend-api` 404) + `gpt-5.6-luna` worker preference for real `api.openai.com` sessions. Independent bug; contained (no auth poisoning, no main-turn coupling).


BYK:
## Adversarial review addressed (verdict was MERGE WITH FIXES — no blockers)

Review confirmed: core anti-hang fix correct, gate airtight against recall leaks (M1: dropping `!hasRecallTool` → recall-codex tests RED), refactor behavior-identical, no ReadableStream lock races.

Both SHOULD-FIX items fixed in `dc6c2773`:

- **S2 — untyped/`[DONE]` frames no longer forwarded.** `parseSSEStream` synthesizes `event: message` for untyped `data:` lines and yields the `[DONE]` sentinel; the passthrough forwarded both verbatim, which the old buffered path never did. Now only real Responses events reach the client; untyped/`[DONE]` frames are still consumed for accumulation. New test `does not forward untyped 'message' frames or the [DONE] sentinel` — mutation-verified (revert filter → RED).
- **S1 — vacuous verbatim test fixed.** Now asserts the wire form `event: <type>\ndata:` instead of a bare substring (which also matched the type inside the JSON payload). Mutation-verified (drop `event:` line in `reserializeSSE` → RED).
- **N1** — added `onComplete`-exactly-once assertion on the error path.

NITs N2 (postResponse on cancel — matches old buffered codex behavior, not a regression), N3 (CRLF — pre-existing `parseSSEStream` quirk, ChatGPT is LF), N4 (bounded extra keepalive tick) acknowledged, no action needed.

Green: typecheck, lint, full suite (334 files, 6439 tests passed | 226 skipped).

PR #1437, opened 2026-07-21
fix(gateway): ChatGPT-backend workers reuse session model; prefer gpt-5.6-luna for real openai

BYK:
Two independent worker-model selection fixes for OpenAI-provider sessions (queued after the codex hang fix #1435).

Both hinge on a fact `config.ts` encodes but `getWorkerModel` couldn't see: a **real `api.openai.com` API-key session** and a **ChatGPT-backend (Codex/OAuth-JWT) session** BOTH report `providerID="openai"` and `protocol="openai-responses"` — only the upstream URL host differs (`api.openai.com` vs `chatgpt.com/backend-api`).

## Part 1 — correctness (the production 404)

A ChatGPT-backend session's endpoint serves **ONLY the session's own model**. `WORKER_DEFAULTS["openai"]` family selection picked a cheaper sibling (`gpt-5.4-mini`) and sent it to `chatgpt.com/backend-api` → **404 Not Found**, degrading background work (observed: session `1cvqnuwLYFtVqXL5`, `worker model: newest in family openai/gpt-mini → gpt-5.4-mini` then 404). The cross-provider guard (`pipeline.ts:2531`) doesn't fire because both sides are `providerID="openai"`.

**Fix:** thread the session's upstream `url` into `getWorkerModel` (widened optional param — every real caller already passes the full `UpstreamSnapshot`; the no-arg `cost-tracker.ts` call stays safe) and add `isChatGPTBackend(url)` (host `chatgpt.com` or path `/backend-api`, `URL`-parsed with a scheme-less substring fallback). When it matches, the worker **reuses the session's own model**, skipping family/cost-aware selection.

- `providerID "openai-codex"` is **EXEMPT** — that endpoint legitimately serves the cheaper `gpt-5.1-codex-mini` for codex sessions (validated downgrade preserved).
- Reuse overrides `cfg.workerModel` (the endpoint physically can't serve another model); `LORE_WORKER_MODEL` remains the global escape hatch.
- Extends the standing invariant: a worker must never send a model an endpoint can't serve — now covers same-provider single-model endpoints, not just cross-provider.

## Part 2 — preference (gpt-5.6-luna)

For real `api.openai.com` sessions, prefer **`gpt-5.6-luna`** ($1/M in — ~5× cheaper than the sol tier $5/M, higher quality than the mini tier $0.75/M).

Data-driven: `WORKER_DEFAULTS["openai"]` family changed `gpt-mini` → `gpt-luna`, so `resolveNewestInFamily` tracks the newest luna generation from models.dev (`modelID: gpt-5.6-luna` is only the offline fallback). `alreadyCheap` widened to `luna|mini|nano` so an already-cheap session isn't downgraded.

> Note: the general "closest cheaper tier" (#1407) logic would pick `gpt-5.6-terra` ($2.5) as the highest tier still cheaper than sol; luna is an explicit preference (per directive) since distillation/curation don't need terra-level quality.

## Tests + verification

7 new tests (`worker-model.test.ts`): ChatGPT-backend reuse; `/backend-api` path detection on any host; real-openai luna preference (over terra AND mini); `openai-codex` exemption → codex-mini; already-cheap ChatGPT-backend reuse; already-luna real-openai no-downgrade; no-url fail-safe.

Mutations **M1–M5 all RED** (remove short-circuit; remove codex exemption; revert family to gpt-mini; `isChatGPTBackend` always-false; always-true).

- typecheck: clean
- lint: clean (only pre-existing unrelated warnings)
- full suite: 6448 passed | 226 skipped


BYK:
## Adversarial review addressed (verdict was MERGE — no blockers)

Review confirmed both fixes correct for the production paths; `isChatGPTBackend` robust against adversarial URLs (15-case truth table); mutations M1/M3/M4 non-vacuous. Also rebased onto latest main to clear a `.lore.md`-only merge conflict that was blocking CI.

Addressed:

- **S1 (SHOULD-FIX) — short-circuit now reuses the LIVE session model, not `effectiveModelID`.** The ChatGPT backend serves ONLY the session's live model, so a `.lore.json` `cfg.model` override naming a different model would also 404. Changed the reuse to `{ providerID: sessionProviderID, modelID: sessionModelID }` and keyed the `openai-codex` exemption on `sessionProviderID` (not `effectiveProvider`) — this also closes the focus-5 edge where `cfg.model.providerID="openai"` could mask a real codex session.
- **Vacuous test replaced.** The old "already-cheap luna reuses its own model" passed even with the short-circuit removed (luna is `alreadyCheap` → same result via fallback). Replaced with "ChatGPT-backend session on a pricier tier (terra) reuses it, never downgrading to luna" — **mutation-verified**: removing the short-circuit makes it RED (`expected 'gpt-5.6-luna' to be 'gpt-5.6-terra'`). Terra is NOT `alreadyCheap`, so the fallback can't coincidentally produce it.

NITs N1 (path-substring host check — benign, fails safe to reuse) and N2 (stale `sessionUpstream` type on the compaction path — structural typing keeps it working) acknowledged; N2 is a reasonable pre-existing cleanup but out of scope here.

Green after rebase + fix: typecheck clean, full suite re-running.

PR #1449, opened 2026-07-22
fix(build): embed only the target platform's native assets per binary

BYK:
## Problem

The 0.38.0 release build aborted in postject during the SEA blob inject — twice, on the linux (ELF) targets:

```
Injecting blob into node executable: ...lore-linux-arm64
Aborted()
✗ fossilize failed: RuntimeError: Aborted().
    at abort (.../postject@1.0.0-alpha.6/dist/api.js:399)
```

## Root cause

The build packed **every** platform's native onnxruntime + vec0 libraries into **every** binary — fossilize embeds one shared asset manifest into all targets — so each SEA blob was **~305 MB**. postject@1.0.0-alpha.6's ELF injection aborts once the blob passes ~300 MB (a memory ceiling in its 32-bit-WASM/LIEF ELF section rebuild). The onnxruntime-node 1.21→1.27 bump (#1444) grew every platform's lib and pushed the combined blob over the edge. Mach-O injection tolerated the large blob, which is why only the linux (ELF) targets aborted.

Reproduced deterministically in isolation: injecting the 305 MB all-platforms blob into the linux-arm64 ELF aborts; a 163 MB per-platform blob injects cleanly.

## Fix

Embed only the running platform's native assets in each binary (the shared embedding model + workers still ride along in all of them), invoking fossilize **once per target** with a per-target asset manifest. Fossilize wipes its outDir per run, so each target builds into its own subdir and the binary is moved up afterward.

Result:
- Each SEA blob drops from ~305 MB to ~163 MB — well under the ELF limit.
- Every shipped binary is ~140 MB smaller (darwin 424 MB → 301 MB; linux ~123 MB → ~285-288 MB).
- Each fossilize run embeds 11 assets (14 for windows) instead of 23.

### Relationship to #1447

#1447 serialized the injects on a shared-heap-race theory. That was a **red herring** — a single serialized inject still aborts, as this PR's investigation showed — but it's retained since it lowers peak memory. This is the actual fix.

## Verification

Local 4-platform `--release` build: all four platforms build with **no abort**. The linux-x64 binary passes:
- `--version` → `0.37.0`
- `--check-embeddings` → `ok dim=768` (ONNX model loads + embeds)
- `--check-vec` → `ok vec_version=v0.1.9`, `worker vec_available=true`

confirming per-platform native asset extraction still works at runtime. `typecheck`, `lint`, `format:check` all clean.

### DIFF
--- a/packages/gateway/script/build-binary-sea.ts
+++ b/packages/gateway/script/build-binary-sea.ts
@@ -498,6 +498,11 @@ async function buildBinary() {
   }
 
   mkdirSync(distBinDir, { recursive: true });
+  // Wipe staging first so a stale artifact from a previous build (e.g. an
+  // old all-platforms ort-manifest.json) can never be picked up and mask a
+  // regression. Everything below regenerates the staging dir from scratch.
+  // (The --from-staging path returns earlier and never reaches here.)
+  rmSync(stagingDir, { recursive: true, force: true });
   mkdirSync(stagingDir, { recursive: true });
 
   // -------------------------------------------------------------------------
--- a/packages/gateway/script/build-binary-sea.ts
+++ b/packages/gateway/script/build-binary-sea.ts
@@ -35,6 +35,7 @@ import {
   mkdirSync,
   readFileSync,
   renameSync,
+  rmSync,
   unlinkSync,
   writeFileSync,
 } from "node:fs";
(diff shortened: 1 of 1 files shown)


BYK:
Adversarial review caught a real bug in the first commit: fossilize keys every asset by its `file` field and reads the bytes from `<manifestDir>/<file>` — it **ignores `src`**. The aliased `{ file: "ort-manifest.json", src: "ort-manifest-<target>.json" }` therefore made fossilize look for `<staging>/ort-manifest.json`, which this change had stopped writing, so a clean build would abort at the fossilize step (ENOENT on the asset).

It passed local verification only because the build never wipes `.sea-staging`, so a **stale** `ort-manifest.json` from the previous all-platforms code was still on disk and got picked up. A clean CI checkout has no such file — exactly the passes-locally/fails-in-CI trap.

Fixed in `0c9942b4`: persist each target's content as `ort-manifest-<target>.json` and, inside `runFossilize` (fossilize runs serially, once per target), copy the right one onto the shared `ort-manifest.json` filename right before that target's build.

Re-verified with a **wiped `.sea-staging`** (true CI equivalent): all 4 platforms build with no abort; `lore-linux-x64 --check-embeddings` → `ok dim=768` and `--check-vec` → native vec0 loads in main + worker. typecheck/lint/format all clean.

PR #1452, opened 2026-07-22
feat(import): authenticate extraction from each harness's own on-disk credentials

BYK:
## Problem (#1398)

Standalone `lore import` could only authenticate knowledge extraction with `LORE_WORKER_API_KEY` (assumed a raw Anthropic/OpenAI key) and hardcoded the extraction model to `anthropic`/`claude-sonnet-4-6`. A GitHub Copilot subscriber (no raw key) hit repeated `protocol-mismatch` failures: their Copilot token was routed to Anthropic.

## Fix

`lore import` now authenticates extraction by reading **each harness's own on-disk credentials** — no running gateway, no `LORE_WORKER_API_KEY` required.

- **New `AgentAuthProvider` capability** (core), parallel to the existing `AgentHistoryProvider` registry. Each harness owns "read my history" *and* "read my credentials".
- **Per-harness readers**: OpenCode `auth.json` (api-key + oauth), Claude Code `.credentials.json` (oauth bearer), Codex `auth.json` (raw `OPENAI_API_KEY`), Pi `auth.json`. Missing/empty/garbage files return `[]` (never throw).
- **Per-agent credential chain** in `commandImport`: `LORE_WORKER_API_KEY` override → harness on-disk auth (directly-authenticatable + unexpired) → `cfg.model` → accurate failure guidance. Resolved per detected agent, so OpenCode imports with OpenCode's creds, Codex with Codex's, etc.
- **Shared `defaultModelForProvider()`** built on the existing `WORKER_DEFAULTS` table replaces five hardcoded `{anthropic, claude-sonnet-4-6}` literals (`import.ts`, `import-auto.ts`, `api.ts` ×3), so extraction uses the validated per-provider default matching the resolved credential.
- **Rewritten no-auth guidance** that names the real reason and recommends re-auth / `lore run` first, with `LORE_WORKER_API_KEY` as a labeled fallback.

A credential is only usable if its provider has a concrete upstream URL and doesn't need a runtime token exchange — so `github-copilot` (GitHub→Copilot token exchange) and `url:null` local runtimes (ollama, vllm, …) fall through to guidance rather than a doomed call.

## v1 boundaries (follow-ups)

- Expired OAuth tokens are **skipped** (no refresh — the harness owns that).
- Copilot standalone token exchange: `readAuth()` returns `[]`.
- No live-gateway credential borrowing.

## Tests

- Core: per-harness reader behavior, expiry filter (injected clock), garbage/empty → `[]`.
- Gateway: `defaultModelForProvider` table; `resolveAgentImportAuth` precedence (worker-key > harness > cfg.model), routability filter (skips unroutable / null-url / github-copilot / prefers routable).
- Mutation-verified non-vacuous: expiry filter, routability filter (tight + loose), `defaultModelForProvider` branch — each confirmed RED then reverted.
- Gate: `typecheck`, `lint`, `format:check`, and the import/worker-model/auth suites (331 tests) all green.

Adversarially reviewed (assume-a-bug pass): verdict MERGE, no MUST-FIX. The one SHOULD-FIX it surfaced (routability filter accepting null-url / github-copilot providers, suppressing the honest guidance) is fixed in this PR with regression tests.

Fixes the credential half of #1398.


BYK:
Rebased onto main now that #1454 is merged, and carried its three worker-key fixes into the per-agent path so nothing regresses:

- the per-agent `createGatewayLLMClient` call now passes `{ dedicatedWorkerKey: !!workerApiKey }` and the `LORE_WORKER_UPSTREAM`-aware `workerUpstreams` map;
- `cfgModel` is resolved as `workerModel ?? model` (workerModel preferred for background extraction);
- the pre-flight guard (fail loudly when a non-`sk-ant-` worker key would hit Anthropic with no upstream) is applied once before the per-agent loop.

All of #1454's `import-command-local` tests pass unchanged against the per-agent implementation. Full gate green locally (typecheck, format:check, 24 test files across the import/worker-model/auth suites). Driving CI to green now.

BYK:
Reopened as #1455 — GitHub wedged this PR's head on a stale SHA after a force-push and wouldn't let it reopen. Same branch (`fix/import-harness-ondisk-auth`), same work, now correctly tracking HEAD. Continuing there.

PR #1454, opened 2026-07-23
fix(import): plumb dedicatedWorkerKey and workerUpstream to LLM client

JordanTreDaniel:
## Problem

Standalone `lore import` had three bugs blocking extraction when a dedicated worker key is configured:

1. **Missing `dedicatedWorkerKey` flag** — the LLM client was created without `{ dedicatedWorkerKey: true }`, so the protocol-mismatch check in `llm-adapter.ts:1862-1863` always fired even when `LORE_WORKER_API_KEY` was set. `pipeline.ts` passes this correctly (line 2502).

2. **`LORE_WORKER_UPSTREAM` ignored** — the import always used `config.upstreamAnthropic` instead of `config.workerUpstream`. `pipeline.ts` handles this at lines 2485-2487.

3. **`workerModel` config ignored** — the hardcoded `claude-sonnet-4-6` default (Anthropic protocol) always applied even when the user configured a `workerModel` with a different provider in `.lore.json`.

## Fix

1. Pass `{ dedicatedWorkerKey: !!workerApiKey }` to `createGatewayLLMClient` so the mismatch check is skipped when a dedicated worker key is configured.
2. Use `config.workerUpstream` when set, falling back to the default upstreams.
3. Honor `cfg.workerModel` before falling back to `cfg.model` or the hardcoded claude-sonnet-4-6 default.

## Testing

With an OpenAI key and upstream:
```bash
export LORE_WORKER_API_KEY=sk-...
export LORE_WORKER_UPSTREAM=https://api.openai.com/v1
lore import --project ~/dev/my-project --agent opencode -y
```

Or configure `.lore.json`:
```json
{
  "workerModel": { "providerID": "openai", "modelID": "gpt-4o-mini" }
}
```

Previously: all chunks fail with `protocol-mismatch`, 0 entries created.
With fix: extraction proceeds normally.

### DIFF
--- a/packages/gateway/src/cli/import.ts
+++ b/packages/gateway/src/cli/import.ts
@@ -668,8 +668,9 @@ export async function commandImport(
   // built-in default. Track whether the provider was *explicitly* configured so
   // the no-credential message below doesn't tell a Copilot/OpenRouter user to
   // set an "anthropic key" they don't have.
-  const modelExplicit = cfg.model != null;
-  const defaultModel = cfg.model ?? {
+  // Prefer `workerModel` over `model` since imports are background work.
+  const modelExplicit = cfg.model != null || cfg.workerModel != null;
+  const defaultModel = cfg.workerModel ?? cfg.model ?? {
     providerID: "anthropic",
     modelID: "claude-sonnet-4-6",
   };
@@ -721,10 +722,15 @@ export async function commandImport(
     return;
   }
 
+  const workerUpstreams = config.workerUpstream
+    ? { anthropic: config.workerUpstream, openai: config.workerUpstream }
+    : { anthropic: config.upstreamAnthropic, openai: config.upstreamOpenAI };
+
   const llm = createGatewayLLMClient(
-    { anthropic: config.upstreamAnthropic, openai: config.upstreamOpenAI },
+    workerUpstreams,
     getImportAuth,
     defaultModel,
+    { dedicatedWorkerKey: !!workerApiKey },
   );
 
   try {
--- a/packages/gateway/src/cli/import.ts
+++ b/packages/gateway/src/cli/import.ts
@@ -670,10 +670,11 @@ export async function commandImport(
   // set an "anthropic key" they don't have.
   // Prefer `workerModel` over `model` since imports are background work.
   const modelExplicit = cfg.model != null || cfg.workerModel != null;
-  const defaultModel = cfg.workerModel ?? cfg.model ?? {
-    providerID: "anthropic",
-    modelID: "claude-sonnet-4-6",
-  };
+  const defaultModel = cfg.workerModel ??
+    cfg.model ?? {
+      providerID: "anthropic",
+      modelID: "claude-sonnet-4-6",
+    };
 
   // Extraction needs a usable credential. `lore import` is a standalone CLI
   // process: it never proxies a conversation turn, so no client credential is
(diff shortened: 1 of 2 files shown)


BYK:
Thank you for this, @JordanTreDaniel — really appreciate you digging into the worker-key path and sending a focused fix. 🙏

I did a close review. The three changes are all correct and genuinely useful:

- `dedicatedWorkerKey: !!workerApiKey` — correctly scoped; it only skips the early protocol-mismatch prefix check (llm-adapter.ts:1861), and leaves auth resolution and the cross-provider fail-closed route guard intact. This matches how `pipeline.ts` already calls it.
- `LORE_WORKER_UPSTREAM` → `workerUpstreams` — good; note it only redirects `anthropic`/`openai` worker providers (resolveTarget only consults the upstreams map for those two; everything else routes via `PROVIDER_ROUTES`). Worth a one-line comment so it's not surprising.
- preferring `cfg.workerModel` — correct, and matches the "imports are background work" intent.

Two things before merge:

1. **The headline Copilot case still won't work as written** (this is the one blocker). `workerKeyScheme("github-copilot")` returns `api-key` → the token goes out as `x-api-key`, but `api.githubcopilot.com` needs `Authorization: Bearer` + a `Copilot-Integration-Id` header + a `copilot_internal/v2/token` exchange the gateway doesn't do standalone. So a Copilot subscriber setting `LORE_WORKER_API_KEY=<copilot-token>` gets the error *relabeled* (protocol-mismatch → 401), not resolved. The fix is real and valuable for anyone with a **raw** non-anthropic key (OpenAI/OpenRouter/etc.) + `LORE_WORKER_UPSTREAM` — could we just reword the PR/commit to say Copilot import still needs a non-Copilot worker key? (Copilot token-exchange is a tracked follow-up.)

2. **A small regression to guard against**: with the early mismatch check now off, `LORE_WORKER_API_KEY=<non-anthropic-key>` and *no* `LORE_WORKER_UPSTREAM`/`workerModel` will make N doomed `api.anthropic.com` calls + N `Sentry.captureException`s (one per chunk) where before it rejected pre-flight with zero network/Sentry. Worth keeping an early skip for the `api.anthropic.com`-target + non-`sk-ant-` key case, or suppressing Sentry on the session-less import auth path.

Heads-up on sequencing: there's overlap with #1452 (harness on-disk auth), which rewrites this same block into per-agent resolution. Plan is to land your fix first, then I'll rebase #1452 on top and carry these three fixes into the per-agent path (with attribution). Thanks again — this was a sharp catch.

BYK:
Pushed a follow-up commit (8408f67) to address the two review points directly rather than asking you to — hope that's alright! Your three fixes are untouched; I only added on top:

1. **Pre-flight guard for the regression I flagged** — since `dedicatedWorkerKey` now disables the in-adapter mismatch check, a non-`sk-ant-` `LORE_WORKER_API_KEY` with no `LORE_WORKER_UPSTREAM` would default to the Anthropic upstream and fire one doomed request + one `Sentry.captureException` per chunk. Added a single pre-flight check that catches exactly that case and fails loudly with actionable guidance (point at the right provider, or use `lore run`), and notes that a Copilot/ChatGPT subscription token isn't a usable raw key.

2. **Tests** (the plumbing shipped with none) — asserts the LLM client is built with `dedicatedWorkerKey` + the `workerUpstreams` map + the `workerModel`-preferred default model; that `LORE_WORKER_UPSTREAM` redirects both upstreams; and that the new guard fires (and doesn't over-fire when an upstream *is* set). Mutation-verified (disabling the guard turns the test red).

Full gate green locally: typecheck, format:check, and the import/worker-model/auth suites (151 tests). Once CI is green I'll merge this, then rebase #1452 on top and carry these three fixes into the per-agent path. Thanks again for the fix. 🙏

PR #1457, opened 2026-07-23
feat(import): use github-copilot on-disk token directly for extraction

BYK:
## Problem (#1398)

`lore import` skipped OpenCode's `github-copilot` credential, on the assumption it needed a `copilot_internal/v2/token` exchange we can't do from a stored key. That's why Kjaer's Copilot-only setup couldn't import.

## What's actually true (verified live against a real Copilot account)

OpenCode stores the long-lived GitHub OAuth token (`gho_...`) and uses it **directly** as `Authorization: Bearer` against `api.githubcopilot.com` — there is **no token exchange** (the `copilot_internal/v2/token` endpoint 404s with this token; OpenCode never calls it). A `chat/completions` request with just the stored token returns `200`.

## Fix

- **`opencode.ts` reader**: for `github-copilot`, emit the bearer from `refresh` (the field OpenCode itself uses), and treat `expires: 0` as a "no known expiry" sentinel rather than epoch-0-expired (so a valid token isn't filtered out). The sentinel handling is safe for all oauth providers — a genuinely-expired token carries a positive past epoch.
- **`import.ts`**: drop the `github-copilot` exclusion from the routability filter — it IS routable (concrete `api.githubcopilot.com` url + usable stored bearer).
- **`copilotHeaders()`**: add `Copilot-Integration-Id` + `X-GitHub-Api-Version` on worker requests to githubcopilot hosts (matches real Copilot clients; the token alone also authenticates). No-op for other hosts.

## Tests

Reader (refresh preferred, `expires:0` sentinel), routability filter (github-copilot now resolves), and worker-request headers (Bearer not x-api-key, integration headers present / absent for non-copilot). Mutation-verified: reverting each of the four changes turns its regression test red. A live end-to-end check (real creds → 200) was run locally and removed before commit (it hits the network / needs creds).

### DIFF
--- a/packages/core/src/import/auth/opencode.ts
+++ b/packages/core/src/import/auth/opencode.ts
@@ -5,7 +5,13 @@
  * (`$XDG_DATA_HOME/opencode/auth.json`, default `~/.local/share/opencode`).
  * Each entry is keyed by the models.dev provider ID and is one of:
  *   - `{ type: "api", key }`                       → api-key credential
- *   - `{ type: "oauth", access, refresh, expires }` → bearer credential (access token)
+ *   - `{ type: "oauth", access, refresh, expires }` → bearer credential
+ *
+ * For most oauth providers the short-lived `access` token is the bearer. For
+ * `github-copilot`, OpenCode uses the long-lived GitHub OAuth `refresh` token
+ * (`gho_...`) directly as the bearer against api.githubcopilot.com (no
+ * `copilot_internal/v2/token` exchange), and stores `expires: 0` as a
+ * "no known expiry" sentinel — handled below.
  *
  * We map each entry to an AgentResolvedAuth. OpenCode uses models.dev provider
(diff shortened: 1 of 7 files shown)


BYK:
Adversarial review: **MERGE**, no MUST-FIX.

Verified live against a real Copilot account — OpenCode's stored `gho_` token works directly as `Authorization: Bearer` against `api.githubcopilot.com` (chat/completions → 200); no `copilot_internal/v2/token` exchange (that endpoint 404s). Routing traced: `github-copilot` → protocol `openai` → `api.githubcopilot.com/chat/completions` (no `/v1`), bearer via `authHeaders` (never x-api-key, which 400s). All 4 change points mutation-verified (revert → RED). Token never logged or sent to Sentry.

Added the two SHOULD-FIX edge tests from review (refresh-absent → access fallback, both-absent → omitted).

PR #1459, opened 2026-07-23
fix(import): use chat-completions-capable Copilot worker model (gpt-5-mini)

BYK:
## Problem (#1398 follow-up)

After #1457 made `lore import` authenticate off the GitHub Copilot on-disk token, @Kjaer's import got past auth but still failed on every chunk:

```
400 Bad Request — url=https://api.githubcopilot.com model=github-copilot/gpt-5.4-mini cred=bearer worker=lore-import
{"error":{"message":"model \"gpt-5.4-mini\" is not accessible via the /chat/completions endpoint","code":"unsupported_api_for_model"}}
```

## Root cause

The Copilot worker default is `gpt-5.4-mini`, which Copilot serves **only** on `/responses` (+ ws), per the live models list (`supported_endpoints: ["/responses", "ws:/responses"]`). Background workers always use the Chat Completions path (`buildOpenAIChatCompletionsUrl`), so that model is unreachable → 400.

Live-verified against a real Copilot account:
| model | `/chat/completions` |
|---|---|
| `gpt-5.4-mini` | **400** (responses-only) |
| `gpt-5-mini` | **200** |
| `claude-sonnet-4.6` | **200** |

## Fix

Switch the non-claude Copilot worker default (`WORKER_DEFAULTS["github-copilot"]`, `_GENERAL_FALLBACK_WORKER`, and `resolveGitHubCopilotWorker`'s non-claude branch) from `gpt-5.4-mini` → `gpt-5-mini` — a cheap Copilot model served on `/chat/completions`. The claude branch (`claude-sonnet-4.6`) already routes correctly. `gpt-5-mini` still matches `alreadyCheap` (contains "mini"), so cost logic is unchanged. This also fixes the same latent 400 on the **live** worker path for Copilot sessions, not just import.

## Tests

`getWorkerModel` for an expensive non-claude / claude Copilot session, and `defaultModelForProvider("github-copilot")`. Mutation-verified: reverting to `gpt-5.4-mini` turns both red.

### DIFF
--- a/packages/gateway/src/worker-model.ts
+++ b/packages/gateway/src/worker-model.ts
@@ -1047,7 +1047,10 @@ const WORKER_DEFAULTS: Record<
   // "claude-sonnet-4.6"), so this entry serves only as a last-resort fallback.
   "github-copilot": {
     providerID: "github-copilot",
-    modelID: "gpt-5.4-mini", // default; overridden by resolveGitHubCopilotWorker
+    // gpt-5-mini (NOT gpt-5.4-mini): only /chat/completions-served Copilot
+    // models work for background workers. gpt-5.4-mini is /responses-only on
+    // Copilot and 400s on the worker's Chat Completions path (live-verified).
+    modelID: "gpt-5-mini", // default; overridden by resolveGitHubCopilotWorker
     alreadyCheap: (id) =>
       id.includes("mini") ||
       id.includes("nano") ||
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial review: **MERGE** (was MERGE-WITH-FIXES; the one SHOULD-FIX is now addressed in this PR).

Review confirmed the headline import fix is correct and mutation-verified (revert to `gpt-5.4-mini` → both default tests RED). It also caught a latent gap on the **live** worker path: a *cheap* Copilot session whose own model is `/responses`-only (e.g. `gpt-5.4-mini` at $0.75, or `gpt-5.5`/`gpt-5.6-*`) skipped the expensive-branch remap and echoed the session model verbatim → same 400 on `/chat/completions`.

Fixed here: the Copilot worker remap now fires **regardless of session-model cost** — for Copilot the pick is about routability (chat-completions vs responses), not price (all Copilot models cost the same to the user). Added a regression test for the cheap-responses-only case; mutation-verified (re-gating the remap on cost turns it RED).

Verified: `gpt-5-mini` + `claude-sonnet-4.6` both live-200 on `/chat/completions`. Full gate green (typecheck, lint, format, worker-model/default-model/import-command-local/worker-cross-provider suites).

PR #1460, opened 2026-07-23
feat(binpatch): chain discovery behind a pluggable SourceStrategy

BYK:
PR2 of the [binpatch extraction plan](../blob/main/.opencode/plans/binpatch-reusable-delta-update.md). Lifts delta-upgrade **chain discovery + apply orchestration** out of the gateway into the `binpatch` library, behind a pluggable `SourceStrategy`, so the engine has zero product coupling and can resolve chains from any layout. Behavior-preserving for Lore.

Follows PR1 (#1458, apply-core extraction).

## New in `packages/binpatch/`

| File | Role |
|---|---|
| `discover.ts` | `SourceStrategy` interface + `resolveAndApply()` — cache-first resolution, apply, SHA-256 verify, progress events. All product concerns injected. |
| `sources/oci.ts` | Generalized OCI anonymous-pull `OciClient` (registry/repo/user-agent injected), incl. the GHCR 307→blob-storage redirect dance. |
| `sources/ghcr.ts` | `ghcrSource()` — nightly chain resolution over `patch-<version>` manifest tags; version→target-tag scheme + semver comparator injected. |
| `sources/github-release.ts` | `githubReleaseSource()` — stable chain over Release assets; releasesUrl/binaryName/userAgent/fetch injected. |
| `contract.ts` | Wire-contract constants + shared types (`PatchChain`, `PatchLink`, `ChainStep`, `DeltaResult`). |
| `errors.ts` | Product-neutral `BinpatchError` (generalized from `UpgradeError`). |
| `events.ts` | `ProgressEvent`/`ProgressHandler` + `safeProgress`. The library **never renders** — it emits phase/bytes/done events; a throwing handler can never abort the operation. |

## Gateway becomes thin glue

`delta-upgrade.ts` now injects Lore's GHCR/Release coordinates, binary name, user-agent, `getPatchCache()`, `compareVersions`, and the Sentry `spanDeltaUpgrade` telemetry, picks the channel, and calls `resolveAndApply`. `canAttemptDelta`, channel selection, and the prefetch wrappers stay in the gateway. `ghcr.ts` keeps serving the **full-download** path (`upgrade.ts`); it only gains an exported `GHCR_REGISTRY`.

Progress migrates to the decided **pure event-hook** API (plan Open Q1): the library emits `{ type: "phase" | "bytes" | "done"; ... }`; Lore's bar and sentry-cli's spinner become thin consumers. Telemetry stays in the gateway, injected as an optional hook (plan Open Q3).

## Tests (new coverage — discovery had none before)

- `test/sources.test.ts` (16) — pure logic: `extractStableChain`, `extractSha256`, `filterAndSortChainTags`, `validateChainStep`, annotation helpers.
- `test/discover.test.ts` (8) — `resolveAndApply` orchestration with **real TRDIFF10 patches** + fake source/cache: cache-first hit, network+save, offline miss, no-chain, SHA-mismatch throw, ordered progress events, throwing-handler resilience.

**Mutation-verified:** break SHA verify → mismatch test RED; break offline gate → offline test RED; flip `filterAndSortChainTags` bound → 2 tests RED; all restored GREEN.

## Verification

- typecheck (6 packages), lint (exit 0, only pre-existing `unbound-method` warnings), format:check — all clean.
- Full suite: **343 files, 6563 tests, 0 failed** (+24 new binpatch tests).
- Real gateway `bundle` inlines binpatch's discovery with **zero** leftover `binpatch` external.

## Not in scope (later PRs)

PR3 the reusable generation composite Action; PR4 graduate to a standalone public repo + `sentry-cli` adoption.

### DIFF
--- a/packages/binpatch/README.md
+++ b/packages/binpatch/README.md
@@ -23,10 +23,11 @@ The system splits into two halves joined by a single **wire contract**:
 2. **Generation + publishing** — CI-side patch creation and upload to
    GHCR / GitHub Releases. A reusable GitHub Action (separate, later).
 
-This package currently covers the **apply core** (PR1): patch parsing, chain
-application, an offline cache, and byte-progress reporting. Chain **discovery**
-(a pluggable `SourceStrategy` over GHCR tags and GitHub Release assets) lands in
-a follow-up.
+This package covers the **apply core** (patch parsing, chain application, an
+offline cache, byte-progress reporting) and **chain discovery** — a pluggable
+`SourceStrategy` over OCI/GHCR patch-manifest tags and GitHub Release assets,
+plus a cache-first `resolveAndApply` orchestrator. The CI-side patch
+*generation* half ships separately as a composite GitHub Action (later).
 
 ## API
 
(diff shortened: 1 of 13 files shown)


BYK:
Fixed in `44954b13`. `ghcrSource.resolveChain` now wraps its whole body in a try/catch that swallows `BinpatchError` (network) → returns `null`, so a transient outage during blob download (or token/manifest/tag fetch) is reported as `unavailable` and falls back to a full download, matching the manifest-fetch behavior and the documented `SourceStrategy` contract. A programming bug still propagates. Added a regression test (stub global fetch to fail → `resolveChain` resolves to `null`), mutation-verified RED without the swallow. Note: this asymmetry pre-dated the extraction; the library boundary is the right place to fix it now that the contract documents `null` = no usable chain.

BYK:
Fixed in `c650bc8c`. Added a `reported` flag so the offline-miss path reports exactly once: `onOfflineMiss` emits `{source:'offline_miss', result:'unavailable'}`, and the `if (result === null)` branch now skips its redundant source-less report when offline-miss already fired. Note this was pre-existing behavior preserved verbatim by the extraction, and has no observable Sentry change (the span already retained `delta.source` via the `if (t.source)` guard, and distribution metrics only fire on `result === 'ok'`) — the fix just keeps the captured telemetry outcome unambiguous.

BYK:
Fixed in `0f53d6ad`. `OciClient.downloadBlob` now wraps the redirect (307 → blob storage) follow-up fetch's signal with `buildSignal(BLOB_TIMEOUT, signal)`, matching the initial request — so a stalled redirected download can never hang forever, even when the caller passes no signal (e.g. `prefetchNightlyPatches`). This was pre-existing behavior preserved verbatim by the extraction; fixed at the library boundary. Added a regression test (stub fetch → 307 → capture the redirect call's signal → assert it's a live `AbortSignal` with no caller signal), mutation-verified RED without the fix.

PR #1462, opened 2026-07-23
feat(binpatch): reusable composite Action for delta-patch generate + publish

BYK:
## Summary

PR3 of the binpatch extraction (see `.opencode/plans/1784643638774-witty-wizard.md`). Extracts the delta-patch **generate + publish** shell pipeline from `.github/workflows/ci.yml` into a reusable composite GitHub Action at `packages/binpatch/action/`, and migrates Lore's own CI to consume it.

This is the CI-side other half of the `binpatch` wire contract: the library (PR1/PR2) *applies* patch chains it discovers; this Action *produces and publishes* them.

## The Action

`packages/binpatch/action/action.yml` — a composite action with three modes, each mapping 1:1 to an existing CI job so the job graph is unchanged:

| `mode` | replaces | does |
|---|---|---|
| `generate-ghcr` | `generate-patches` | diff vs previous GHCR nightly, bsdiff, size-gate → `patches/` |
| `publish-ghcr` | `publish-nightly` | push binaries + immutable `nightly-<v>` tag + patch manifest w/ annotations |
| `generate-release` | `generate-release-patches` | diff vs previous stable Release, bsdiff, size-gate → `patches/` (Craft artifact) |

Inputs parameterize `registry` / `repo` / `artifact-type-prefix` / tag prefixes / `binary-glob` / `max-ratio` / tool versions + pinned SHAs, so a non-Lore consumer can adopt it. Outputs: `has-patches`, `from-version`.

## Wire contract preserved byte-for-byte (via defaults)

- Binaries: `<repo>:nightly` + immutable `<repo>:nightly-<version>`, `application/vnd.lore.cli.nightly`, **bare-filename** layer titles (push runs from inside the artifacts dir).
- Patches: `<repo>:patch-<version>`, `application/vnd.lore.cli.patch`, annotations `from-version=<prev>` and `sha256-<binaryName>=<hex>` (SHA-256 of the **uncompressed** binary — the sole client trust anchor).
- `bsdiff … --use-zstd` (TRDIFF10 + zstd); `MAX_RATIO=50` size gate.
- Prev-version discovery unchanged: `oras repo tags | grep '^nightly-[0-9]' | sort -V` (nightly) / `gh api releases` (stable).

Any drift in these breaks live delta upgrades — reviewed with that as the top priority.

## CI migration

Each of the three jobs now: `actions/checkout` (so the local action resolves) → `download-artifact` (unchanged) → `uses: ./packages/binpatch/action`. The `generate-patches` upload is regated from the old `env.HAS_PREV` onto the Action's `has-patches` output. ORAS/`zig-bsdiff` installs, prev-download, bsdiff, and validate shell all move into the Action.

## Verification

- `actionlint` on `ci.yml`: clean (exit 0).
- `action.yml`: valid YAML, `runs.using: composite`, 12 steps.
- typecheck / lint / format:check: clean.
- full suite: **343 files, 6568 tests, 0 failed** (CI/YAML-only change, no code impact).

## Notes

- The Action is unpublished/internal for now (staged inside the monorepo per the plan). PR4 graduates `binpatch` (lib + Action) to a standalone public repo.
- Ambient ORT WIP + `.lore.md` in the main worktree left untouched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -981,27 +981,9 @@ jobs:
     runs-on: ubuntu-latest
     continue-on-error: true
     steps:
-      - name: Install ORAS CLI
-        run: |
-          VERSION=1.3.1
-          EXPECTED_SHA256="d52c4af76ce6a3ceb8579e51fb751a43ac051cca67f965f973a0b0e897a2bb86"
-          TARBALL="oras_${VERSION}_linux_amd64.tar.gz"
-          curl --retry 3 --retry-delay 5 --retry-all-errors -sfLo "$TARBALL" \
-            "https://github.com/oras-project/oras/releases/download/v${VERSION}/${TARBALL}"
-          echo "${EXPECTED_SHA256}  ${TARBALL}" | sha256sum -c -
-          tar -xz -C /usr/local/bin oras < "$TARBALL"
-          rm "$TARBALL"
-
-      - name: Install zig-bsdiff
-        run: |
-          VERSION=0.1.19
-          EXPECTED_SHA256="9f1ac75a133ee09883ad2096a86d57791513de5fc6f262dfadee8dcee94a71b9"
-          TARBALL="zig-bsdiff-linux-x64.tar.gz"
-          curl --retry 3 --retry-delay 5 --retry-all-errors -sfLo "$TARBALL" \
-            "https://github.com/blackboardsh/zig-bsdiff/releases/download/v${VERSION}/${TARBALL}"
-          echo "${EXPECTED_SHA256}  ${TARBALL}" | sha256sum -c -
-          tar -xz -C /usr/local/bin < "$TARBALL"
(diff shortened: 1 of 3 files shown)


BYK:
Addressed the adversarial-review CONCERN (patch-push failure-isolation drift) in `a8d9c8e6`. Added a `patch-push-fatal` input (default `false`): the patch-manifest push now runs in a helper whose non-zero rc is downgraded to a `::warning::` (exit 0) by default, restoring the pre-move `continue-on-error` / 'delta upgrades are optional' semantics — binaries and the versioned tag are already published by then, so a transient patch push hiccup no longer fails `publish-nightly`. Binary-push and tag remain separate, always-fatal steps. Verified with a fake-`oras` harness (push-ok→rc0; push-fail non-fatal→rc0+warning; push-fail fatal→rc7+error); shellcheck clean (also guarded the `cd`). Thanks for catching it.

PR #1463, opened 2026-07-23
fix(run): adopt the user's existing upstream env instead of clobbering it

BYK:
## Problem

A user ran Claude Code through the Lore gateway pointed at OpenRouter:

```zsh
export ANTHROPIC_BASE_URL="https://openrouter.ai/api"
export ANTHROPIC_AUTH_TOKEN="$OPENROUTER_API_KEY"   # sk-or-... Bearer key
export ANTHROPIC_API_KEY=""
command lore
```

Two symptoms:
1. The terminal spammed background worker + import errors during the interactive session.
2. Constant authentication errors.

## Root cause

`lore run` builds the child env as `{ ...process.env, ...target.env }` — `target.env` last — so it **overwrote** the user's `ANTHROPIC_BASE_URL` with the gateway URL (intended: the agent must talk to the gateway). But the user's OpenRouter Bearer key survived, and the gateway had **no idea OpenRouter was involved** — `ANTHROPIC_BASE_URL` is a client-side var the gateway never sees. So it fell through to its hardcoded `api.anthropic.com` default and forwarded the `sk-or-` key **to Anthropic** → 401 on every foreground turn and every background worker/import call.

Separately, `lore run` starts the gateway **in-process**, and its `[lore]` log lines (`log.error`/`[worker-health] degraded`) write to the same stderr the interactive agent renders into — `silenceStderr()` (used by the Pi/OpenCode plugins) was never wired into the CLI path. So the 401-storm errors bled straight into the agent's TUI.

## Fix

**1. Adopt the user's existing upstream (all `lore run` targets).**
Before starting the gateway, capture the user's pre-existing base-URL env var for the target agent (`ANTHROPIC_BASE_URL` / `OPENAI_BASE_URL` / `GOOGLE_GEMINI_BASE_URL` / `COPILOT_API_URL`) and adopt it:
- set `LORE_UPSTREAM_<protocol>` so the in-process gateway defaults there (works for every target, even header-less ones like Gemini/Copilot);
- inject `X-Lore-Upstream-URL` (+ `X-Lore-Provider` for known hosts) for header-capable agents (Claude Code, Pi), so a known provider like OpenRouter gets its wire protocol + auth scheme flipped via the existing `providerForUpstreamOrigin` reverse-map.

An explicit user `LORE_UPSTREAM_*` still wins; a base URL already pointing at the gateway is ignored (idempotent re-runs). Lore now adopts the user's config instead of overriding it.

**2. Silence gateway stderr on an interactive `lore run`.**
Gated on `owned && !config.debug && process.stderr.isTTY`. Nothing is lost — the file sink + Sentry sink still receive everything (`lore logs`); `LORE_DEBUG=1` opts back in.

## Tests

- `captureUserUpstream`: adopt / skip-unset / self-pointing (incl. path variant) / unknown-host / opencode (no adoptable var) / gemini wire protocol.
- `run.ts` adoption wiring: sets gateway env + provider-specific env, injects headers with provider tag, respects explicit `LORE_UPSTREAM_*`, unknown host adopts URL but omits provider tag, header-less agent (Gemini) env-only.
- `providerForUpstreamOrigin`: known host → provider, `/v1` normalization, unknown/unparseable → undefined.

Mutation-verified: `captureUserUpstream`→null (7 red), skip gateway-env-set (2 red), skip provider-header (1 red).

### DIFF
--- a/packages/gateway/src/cli/agents.ts
+++ b/packages/gateway/src/cli/agents.ts
@@ -31,6 +31,98 @@ export interface AgentDef {
    * rather than environment variables — we inject `-c key=value` overrides.
    */
   cliArgs?: (gatewayUrl: string, cwd: string) => string[];
+  /**
+   * Base-URL env var(s) this agent honors to pick an upstream, in priority
+   * order (first defined wins). `lore run` rewrites these to point the agent
+   * at the gateway — but if the USER already set one (e.g. pointing Claude
+   * Code at OpenRouter or a corporate proxy), that is their intended upstream.
+   * We capture it BEFORE clobbering and tell the gateway to proxy there, so
+   * Lore adopts the user's existing config instead of silently overriding it
+   * with the hardcoded api.anthropic.com / api.openai.com default.
+   *
+   * The captured value's wire protocol is the agent's `wireProtocol` (below):
+   * an Anthropic-shape client (Claude Code) pointed at OpenRouter is adopted
+   * as an anthropic-ingress request whose upstream host is openrouter.ai; the
+   * gateway's `providerFromUpstreamUrl` reverse-map then flips protocol/scheme
+   * for known hosts. Omit for agents whose base URL cannot be adopted this way
+   * (e.g. opencode, which routes via the plugin, not a base-URL env var).
+   */
+  upstreamEnvVars?: string[];
+  /**
+   * The wire protocol the agent speaks to its upstream — used to pick which
+   * gateway `LORE_UPSTREAM_<protocol>` default to override when adopting the
+   * user's captured upstream (see `upstreamEnvVars`).
+   */
+  wireProtocol?: "anthropic" | "openai" | "gemini";
+}
+
+/**
+ * Whether a hostname is a loopback address. Under `lore run` the local gateway
+ * always binds loopback, so ANY loopback host in the user's base-URL env is
+ * either the gateway itself or a stale gateway from a previous run — never a
+ * real upstream to adopt. Rejecting all loopback hosts (not just the exact
+ * origin) keeps re-launches idempotent even when the gateway restarts on a
+ * different port (port contention). A genuine upstream (OpenRouter, a corporate
+ * proxy) is never loopback.
+ */
+function isLoopbackHost(hostname: string): boolean {
+  const h = hostname.toLowerCase().replace(/^\[|\]$/g, ""); // strip IPv6 brackets
+  if (h === "localhost" || h === "::1" || h === "0.0.0.0" || h === "::")
+    return true;
+  // 127.0.0.0/8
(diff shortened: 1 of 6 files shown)


BYK:
Pushed a follow-up (`a49c612c`):

- **Patch coverage → 81%** (was 76%). Added unit tests for `resolveAgentSelection`, `resolveLaunchTarget`, and `adoptForRemote`. The remaining uncovered added lines are the `commandRun` orchestration (gateway start, remote-mode delegation, child spawn, signal handling) — genuine integration-level code not worth a flaky spin-up-a-gateway unit test.
- **Resolved the adversarial-review MUST-FIX (CRLF header injection).** A crafted `ANTHROPIC_BASE_URL` (e.g. `https://host/\nX-Api-Key: stolen`) could smuggle a second header into `ANTHROPIC_CUSTOM_HEADERS`, since `new URL()` tolerates an embedded newline. Now: `captureUserUpstream` strips control chars and rejects any value with interior whitespace, and `appendCustomHeader` sanitizes every header name/value. Regression test added (mutation-verified).

Full gate green: typecheck, lint, tests (78), format:check.

BYK:
Addressed the Seer HIGH finding (self-pointing guard vs. port shift on restart) in `e0e990da`:

`captureUserUpstream` now rejects **any loopback host** (`127.0.0.0/8`, `localhost`, `::1`, `0.0.0.0`), not just the exact gateway origin. Under `lore run` the local gateway is always loopback, so a stale `ANTHROPIC_BASE_URL=http://127.0.0.1:<old-port>` from a previous run can no longer be adopted as an upstream (which would proxy the gateway into itself) — even if the gateway restarts on a different port. A genuine upstream (OpenRouter, corporate proxy) is never loopback.

Regression tests added (different-port + localhost/`::1`/`127.0.0.5` variants), mutation-verified. Full local gate green (typecheck, lint, 82 tests, format).

BYK:
Addressed the Seer MEDIUM finding (`maybeAutoImport` called with stale `config`) in `ba4f58ef`.

Confirmed real: `config` was captured via `loadConfig()` **before** `applyUpstreamAdoption` sets `LORE_UPSTREAM_*`, so `maybeAutoImport(config)` used the pre-adoption upstream defaults (api.anthropic.com) — the background auto-import would have sent the user's adopted key to the wrong host, the same failure this PR fixes for the foreground path. `startGateway()` re-runs `loadConfig()` after adoption, so its `handle.config` already reflects the adopted upstream.

Fix: thread `handle.config` (post-adoption) into `maybeAutoImport` via an `effectiveConfig` var. Auto-import only runs for an owned (freshly-started) gateway, so `effectiveConfig` is always the fresh post-adoption config there.

Local gate green: typecheck, lint, 92 tests (incl. import-auto), format:check.

PR #1465, opened 2026-07-23
fix(gradient): strict monotonic compression-layer ratchet — never decompress except emergency (#961)

BYK:
## Problem

While measuring cross-session cache cost (#961), the gateway cache logs showed avoidable full prefix busts from the compression **layer** oscillating turn-to-turn. `transformInner` picks the lowest layer that fits **each turn independently** (`effectiveMinLayer` resets to 0 every turn), so a session could:

- **Drop 2→1** when transient token pressure eased (layer 2 strips tool outputs, layer 1 restores them → different prefix bytes → front-bust, then re-compress next turn → bust again).
- **Re-enter layer 0** on a host-side conversation shrink (the `genuineCompaction` release in `prefixPresentFloorApplies`): the distilled prefix vanishes from `messages[0/1]` → full front-bust; next turn re-injects it → bust again (0↔N thrash).

Two partial guards existed (a `calibrated`-only sticky guard pinning to full `lastLayer`, and a `≥1` prefix-present floor), but both **released on a genuine shrink** and the floor only guaranteed `≥1` — leaving the 2→1 and shrink-to-0 gaps open.

Reclaiming window by decompressing is a **false economy**: the reclaim is itself a full cache write. A cheap cache-read on a compressed prefix beats a full re-write every time.

## Fix

Make the compression layer a **strict monotonic ratchet** on layers 1–3. A new per-session `maxSustainedLayer` high-water is floored into `effectiveMinLayer` every turn, **unconditionally** (no release on genuine compaction). Once a session compresses to layer L, it never selects below L again.

- **Layer 4 (emergency overflow) is excluded** from the high-water — it is a deliberate one-shot bust to recover from "prompt too long", not a sustained level, so it never latches the session at emergency. After recovery the session settles back to its sustained 1–3 floor. `forceMinLayer` can still push *above* the ratchet for the recovery turn.
- Replaces both the calibrated sticky guard and the `prefixPresentFloorApplies` floor (removed as now-dead) with the single high-water floor.
- `maxSustainedLayer` is re-seeded from the DB-restored `lastLayer` on cold start (no new migration): a session restored at layer L (1–3) had sustained ≥L, a safe lower bound; the ratchet re-establishes within a turn if the true high-water was higher.

## Tests

Updated the affected gradient tests to the ratchet contract (a once-compressed session holds at its layer 2→2 instead of dropping to 1/0; a genuine host-side compaction HOLDS instead of releasing to layer 0). **Mutation-verified**: disabling the ratchet floor turns 6 ratchet tests red. 228 gradient tests + 124 gateway cache/compaction tests pass; core + gateway typecheck clean.

### DIFF
--- a/packages/core/src/gradient.ts
+++ b/packages/core/src/gradient.ts
@@ -790,6 +790,22 @@ type SessionState = {
   lastTransformedCount: number;
   /** Layer used by the most recent transform() call — sticky-layer guard */
   lastLayer: SafetyLayer;
+  /**
+   * High-water compression layer (1-3) this session has ever sustained. Once a
+   * session compresses to layer L, decompressing to a LOWER layer re-renders the
+   * distilled prefix at messages[0/1] (different budget/strip level) and
+   * front-busts the whole prompt cache — then the next turn re-compresses and
+   * busts again (0<->N / 2<->1 thrash). Reclaiming window by decompressing is a
+   * false economy: the reclaim itself is a full cache write. So the layer is a
+   * strict monotonic RATCHET — `effectiveMinLayer` is floored at this every turn,
+   * unconditionally (no release on a genuine host-side context shrink).
+   *
+   * Layer 4 (emergency overflow) is EXCLUDED: it is a deliberate one-shot bust to
+   * recover from "prompt too long", not a sustained level, and must never be
+   * latched here (that would trap the session at emergency permanently). Only
+   * layers 1-3 ratchet.
+   */
+  maxSustainedLayer: SafetyLayer;
   /** Message IDs in the most recent transform() output — ID-based delta estimation */
   lastWindowMessageIDs: Set<string>;
   /** One-shot force escalation: skip layers below this on the next transform() */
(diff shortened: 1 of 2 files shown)


BYK:
Addressed an adversarial-review finding before merge (self-found, not a Seer flag).

**Cold-start reseed gap (restart-after-emergency):** the ratchet re-seeds `maxSustainedLayer` from the DB-restored `lastLayer` only when it is 1–3. If the last persisted turn was an emergency overflow (`lastLayer === 4`), the reseed was skipped → `maxSustainedLayer` stayed 0 → a small resume turn could re-enter layer 0 and front-bust the prefix for a turn. The replaced `prefixPresentFloorApplies` floor explicitly covered `lastLayer === 4`, so this was a real (self-healing, cost-only) regression versus the old code.

**Fix:** treat a restored `lastLayer === 4` as "was compressed, hold ≥ 1" (layer 4 still never latches the ratchet itself). Added a regression test (`restart after an emergency (lastLayer=4) re-seeds the ratchet to >= 1`) — mutation-verified: removing the reseed branch turns it red (`expected 0 to be >= 1`).

229 gradient tests + 124 gateway cache/compaction tests pass; core + gateway typecheck clean.

PR #1466, opened 2026-07-23
fix(import): honor env-var + explicit-model credentials for standalone import

BYK:
## Problem

The OpenRouter user from #1398 upgraded to the latest nightly (which has the on-disk-auth import + `lore run` adoption fixes) but standalone `lore import` still fails:

```
[lore] Skipping Claude Code: no usable credential found in its on-disk auth
       (none stored, expired, or provider not proxied).
[lore] Can't import: no usable credential found for background extraction.
```

His whole setup authenticates through shell env vars, not Claude Code's on-disk auth:

```zsh
export ANTHROPIC_BASE_URL="https://openrouter.ai/api"
export ANTHROPIC_AUTH_TOKEN="<openrouter key>"
export ANTHROPIC_API_KEY=""
```

That's a fully usable credential — but `resolveAgentImportAuth` never looked at it. It reads the agent's on-disk auth file, `LORE_WORKER_API_KEY`, and last-seen session creds only. So a very common OpenRouter / corporate-proxy setup falls through to "no usable credential" and gets told to run `lore run` or set a worker key, when nothing more was needed.

## Fix

Add a new credential tier (between on-disk auth and last-seen session) that reads the agent's **own base-URL + auth-token env vars**:

- New `AgentDef.authTokenEnvVars` field:
  - Claude Code: `ANTHROPIC_AUTH_TOKEN` (Bearer, wins) / `ANTHROPIC_API_KEY` (x-api-key)
  - Codex: `OPENAI_API_KEY` (Bearer)
- The captured base URL resolves the provider via `providerForUpstreamOrigin` (`openrouter.ai` → `openrouter`), and extraction is routed straight at that upstream (treated like a dedicated worker key so the in-adapter protocol-mismatch pre-flight is skipped).
- Unknown hosts still work — route at the captured upstream, default the model to the agent's wire protocol.
- A token with no base URL falls back to the wire-protocol provider default.

New shared helper `captureUserEnvCredential` reuses the same URL sanitization as `captureUserUpstream` (control-char strip + interior-whitespace reject). Guidance message refreshed to mention the env-var path.

**Precedence (unchanged for existing paths):** worker key (1) > on-disk auth (2) > **env credential (3, new)** > last-seen session (4) > `cfg.model` (5).

## Tests

9 new cases in `import-auth-chain.test.ts`: the OpenRouter setup, bearer>api-key, api-key fallback, no-base-URL fallback, unknown host, tier ordering vs on-disk auth and worker key, null when no token. Mutation-verified (disable tier 3 → 5 tests red). Full gate green (typecheck, lint, 97 tests, format).

### DIFF
--- a/packages/gateway/src/cli/agents.ts
+++ b/packages/gateway/src/cli/agents.ts
@@ -54,6 +54,18 @@ export interface AgentDef {
    * user's captured upstream (see `upstreamEnvVars`).
    */
   wireProtocol?: "anthropic" | "openai" | "gemini";
+  /**
+   * Env var(s) the agent reads for its API credential, in priority order
+   * (first defined non-empty wins). Paired with `upstreamEnvVars`: a user who
+   * points the agent at a provider purely via shell env (e.g. Claude Code at
+   * OpenRouter via `ANTHROPIC_BASE_URL` + `ANTHROPIC_AUTH_TOKEN`) has a fully
+   * usable credential Lore can reuse for standalone `lore import` — even
+   * though nothing is stored in the agent's on-disk auth file. `scheme` is how
+   * the value is sent upstream: `bearer` (Authorization: Bearer) vs `api-key`
+   * (x-api-key / Anthropic-style). Omit for agents whose credential can't be
+   * read from a static env var (e.g. opencode, github-copilot token exchange).
+   */
+  authTokenEnvVars?: { var: string; scheme: "bearer" | "api-key" }[];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial review returned **MERGE WITH FIXES**. Addressed in `b73df47b`:

**MUST-FIX (aggregator empty model)** — for OpenRouter (and any provider absent from `WORKER_DEFAULTS`: deepseek, groq, …), `defaultModelForProvider` returns an empty `modelID`, so the credential would resolve but extraction would send `model=""` → 400. The fix authenticated but couldn't import — the exact setup this PR targets.

Now tier 3 signals `needsModel` when it has the key but no resolvable model, and the caller prints targeted guidance:

```
[lore] Can't import: found your openrouter credential, but openrouter has no
[lore] built-in default worker model... Pick one and retry:
[lore]   export LORE_WORKER_MODEL=openrouter/gpt-5-mini
[lore]   lore import
```

Setting `LORE_WORKER_MODEL` flows back in as `cfgModel` and tier 3 resolves cleanly. So the OpenRouter user is one env var away, with an actionable message instead of a silent 400.

**SHOULD-FIX (untested upstream routing)** — extracted `resolveExtractionUpstreams` as a pure exported helper and unit-tested it (a review mutation showed the captured-upstream routing was uncovered).

Mutation-verified all three (disable tier 3, remove needsModel guard, ignore captured upstream → tests red). Full gate green: typecheck, lint, 199 tests, format.

BYK:
Expanded to cover Kjaer's newest failure too (`e3bb9817`).

He set `LORE_WORKER_MODEL=github-copilot/claude-sonnet-5` (his Copilot plan doesn't serve the default `gpt-5-mini`), ran `lore import --agent opencode`, and still got `400 model_not_supported` for `gpt-5-mini` — the env var was ignored.

Root cause: **tier 2 (on-disk auth) ignored `cfgModel`.** His OpenCode github-copilot token resolves via tier 2, which computed the model as `credential.modelID || defaultModelForProvider(provider)` = `gpt-5-mini`, never consulting his `LORE_WORKER_MODEL`.

Fix: tier 2 now honors an explicit `cfgModel` **when it targets the credential's own provider** — the credential authenticates the provider, the user picks the model. A cross-provider `cfgModel` is still ignored (never routes a copilot token to an anthropic model). Two new tests (same-provider honored / cross-provider ignored), mutation-verified.

So this PR now fixes both standalone-import credential gaps for the OpenRouter/Copilot user:
- env-var-only setups (tier 3, new)
- explicit `LORE_WORKER_MODEL` override on the on-disk path (tier 2)

Full gate green: typecheck, lint, 184 tests, format.

BYK:
Pushed `06fde668` addressing both Seer findings + the docs CI failure:

**Seer HIGH (`agents.ts:192`) — OpenRouter env with no `LORE_WORKER_MODEL` → empty model → 400.** Already handled by the MUST-FIX from the adversarial review: tier 3 returns `{ needsModel: providerID }` (discriminated union) when the captured provider has no `WORKER_DEFAULTS` entry and the user set no model, and the caller surfaces guidance to set a worker model — no silent empty-model 400. Hardcoding an OpenRouter default was rejected (model availability varies per account and can't be verified). Covered by the "needsModel signal" test.

**Seer MEDIUM (`import.ts:162`) — tier 3 applied `cfgModel` without a provider-match guard (unlike tier 2).** Fixed: tier 3 now honors an explicit `cfgModel` only when `cfgModel.providerID === <captured provider>`, else falls back to the captured provider's default (mirrors tier 2 exactly). A mismatched override (e.g. env → OpenRouter but `LORE_WORKER_MODEL=github-copilot/…`) is ignored rather than routing the key to the wrong model. Two new tests (mismatch ignored → `needsModel`; same-provider honored), mutation-verified.

**"Check generated docs" CI failure.** The env-doc generator scrapes JSDoc mentioning an env-var name; a comment I added for the `needsModel` branch mentioned `LORE_WORKER_MODEL` literally and got picked up as the var's description, making `environment.md` stale. Reworded the comment to reference it descriptively; `pnpm run check:docs` is green again.

Full gate green: typecheck, lint, format, 6 import-suite files pass.

BYK:
Pushed `ca9cdf32` addressing the new Seer MEDIUM (`import.ts:933`): `needsModelProvider` was a single string overwritten each loop iteration, so a multi-agent run where several agents resolve to defaultless providers would name only the LAST one in the final guidance.

Fix: collect providers into a `Set<string>` (`needsModelProviders`) across all agents; the summary now lists every affected provider and prints one `export LORE_WORKER_MODEL=<provider>/<model>` example per provider. Per-agent skip lines were already emitted for each agent; this just makes the closing summary complete. Full gate green (typecheck, lint, format, 39 import-suite tests).

BYK:
Pushed `a6d794db` addressing the new Seer MEDIUM (`import.ts:1025`): the consolidated `needsModel` guidance was gated on `!anyAuthResolved`, so in a multi-agent run where one agent authenticated and another hit `needsModel`, the per-agent skip line promised "(see below)" but the guidance was suppressed.

Fix: extracted a pure, exported `buildNeedsModelGuidance(providers, anyAuthResolved)` helper and decoupled it from `anyAuthResolved` — the guidance now appears whenever ANY agent hit `needsModel` (worded "Can't import some agents…" when others succeeded). The caller only uses `!anyAuthResolved` to decide whether to also STOP the run (so a partial success still records + summarizes the agents that worked). 4 unit tests including the partial-success regression; mutation-verified (make the helper return null when `anyAuthResolved` → the partial-success test goes RED). Full gate green (typecheck, lint, format, check:docs, import suite).

BYK:
Adversarial review found a **MUST-FIX** (now fixed in `a3352242`): the empty-model guard existed only in **tier 3** (env credential), not **tier 2** (harness on-disk auth). A routable-but-defaultless on-disk key — e.g. an OpenRouter/DeepSeek key in OpenCode's `auth.json` with no `modelID` hint and no matching `LORE_WORKER_MODEL` — would resolve to `{ modelID: "" }`, sending `body.model=""` to a routable upstream → 400 per chunk. That's the exact silent failure this PR set out to eliminate; it just wasn't closed on the on-disk path.

Fix: tier 2 now applies the same `if (!model.modelID) return { needsModel: providerID }` guard as tier 3. Tests: added a tier-2 regression ("routable-but-defaultless provider + no model → needsModel") + a same-provider-cfgModel escape-hatch test, and strengthened the pre-existing on-disk-key test to assert a non-empty modelID (it previously blessed the empty-model result). Mutation-verified (drop the guard → the new test goes RED).

Also re-ran the #1468 review delta (retry-budget fix) — verdict MERGE, no MUST-FIX, termination proven bounded.

The review's SHOULD-FIX (loop-level `commandImport` integration coverage for the needsModel/multi-agent paths) is deliberately deferred to the queued follow-up e2e PR (fake-provider harness driving the full decision matrix) — the gating logic is already unit-covered via the extracted `buildNeedsModelGuidance` helper. Full gate green: typecheck, lint, format, check:docs, 6 import suites.

PR #1468, opened 2026-07-24
fix(worker): fall back to a backup model when the preferred one is unavailable

BYK:
## Problem

When a worker's preferred model isn't available on the user's account/plan, the call 400s (`model_not_supported` / `unsupported_api_for_model`) and **every** chunk fails — the whole distillation / curation / `lore import` backlog produces nothing, with no recovery short of the user manually setting `LORE_WORKER_MODEL`.

This is the GitHub Copilot case from #1398: Copilot's model catalog varies by subscription tier, so the built-in default (`gpt-5-mini`) 400s on plans that only serve, e.g., claude. Kjaer hit exactly this.

## Fix (adapter-level, all worker kinds)

- **`WORKER_MODEL_FALLBACKS` + `workerModelCandidates(preferred)`** (`worker-model.ts`): an ordered, deduped, **same-provider** candidate list (preferred first). GitHub Copilot ships a real list — `gpt-5-mini`, `gpt-4o-mini`, `claude-sonnet-4.6`, `claude-haiku-4.5` — spanning the OpenAI + Claude families so at least one is reachable on any plan. Other providers just get `[preferred]` (no behavior change). Backups never switch provider (a cross-provider worker call is always doomed — wrong credential/wire).
- **`isModelUnsupported400(status, body)`** (`llm-adapter.ts`): detects a model-scoped 400. Deliberately narrow so a generic 400 (bad param, quota) never swaps the model.
- **Worker retry loop**: on such a 400, `markWorkerIncapable(provider, model)` (account-wide) and retry with the next candidate, rebuilding via `buildWorkerRequest` so the OAuth billing signature is recomputed. Mirrors the existing data-policy-404 blocklist+re-resolve pattern. Bounded by the finite candidate list.
- **Top-of-prompt advance**: skip PAST any already-blocklisted candidate to the first usable backup, so once one chunk blocklists the preferred model, every later chunk starts directly on the working backup instead of re-failing the swap (or being skipped).

This complements #1466 (honor `LORE_WORKER_MODEL`): that lets a user pick a model explicitly; this makes the default path self-heal when the preferred model isn't on their plan, without any config.

## Tests

`isModelUnsupported400` detector (positive/negative/non-400), `workerModelCandidates` (ordering/dedup/same-provider), and adapter behavior: recover on backup after a 400; a no-backup provider surfaces the 400 (one call); a second call after a blocklist starts on the backup with no repeated 400. Mutation-verified (disable the fallback retry, disable the advance loop → tests red). Also fixed a latent test-isolation leak in the data-policy suite that the new account-wide incapable check surfaced. Full gate green: typecheck, lint, 290 tests, format.

### DIFF
--- a/packages/gateway/src/llm-adapter.ts
+++ b/packages/gateway/src/llm-adapter.ts
@@ -73,7 +73,7 @@ import {
   recordEmptyWorkerResponse,
   clearEmptyWorkerStreak,
 } from "./worker-health";
-import { getModelEntrySync } from "./worker-model";
+import { getModelEntrySync, workerModelCandidates } from "./worker-model";
 
 // ---------------------------------------------------------------------------
 // Worker call tracking
@@ -254,6 +254,28 @@ export function isDataPolicyBlocked404(status: number, body: string): boolean {
     /data\s+policy/i.test(body) ||
     /guardrail/i.test(body) ||
     /settings\/privacy/i.test(body)
+  );
+}
+
+/**
+ * True when a 400 body says the requested MODEL is unavailable on this
+ * account/plan (as opposed to a bad request about params). Copilot returns
+ * `code:"model_not_supported"` / `"unsupported_api_for_model"`; other providers
+ * phrase it as "model not found/supported/does not exist". Kept to model-scoped
+ * markers so a generic 400 (bad param, quota, etc.) does NOT trigger a model
+ * swap. Used by the worker retry loop to fall back to a same-provider backup
+ * model (see {@link workerModelCandidates}).
+ */
+export function isModelUnsupported400(status: number, body: string): boolean {
+  if (status !== 400) return false;
+  return (
+    /model_not_supported/i.test(body) ||
+    /unsupported_api_for_model/i.test(body) ||
+    /model_not_found/i.test(body) ||
+    /\bmodel\b[^"]{0,40}?(?:is\s+)?not\s+(?:supported|found|available)\b/i.test(
+      body,
+    ) ||
+    /\bmodel\b[^"]{0,40}?does\s+not\s+exist\b/i.test(body)
   );
 }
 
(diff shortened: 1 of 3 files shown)


BYK:
Adversarial correctness + security review: **MERGE, no MUST-FIX.**

Highest-risk item — the unbounded-loop / retry-budget interaction — is **strictly bounded**: `modelFallbacks` is seeded once and only ever `.shift()`ed (never refilled), the swap branch has no `attempt < maxRetries` gate so it can neither loop forever nor abort early, and the terminating condition is `modelFallbacks.length === 0` → falls through to the generic failure. A backup that also 400s shifts to the next and terminates cleanly.

Other findings:
- **Target/protocol after swap**: PASS — candidates are always same-provider, and `buildWorkerRequest` rewrites `body.model` + re-applies `copilotHeaders` on every rebuild.
- **markWorkerIncapable scope**: PASS — ANY_WORKER is correct (account-wide availability, mirrors the data-policy-404 path); no cross-worker over-suppression since the empty-response capability path marks per-worker keys.
- **isModelUnsupported400 precision**: PASS on ordering (earlier temperature/thinking/beta/data-policy branches handle those first) and ReDoS (linear, 2MB body in ~4ms); narrowness is unit-tested (mutation to `status===400` → RED).
- **Test quality**: non-vacuous — 3 mutations (neuter fallback, neuter advance loop, drop body check) all RED.
- **Data-policy test reset**: pure isolation fix exposing a pre-existing ANY_WORKER leak the new account-wide check surfaced; no assertion changed, no masked regression.

Verified in a throwaway worktree: baseline 209 green, typecheck clean, lint exit 0, real tree untouched (sha256 match), worktree removed.

Optional non-blocking follow-ups: (1) `markWorkerIncapable`'s log line ("returned no usable text after N consecutive responses") is inaccurate for the model-not-supported path — a reason param would make it truthful; (2) an integration test locking in that a temperature/prompt-content 400 never swaps through the loop.

BYK:
Addressed the Seer MEDIUM finding (`llm-adapter.ts:2779`) in `6fc8cc0e`.

Seer was right that a model swap did `retryCount++; continue`, and the loop's `attempt++` on `continue` meant each swap silently consumed one unit of the transient-error (`maxRetries`) budget. With a full Copilot fallback chain that could leave the eventual working model with a reduced 429/5xx allowance.

Fix: a model swap now also does `attempt--`, cancelling the loop's `attempt++` so the swap is budget-neutral — model-availability retries (bounded by the finite `modelFallbacks` list) no longer borrow from the transient-retry budget. Termination is still guaranteed by the shrinking candidate list, independent of `attempt`.

New regression test "a model swap does NOT consume the transient-error retry budget": with `LORE_MAX_RETRIES=1`, preferred 400s (swap) → backup hits one 429 → backup 200. Passes only because the backup keeps its full transient allowance. Mutation-verified: revert the `attempt--` → this test goes RED. Full gate green (typecheck, lint, 274 tests, format).

Issue #1478, opened 2026-07-24
Lore-on-Pi loses buried facts across compaction: offline compaction summary carries distilled observations, not the concrete captured values


BYK:
## Summary

On agents that do their **own** client-side compaction (e.g. Pi via `@loreai/pi`), Lore does **not** get the "manage the window so native compaction never fires" benefit it gets on OpenCode. When the agent compacts, Lore's offline compaction summary becomes the *sole* carrier of the dropped history — and that summary is assembled from **distilled observations + long-term knowledge**, not from the raw captured messages. If a concrete fact was captured but the distiller didn't preserve its specific value (and it wasn't promoted to knowledge), it is silently lost at the first compaction.

Found via a single-long eval comparing OpenCode vs Pi (issue #961, Pi driver arm).

## Evidence (single-long, Sonnet-5, cap 200K, N=3)

| agent + arm | retention | compactions |
|---|---|---|
| OpenCode + Lore | **100%** | **0** |
| Pi + Lore | **0%** | 16 / 16 / 15 |

Same task, same model, same buried-aside facts (`WHOLESALE` / `EMEA` / `WH-07` / `SUBMITTED` stated once in an early turn, probed in the final turn).

Ruled out the obvious causes:
- **Not a routing/fallback bug:** gateway log shows `POST /v1/compact` called 16× and succeeding ("compact endpoint: generating summary") — no `session_not_found` 404 fallback to Pi-native compaction.
- **Not a cache bug:** `cache_bust_stats` clean on all 3 lore-Pi runs (the #1461/#1465/#1469 fixes generalize to Pi's openai-completions path).
- **Not a capture bug:** distillation + embedding ran fine (35 distillations, 18 embedded, 8 knowledge entries).

Root cause, from `lore.db` on a failing run:
- The buried values are in `temporal_messages` (1 row — the raw early turn) …
- … but in **0 of 35** distillations' `observations`/`facts`, and **0 of 8** knowledge entries (the 8 knowledge rows captured project *conventions* — "stdlib-only", `now_iso()`, warehouse API spec, soft-delete — not the order-specific values).
- `assembleOfflineCompaction` (`packages/gateway/src/compaction.ts:500`) builds the summary from: prior summary + distillation `observations` + **undistilled** raw tail (capped 4000 chars) + knowledge block. Once a message is distilled it leaves the "undistilled tail", so the only carrier for its content is the distilled `observations` — which here dropped the concrete values.

[...]

PR #1480, opened 2026-07-24
feat(pi): cancel compaction when raw context fits — keep buried facts end-to-end

BYK:
## What

Add a `shouldCancelCompaction` policy to `@loreai/pi` that returns `{ cancel: true }` from the `session_before_compact` hook when the would-be compacted context fits under a configurable threshold. Pi honors cancel on both manual (agent-session.js:1275) and auto (agent-session.js:1498) paths. Result: Lore keeps the raw context, raw facts survive end-to-end, no need for the summary path to be lossless.

## Why

The current `@loreai/pi` only returns the summary branch from `session_before_compact`, so Pi compacts natively even when Lore could keep the raw context. The summary path can drop buried facts (issue #1478). On OpenCode, `--cap-context` caps the layer-0 budget so OpenCode never natively compacts — that's why lore-OpenCode gets 100% retention with 0 compactions on single-long. This PR brings the same behavior to lore-Pi: instead of relying on a possibly-lossy summary, just **don't compact when the raw context fits**, and let Lore manage the window via recall on the next turn.

## How

- New pure policy `shouldCancelCompaction(tokensBefore, { cancelThreshold, loreActive })` — three guards:
  - `loreActive` must be true (so we don't try to cancel when Lore is disabled for this session)
  - `tokensBefore` must be finite and > 0 (unknown size falls through to the summary path)
  - `tokensBefore < cancelThreshold` (else we *must* compact — raw context would exceed the model's effective window on the next turn)
- `runCompaction` now checks the policy first; if it returns true, returns `{ cancel: true }` and never calls `/v1/compact`.
- New `DEFAULT_PI_CANCEL_THRESHOLD = 200_000` — matches the eval cap and Anthropic's 200K usable window. Configurable via `LORE_PI_CANCEL_THRESHOLD` env var.

## Tests

35 tests pass (was 26). 9 new tests cover the policy: cancel at < threshold, fall-through at >= threshold, boundary at threshold-1 vs threshold, loreActive=false guard, tokensBefore<=0/NaN guard, env var resolution, invalid env values rejected.

**Mutation-verified** all three critical guards are load-bearing:
- M1 (flip `<` to `>` in `tokensBefore < opts.cancelThreshold`): 11 tests failed
- M2 (remove `loreActive=false` guard): 1 test failed
- M3 (remove `tokensBefore <= 0` guard): 1 test failed

Existing fetch-path tests updated to use `tokensBefore: 250_000` (above threshold) so they continue to exercise the path they care about.

## Relationship to #1478

This is the first of two clean fixes for #1478 (offline compaction summary drops captured concrete values). The other fix — widening `assembleOfflineCompaction` to carry salient temporal facts — is still on the table for cases where the session grows past the cancel threshold. With this PR, the threshold becomes the dividing line: below it, cancel; above it, fall through to the summary path (which still needs the #1478 widening for the cases it gets called on).

## Eval

Eval launched at `/home/byk/.local/share/lore-eval-live/PI-XLONG-S5-CANCEL/`. Initial runs hit `403 Budget limit exceeded (monthly limit)` on the OpenRouter coding-plan key — pre-existing infra issue, unrelated to this change. Will re-run on a working key to confirm the policy actually drives retention to 100%.

### DIFF
--- a/packages/pi/src/internal.ts
+++ b/packages/pi/src/internal.ts
@@ -265,6 +265,71 @@ export function buildProviderRegistrations(opts: {
 }
 
 /**
+ * Default cancel threshold (tokens). Above this, we let Pi compact —
+ * below it, we cancel so Lore manages the window via raw context + recall.
+ * Mirrors OpenCode's `--cap-context` mechanism (Lore caps the layer-0 budget
+ * so OpenCode never natively compacts; on Pi we cap at the same threshold
+ * and cancel the native compaction so the raw context carries the buried
+ * facts into the next turn, with recall as the backup).
+ *
+ * Configurable via `LORE_PI_CANCEL_THRESHOLD` env var so tests + power users
+ * can tune it. The default 200_000 matches the eval cap and Anthropic's
+ * 200K usable window for Sonnet 5 / Sonnet 4 / Opus 4 — large enough to
+ * hold a long session's worth of raw context, small enough to never silently
+ * exceed the model's effective window.
+ */
+export const DEFAULT_PI_CANCEL_THRESHOLD = 200_000;
+
+/**
+ * Resolve the cancel threshold from env, falling back to the constant.
+ * Treats empty / non-numeric / non-positive values as "not set".
+ */
+export function resolveCancelThreshold(env: NodeJS.ProcessEnv = process.env): number {
+  const raw = env.LORE_PI_CANCEL_THRESHOLD;
+  if (typeof raw !== "string" || raw.trim() === "") return DEFAULT_PI_CANCEL_THRESHOLD;
+  const n = Number(raw);
+  if (!Number.isFinite(n) || n <= 0) return DEFAULT_PI_CANCEL_THRESHOLD;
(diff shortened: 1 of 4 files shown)


BYK:
Superseded by #1483 (merged). That PR took the gateway-driven approach you pushed for: the gateway is the authoritative source for "does this session fit in the layer-0 budget?", and the plugin is a dumb relay. PR #1480's client-side `DEFAULT_PI_CANCEL_THRESHOLD` was the wrong layer (per the review on the prior branch).

The fix shipped in #1483:
- `handleCompactEndpoint` accepts `tokens_before` and returns `{ cancel: true }` when `tokensBefore <= context - output` for the session's resolved model.
- `@loreai/pi` just sends `tokens_before` and relays `{ cancel: true }` to Pi's `session_before_compact` hook.
- 50/50 tests, 3 mutation-verified guards, adversarial review SHIP-WITH-FOLLOWUPS.

Closing this in favor of #1483.

PR #1490, opened 2026-07-25
fix(gateway): inject knowledge-delta payload on user turn, not assistant

BYK:
## Problem

Users (and a live repro in this repo's own dev session) report a recurring
`⏺ Long-term Knowledge` **dump** followed by a **stall** — the agent does real
work, dumps the LTM block, then needs "continue" to make a bit more progress,
3–4× per session, especially when a background task is running.

## Root cause

The durable knowledge-delta block is a `[user, assistant]` pair (#1315) that
closes the exchange so instruction-literal models (e.g. MiniMax M3) don't ack a
lone user block. The pair put the `## Long-term Knowledge` **markdown payload on
the ASSISTANT turn**. That single choice produced **both** symptoms:

1. **Visible dump.** Agent harnesses (Claude Code REPL, OpenCode) render the
   historical assistant message as a visible turn. The pair is a *durable*
   `session_prompt_delta` replayed byte-identically every turn, so the dump
   recurs — and background curation appends a fresh delta per turn, amplifying
   it on long / Layer-4 sessions.
2. **Premature loop exit.** The model treats the fake assistant turn as an
   already-completed turn and ends early, so the agent loop exits.

### Evidence (verified against production)

- `~/.local/share/lore/lore.db`: **108** post-#1315 `session_prompt_deltas`
  rows carry the payload on `role:"assistant"` (coexisting with 142 pre-#1315
  single-user rows).
- Running gateway `0.39.0-dev.1784906060` (post-#1461/#1465/#1467) **still**
  emits the assistant-role payload — the bug is live on current main, not fixed
  by those (cost-focused) patches.
- Deltas sit **mid-history** (e.g. `insertAt=152` of 286 messages; another
  session 257 of 1750) — not a tail-continuation issue.
- **Not a model echo**: MiniMax-M3 never reproduced the payload across 15+ live
  turns (headless `-p` and direct-route). It is a **client-render + loop-exit
  artifact**, which `-p` mode masks (it extracts only a final result).

## Fix

Move the payload onto the **USER** message (ambient context, where it belongs)
and make the assistant message a tiny inert closer (`"Understood."`).

The `[user, assistant]` pair is preserved, so the exchange stays closed — the
#1315 no-ack property is kept — but **no markdown rides a rendered/completed
assistant turn**, eliminating both the dump and the premature loop exit.

## Regression test

`knowledge-delta-overflow.test.ts` now asserts the payload rides the USER turn
(contains the entry + `## Long-term Knowledge`) **and** the assistant turn is an
inert closer (`not` the payload, `< 40` chars).

**Mutation-verified non-vacuous**: in a throwaway worktree, reverting the fix
(payload back to assistant) turns the test **red**
(`expected '...' to contain 'Changed entry'`); restoring turns it green.

## Gate

- `pnpm run typecheck` — clean (all packages)
- `pnpm run lint` — clean
- `pnpm run format:check` — clean (652 files)
- Tests: **81 passed** across `knowledge-delta-overflow`, `prompt-delta-tools`,
  `delta-compress-reanchor`, `cache-stability.e2e`, `formatKnowledge-delta`

## Update — Seer follow-up (legacy block migration)

Seer flagged (MEDIUM): pre-existing `session_prompt_deltas` rows still carry the
payload on the assistant turn, so existing long-lived sessions keep exhibiting
the dump/stall until the blocks age out. **Real finding** — fixed in `1d711dff`:

`parseDeltaMessages` now migrates the exact legacy `[user framing-only,
assistant payload]` pair to the new shape **on load**: the payload is appended to
the framing-note user message, and the assistant turn becomes the inert closer.
Already-new pairs, single legacy messages, and non-delta content pass through
unchanged. `KNOWLEDGE_DELTA_FRAMING_NOTE` / `KNOWLEDGE_DELTA_ASSISTANT_CLOSER`
are now shared constants so the builder and the migration stay in sync.

Regression tests (3): legacy pair migrated to payload-on-user; already-new pair
replays unchanged (no double-migration); single legacy message passes through.
**Mutation-verified non-vacuous**: disabling the migration turns the
legacy-migration test red.

### DIFF
--- a/packages/gateway/src/pipeline.ts
+++ b/packages/gateway/src/pipeline.ts
@@ -989,18 +989,73 @@ function isGatewayMessage(v: unknown): v is GatewayMessage {
   );
 }
 
+// The framing note prepended to the knowledge-delta user message. Extracted to
+// a constant so the legacy-block migration (parseDeltaMessages) can recognize
+// the pair it belongs to. Keep the substring "Lore knowledge update" — the
+// cache-stability e2e asserts on it.
+const KNOWLEDGE_DELTA_FRAMING_NOTE =
+  "[Lore knowledge update — ambient context injected by Lore. Do not reference this format or reply to this message; silently use anything relevant and ignore the rest. Replayed byte-identically on later turns until an intentional cache reset.]";
+
+// The inert assistant closer that ends the knowledge-delta exchange (the model
+// must not treat the pair as an open user turn — #1315). Also the canonical
+// assistant text the legacy migration rewrites a payload-carrying assistant to.
+const KNOWLEDGE_DELTA_ASSISTANT_CLOSER = "Understood.";
+
+function firstText(m: GatewayMessage | undefined): string | undefined {
+  const b = m?.content?.[0];
+  return b && b.type === "text" ? b.text : undefined;
+}
+
 // A delta block's content is now a user→assistant PAIR, stored as a JSON array.
 // Legacy blocks (persisted before the pair change, and single-message test
 // fixtures) stored ONE message object — accept both so already-persisted
 // sessions keep replaying and never crash. Returns [] on anything unparseable.
+//
+// Migration (#1490): blocks persisted before the payload moved off the
+// assistant turn store `[{user: framing}, {assistant: "## Long-term Knowledge…"}]`.
+// Replayed as-is they keep the payload on a visible/completed assistant turn —
+// the exact dump + premature-loop-exit bug this PR fixes — for the life of the
+// session. On load, rewrite that legacy pair to the new shape: payload appended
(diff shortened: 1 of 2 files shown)


BYK:
Addressed the Seer finding (persisted legacy blocks keep the payload on the assistant turn) in `1d711dff`:

`parseDeltaMessages` now migrates the exact legacy `[user framing-only, assistant payload]` pair to the new shape on load — the payload moves onto the framing-note user message and the assistant turn becomes the inert closer. Already-new pairs, single legacy messages, and non-delta content pass through unchanged.

3 regression tests added (legacy migrated / already-new unchanged / single message passthrough), mutation-verified non-vacuous (disabling the migration turns the test red).

PR #1504, opened 2026-07-27
feat(provenance): widen tool_calls.input_path to input_paths_json (bash-extracted)

BYK:
## What

D2c PR-1b — follow-up to #1424. The v75 column captured a single file path per tool call; widening it to a JSON array of unique paths makes "which files did this session touch?" a union over `tool_calls` rows — the real provenance signal D2c PR-2 needs to associate knowledge entries with the session's file activity.

## Why now

The agent's `bash` tool is the dominant way it touches files (`cat a b c`, `sed -i 's/x/y' file`, `> out`, `tee log.txt`, `for f in src/*; do …`). PR-1 missed all of these because:

- The single `input_path` column stored the **first** path only — `cat a b c` lost `b` and `c`.
- The plaintext path regex required a `/` and couldn't tell a `cat` operand from prose.
- The bash command string was never even parsed.

The result: `tool_calls.input_path` was useful for `Read`/`Edit`/`Write` but blind to bash. That's the wrong shape for D2c.

## How

- **v76 migration**: `RENAME COLUMN tool_calls.input_path → tool_calls.input_paths_json` (SQLite ≥3.25). v75 recovery def subsumed: a DB with neither column now gets `input_paths_json` added directly; a DB created on v75 gets an in-place rename. No production data lost (v75 is 6 days old).
- **`extractFilePath` → `extractFilePaths` (string[])**: same object / JSON / plaintext shapes as v75 (one path → one-element array; same ReDoS guards on the plaintext fallback). Bash command string dispatches to the new helper.
- **`extractFilePathsFromCommand` (unbash AST walk)**: zero-dep tolerant bash parser, surfaces
  - redirect targets for file-touching operators (`>`, `>>`, `<`, `<>`, `>|`, `>&`),
  - positional operands of `FILE_OPERAND_COMMANDS` (`cat`, `tee`, `cp`, `mv`, `rm`, `head`, `tail`, `less`, `more`, `wc`, `stat`, `touch`, `mkdir`, …) — deliberately **excluded** `sed`/`awk`/`grep`/`rg` (their first non-flag arg is a script/pattern, not a file — wrong signature for our heuristic),
  - for-loop wordlists.
  - Words containing parameter/command/arithmetic expansions (`$1`, `$(cmd)`, `${var}`) are skipped; unbash has the parts so we can detect them statically. Tilde-prefixed words skipped even when unbash doesn't tag the part (runtime expansion).
  - Tolerant of syntax errors (unbash never throws; empty result on parse failure). Bounded at `PATH_SCAN_LIMIT` (8KB) and `PATH_TOKEN_MAX` (260ch) for the same ReDoS-safety reason as the plaintext fallback.
- **`recordToolCalls`**: stores `JSON.stringify(paths)` (or NULL). Same COALESCE-keeps-first-non-null UPSERT semantics as v75 (a re-seed never clobbers a captured list). `COLS_PER_ROW` stays 12 — column count unchanged, just the name.
- **`stripAppliedAlters`** now also handles `RENAME COLUMN` and **pre-strips** before exec instead of relying on the catch-and-retry path. Necessary for the v75→v76 ADD+RENAME pair: a DB with `input_paths_json` must treat v75 as a no-op even when v75's ADD itself doesn't throw (because `input_path` is genuinely absent). Without pre-strip, a fresh-DB re-apply of v75 adds `input_path` next to the already-present `input_paths_json`, then v76's RENAME throws `duplicate column name: input_paths_json`. The catch handled the throw-then-strip path; pre-strip is what handles the silent-ADD-because-rename path.

## Tests

- 11 new bash-walker tests: redirects (>/>>, <), pipe, tee, chain (&&, ;), dedup, expansions, tilde, control flow (for wordlist + recurse), tolerant of syntax errors, multi-KB cap.
- Updated the 4 existing `input_path` tests to assert against `input_paths_json`.
- New integration tests: multi-file bash command stores the full list; `>` redirect target stored; COALESCE direction hardened (re-seed with a different non-null list also keeps the first).
- Schema-version test pins bumped 75→76 (db.test.ts ×2, ltm-versioning.test.ts ×1).
- **Mutation-verified**: 4 tests fail when `recordToolCalls` population is neutered; 3 tests fail when bash dispatch is disabled — non-vacuous guards on both code paths.

Format clean. Full `@loreai/core` suite green: **3287 passed / 6 skipped** (3293).

## Risks / out of scope

- `tool_calls` is **not** in `SYNCED_TABLES` (verified `sync-data.ts`) — local-only, zero sync ripple. The column-shape change is invisible to remote.
- The plaintext-fallback slash requirement is preserved (prose risk on raw strings). The AST walk accepts bare relative paths (e.g. `> out.ts`) because the AST tells us the token came from a real shell construct.
- `sed -i` is not surfaced. Acceptable for best-effort provenance; restoring it requires a per-tool option table to distinguish script-vs-file operands.
- `[ -f file ]` test builtin is not surfaced. Out of scope for v1; same family of per-builtin option parsing.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -34,6 +34,7 @@
     "pickleparser": "0.2.1",
     "remark": "^15.0.1",
     "sqlite-vec": "0.1.9",
+    "unbash": "^4.0.4",
     "uuidv7": "^1.1.0",
     "zod": "^4.3.6"
   },
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -1890,6 +1890,20 @@ const MIGRATIONS: string[] = [
   -- with the files that produced them (D2c).
   ALTER TABLE tool_calls ADD COLUMN input_path TEXT;
   `,
+
+  `
+  -- Version 76: rename tool_calls.input_path → tool_calls.input_paths_json
+  -- and widen the value to a JSON array. A file-tool call already had at most
+  -- one path, but a single bash command (cat a b c, sed -i a, > b, tee d …) can
+  -- touch N files — the union is the actual provenance signal D2c PR-2 needs.
+  -- The shape is now: NULL | '["a.ts","b.ts"]' (a JSON string of unique,
+  -- non-empty, slash-bearing, non-pathological tokens extracted at record time).
+  -- SQLite's ALTER TABLE … RENAME COLUMN only updates the schema; pre-existing
+  -- rows keep their (non-JSON) value, so readers use a safe-parse helper that
+  -- also accepts a bare path. v75 is only 6 days old in production, so the
+  -- back-window of non-JSON values is tiny.
+  ALTER TABLE tool_calls RENAME COLUMN input_path TO input_paths_json;
+  `,
 ];
 
 // Index of the migration whose work is performed by a column-presence-aware JS
@@ -2927,17 +2941,25 @@ function migrate(database: Database) {
       // (skips if already migrated) so a partial apply can't boot-loop (#827 E-4c-3a).
       applyScopeKeyEpochPk(database);
     } else {
+      // Pre-strip any column-side ALTERs that are already satisfied by the
+      // current schema (ADD COLUMN where the column exists, RENAME COLUMN
+      // where the source has already been renamed to the target). This makes
+      // every migration idempotent without relying on the catch-and-retry
+      // path below — necessary for migrations like v75→v76 where the original
+      // column was renamed in a later migration and the ADD+RENAME pair must
+      // be jointly treated as a no-op when the target name already exists.
+      const preStripped = stripAppliedAlters(MIGRATIONS[i], database);
       try {
-        database.exec(MIGRATIONS[i]);
+        database.exec(preStripped);
       } catch (e: unknown) {
         // Multi-statement migrations can partially fail when an early
-        // statement (e.g. ALTER TABLE ADD COLUMN) hits a duplicate-column
-        // error from a prior partial run. Swallow duplicate-column errors
-        // so the rest of the migration loop and the version bump proceed.
-        // Any genuinely new error is re-thrown.
+        // statement (e.g. ALTER TABLE ADD COLUMN, or RENAME COLUMN) hits a
+        // duplicate-column / no-such-column error from a prior partial run.
+        // Swallow these so the rest of the migration loop and the version
+        // bump proceed. Any genuinely new error is re-thrown.
         if (e instanceof Error && /duplicate column name/i.test(e.message)) {
-          // The ALTER TABLE already applied — run remaining statements in
-          // this migration by stripping the offending ALTER and re-exec'ing.
+          // The column-side statement already applied — run remaining
+          // statements in this migration by stripping it and re-exec'ing.
           // (Important: migrate() in db.ts runs each migration via database.exec()
           // which stops at the first error in a multi-statement string.)
           const stripped = stripAppliedAlters(MIGRATIONS[i], database);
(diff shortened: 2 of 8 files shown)


BYK:
## Adversarial review response (BLOCKERs all fixed)

Adversarial review before merge surfaced 3 BLOCKERs and 3 SHOULD-FIXes. All fixed in `af19a618` (force-pushed). Re-review requested.

### BLOCKERs

- **B1 — gateway crash on BOTH-columns DB state** (regression vs. pre-PR code).
  *Root cause*: my `stripAppliedAlters` handled `hasTo && !hasFrom` but not `hasTo && hasFrom` (an aborted v75→v76 mid-apply, or a manual sidecar `ADD COLUMN input_paths_json`). The RENAME was left in the migration string, exec threw `duplicate column name: input_paths_json`, the catch branch re-ran the *same* failing SQL with no nested try/catch, and the second throw escaped `migrate()`.
  *Fix*: in `stripAppliedAlters`, when BOTH `input_path` AND `input_paths_json` are present, prepend `ALTER TABLE tool_calls DROP COLUMN input_path;` to the migration string AND strip the RENAME (since the target already exists). The same DROP is mirrored in `recoverMissingObjects` as a safety net. Independently verified by probe + regression test `db.test.ts:v75/v76: both-columns state (partial v75→v76) converges on re-open`.

- **B2 — compound bodies + Statement-level redirects silently lost**.
  *Root cause*: the visitor unwrapped `Script.commands[i].command` (Statement → inner command) but did NOT recurse into Statement nodes inside `CompoundList.commands`. Every compound construct's body is a `CompoundList` of `Statement`s, and the visitor missed them. Additionally, the trailing redirect on `(cat a) > out.ts` / `for … done > out.ts` / `if … fi > out.ts` / `function foo { … } > log` lives on the OUTER Statement/Function node, not the inner Command, so the redirect walker never saw it.
  *Fix*: visit `n.command` for Statement nodes; process `n.redirects` on every node (Command, Statement, Function, Coproc) via a shared `processRedirects` helper; descend into Case `items[i].body`; descend into Subshell, BraceGroup, etc. via the generic `n.commands` recursion (which now triggers because Statements ARE visited).
  *Independent verification*: 6 regression tests added (brace-list, subshell redirect, if-redirect, while-redirect, for+redirect, function body). All pass.

- **B3 — single-quoted path tokens silently dropped (type-discriminator typo + quote not stripped)**.
  *Root cause*: the `isStaticLiteralWord` check was `p.type !== "SingleQuotedPart"`, but the runtime discriminator is `"SingleQuoted"` (`SingleQuotedPart` is just the TS interface name in `unbash/dist/types.d.ts:14`). Compounding the bug, the visitor pushed `w.text` (which includes the quote characters) — even with the discriminator fix, `cleanPathToken` would reject `'db.ts'` because the trailing `'` breaks the `/^[^.]+\.[A-Za-z]{1,5}$/` extension regex.
  *Fix*: change `"SingleQuotedPart"` → `"SingleQuoted"`, and switch all `push(w.text)` → `push(w.value)` (unbash's `value` strips quotes; same as `text` for unquoted words).
  *Independent verification*: 2 regression tests added (`cat 'db.ts'`, `cat 'src/a.ts'`). All pass.

### SHOULD-FIXes (all addressed in the same fix pass)

- **S1 — unmodeled glob expansions leaked as static literals** (`src/*`, `[abc].ts`). Added a glob-metacharacter guard (`/[*?[\]]/.test(word.text)`) up front. 4 negative tests.
- **S2 — heredoc/herestring tags captured as file paths**. Added a `FILE_TOUCHING_REDIRECT_OPS` allowlist that EXCLUDES `<<`, `<<-`, `<<<`. 2 negative tests.
- **S3 — Case clause bodies not walked**. Added `if (n.items) for (const it of n.items) { if (it.body) visit(it.body); }`. 1 regression test.

### NITs

- **N1** — `w.text` → `w.value` (now consistent everywhere; covered by the B3 fix).
- **N2** — Function/Coproc redirects (fixed by the `processRedirects` shared helper processing `n.redirects` on every node type). 1 regression test.

### Test coverage additions

- 10 new bash-walker tests covering all 6 fix categories above.
- 1 new migration test (B1) for the BOTH-columns state.
- 4 new negative tests (glob patterns: `src/*`, `src/*.ts`, `[abc].ts`, `src/file?.ts`).
- 2 new heredoc/herestring negative tests.
- 1 new Case body test.
- 1 new Function-redirect test.

**Total: 11 new tests** (was 30, now 41 in `tool-file-provenance.test.ts`; +1 in `db.test.ts`). Full `@loreai/core` suite: **3298 passed / 6 skipped (3304)**. Format clean.

The 3 BLOCKERs the reviewer found are real and would have shipped silently as the bash gap the PR was meant to close. Thanks for the rigorous probe.

BYK:
## Re-review response (post-fix)

Re-review complete — verdict **safe to merge**. All 3 BLOCKERs and all 3 SHOULD-FIXes from the first review are properly fixed. Two new SHOULD-FIXes surfaced and addressed in `3c8854c4`:

### Re-review fixes

- **SHOULD-FIX-1 — `DoubleQuoted` with all-`Literal` children rejected.** Agents commonly use double-quoted literal paths (e.g. `cat "src/a.ts"` to be safe with special chars). bash strips the quotes; the path is real. The walker now recurses into `DoubleQuoted.parts: DoubleQuotedChild[]` and accepts iff every child is `Literal` (rejects `$VAR`, `$(…)`, `${…}` inside).
  - **Verify**: `cat "src/a.ts"` → `["src/a.ts"]` ✓ ; `cat "$HOME/src/a.ts"` → `[]` ✓ ; `cat "$(echo src/a.ts)"` → `[]` ✓.

- **SHOULD-FIX-2 — `&>` / `&>>` redirect operators missing from allowlist.** Bash 4+ "redirect stdout AND stderr to file" shorthand. Common in modern scripts and CI configs. Added to `FILE_TOUCHING_REDIRECT_OPS`.
  - **Verify**: `echo done &> combined.log` → `["combined.log"]` ✓ ; `echo done &>> combined.log` → `["combined.log"]` ✓.

### Re-review clean

All probes from the first review still pass. Plus I confirmed:
- 7 original-reviewer probes (`{ cat a; cat b; }`, `(cat a) > out`, `if … fi > out`, `while … < in > out`, `for … done > out`, `function`, `function > log`) — all pass.
- 5 extra probes (`case`, `until`, subshell with pipe, deeply nested `if→for→>→fi>final`, `coproc`) — all pass.
- 11 new tests in `tool-file-provenance.test.ts` + 1 in `db.test.ts` cover B1/B2/B3/S1/S2/S3/N1/N2 as labeled.

### Test totals
- `tool-file-provenance.test.ts`: 43 tests (was 30; +13 covering BLOCKER/SHOULD-FIX/SHOULD-FIX-1/SHOULD-FIX-2)
- `db.test.ts`: 123 tests (+1 for B1 both-columns state)
- `@loreai/core` full suite: **3301 passed / 6 skipped (3307)**
- Format clean, lint clean, build green (tsconfig.build.json).

The two new SHOULD-FIXes were easy corrections and made before merge. The first reviewer was right that the BLOCKERs were load-bearing — they would have shipped silently as the bash gap the PR was meant to close.

Issue #1052, opened 2026-06-30
bug: github-copilot provider returning 404s


jonlong:
Hey, really excited to give this project a try, it looks great. Thanks for sharing it!

I wanted to pass along a bug I found when attempting to use the `github-copilot` provider (via the OpenCode plugin). 
I had my agent help me with some of the technical diagnoses, but I am happy to help troubleshoot further if they're not sounding right.

## Bug Description

The gateway returns `404 page not found` for the `github-copilot` provider.

Working theory is that the GitHub Copilot's API serves at `/chat/completions` (no `/v1/` prefix), but the gateway unconditionally prepends `/v1/` when forwarding upstream.

## To Reproduce

1. Configure opencode with a github-copilot model and the lore plugin enabled
2. Run any prompt: `opencode run 'howdy'`

```
lore active: /
lore routing through http://127.0.0.1:3207
lore dashboard: http://127.0.0.1:3207/ui

> build · claude-opus-4.5

lore upstream error: 404 404 page not found

Error: Not Found: 404 page not found
```

## Expected behavior

No 404s :) 

From what I can tell, it looks like requests should be forwarded to `https://api.githubcopilot.com/chat/completions`, not `https://api.githubcopilot.com/v1/chat/completions`.

Here's a snippet from my logs:

```
2026-06-30T02:30:35.757Z INFO  fetch-interceptor: api.githubcopilot.com/chat/completions → gateway (body-detected openai)
2026-06-30T02:30:35.773Z INFO  upstream: https://api.githubcopilot.com (provider=github-copilot, providerURL=https://api.githubcopilot.com, modelRoute=https://api.openai.com, headerUpstream=yes, protocol=openai, auth=bearer:gho_…)
2026-06-30T02:30:36.341Z ERROR upstream error: 404 404 page not found
```

It looks like this is showing :
1. The original SDK request is to `api.githubcopilot.com/chat/completions` — **no `/v1/` prefix**
2. The interceptor catches it via body-shape detection (confirming URL pattern match failed — because there's no `/v1/`)
3. The gateway resolves upstream as `https://api.githubcopilot.com` and then forwards to `${upstream}/v1/chat/completions` → 404

## Environment

- OS: macOS arm64
- `@loreai/opencode`: 0.34.0
- `@loreai/gateway`: 0.34.0

BYK:
On it, thanks and sorry! :)

jonlong:
[...]

PR #1505, opened 2026-07-27
fix(gateway): split recall marker envelope for Anthropic-native clients

BYK:
## Problem

When the gateway suppresses a recall tool_use during streaming, the marker text (e.g. `📚 Searching lore for "X"…`) needs to be visible to the client so the next-turn replay path can reconstruct the tool pair. But for Anthropic-native clients, emitting the marker as an inline content block inside the original message made the marker look like model text rendered before the model's real response — it landed mid-history and looked like a stray reply.

## Solution

For Anthropic-native clients: emit the marker as a SEPARATE Anthropic SSE message envelope (own message_start/message_stop). The original envelope is closed BEFORE the marker opens so the wire is balanced (preamble close → marker envelope → continuation). The continuation opens with its own message_start (don't suppress) since the marker envelope closed the previous one.

For non-Anthropic clients (OpenAI Chat Completions / Responses / Gemini): emit the marker as an inline synthetic text content block at the original block offset. The existing per-client translators forward text blocks as their native `delta.content` / `output_text` / `part` text — so the marker reaches the client inline. This is the correct UX for single-message-per-stream protocols (OpenAI Chat Completions, OpenAI Responses) where multiple messages per stream aren't representable.

## Wire-contract invariants

- Every `message_start` is matched by exactly one `message_stop`
- Anthropic native: 3 envelopes (preamble + marker + continuation), each properly closed
- OpenAI Chat Completions: exactly one `[DONE]` sentinel, exactly one non-null `finish_reason` (`stop` for recall-only, `tool_calls` for mixed)
- Non-Anthropic success path: no double-emission of preamble close (the continuation's terminal events close the envelope inline)

## Key changes

1. `packages/gateway/src/stream/anthropic.ts`: Added `takeHeldBackEvents()` (atomic return + clear) to the recall-aware accumulator interface.
2. `packages/gateway/src/pipeline.ts`:
   - `clientSpeaksAnthropic` field on `recallContext` (set at request ingress based on `req.protocol`)
   - For Anthropic: forward preamble's held-back BEFORE the marker envelope (closes the original envelope)
   - For non-Anthropic: emit the marker as an inline text block (existing per-client translators forward it)
   - All 4 `currentAccum.heldBackEvents()` call sites switched to `takeHeldBackEvents()` to prevent double-emission
3. `packages/gateway/test/recall-marker-split.e2e.test.ts` (new, 7 tests):
   - Anthropic native: split envelope, marker positioning, SSE event order
   - OpenAI Chat Completions: marker as `delta.content`, wire-shape invariants
   - OpenAI mixed-tools: `finish_reason=tool_calls`, single `[DONE]`, Read tool_use forwarded
   - Anthropic mixed-tools: marker envelope + other tool_use forwarded
   - Multi-recall drill-down: unique marker ids + balanced envelopes

## Adversarial review

5 rounds of adversarial review per `quality/REVIEW.md`:
- Round 1: 7 MUST-FIX items (F1, F3, F4-F6, F7, F8) — all fixed
- Round 2: critical SSE bug (original envelope never closed before marker) — fixed
- Round 3: vacuous SSE event-order test + missing `takeHeldBackEvents()` — fixed
- Round 4: duplicate `[DONE]` for non-Anthropic (over-forwarding held-back) — fixed
- Round 5: 3 SHOULD-FIX (mixed-tools coverage, drill-down assertions, JSDoc cleanup) — fixed

Round-5 verdict: **MERGE WITH MINOR CONCERNS** (no MUST-FIX after fixes).

## Verification

- Full recall suite: **111/111** tests pass
- Typecheck: clean
- Lint: clean
- Format: clean
- Mutation tests: 4 of 4 realistic wire-contract regressions caught (M1 re-introduce held-back, M2 swap held-back to after marker, M6 always split, M7 empty marker)

Fixes the recall marker that "never makes it to the UI" for Anthropic-native clients (#1052 symptom) — the marker now lands as a distinct assistant message in the client's transcript instead of inline with model text.


BYK:
Verified the concern and applied a defense-in-depth fix in commit a4693e52.

**Analysis**: The heldBack in the current control flow IS read exactly once:
- If `contAccum.hasRecall() && recallDepth < MAX_RECALL_DEPTH` (line 4973) → promote via `currentAccum = contAccum` and `continue` (line 4978). Line 4998 is NOT reached. The heldBack is then read in the NEXT iteration via `currentAccum.takeHeldBackEvents()` at line 4797.
- Otherwise (no more recall OR depth exhausted) → reach line 4998. `contAccum` is NOT promoted. Its heldBack is read at 4998 and the function returns.

The multi-recall drill-down test at `packages/gateway/test/recall-marker-split.e2e.test.ts:958-1061` exercises this path with wire-shape assertions:
- Balanced `message_start` / `message_stop` counts (would fail on double emission)
- Per-range check: exactly 1 stop between consecutive starts
- Unique marker ids per iteration
Test passes on the current code, proving the control flow is correct.

**Fix applied**: Switched `contAccum.heldBackEvents()` (peek) to `contAccum.takeHeldBackEvents()` (consume) at line 4998. This makes the consume semantics explicit and prevents any future refactor from introducing a double-read. The current control flow was already correct, but the change eliminates the theoretical concern at zero cost.

All 111 recall-related tests pass. Typecheck/lint/format clean.

PR #1543, opened 2026-07-31
fix(import): cover all return-null paths with lastWorkerError (Seer 15617416/0,15617416/1)

BYK:
## Summary

Follow-up to PR #1542. Seer review flagged two MEDIUM bugs that PR #1542 didn't cover.

## Bug 1: HTTP/2 statusText is empty (Seer 15617416/0)

The retry-exhausted path at `llm-adapter.ts:2944` set `lastWorkerError = `HTTP ${response.status}: ${response.statusText}``, but HTTP/2 responses don't include reason phrases — `response.statusText` is empty. Users see `HTTP 429: ` (empty after colon) which is unhelpful.

**Fix:** Use the response body (`text`) as the suffix instead, falling back to statusText for HTTP/1.x.

```ts
lastWorkerError = `HTTP ${response.status}: ${text || response.statusText || "no body"}`;
```

## Bug 2: Many return-null paths don't set lastWorkerError (Seer 15617416/1)

PR #1542 only covered the retry-exhausted and data-policy paths. Several pre-existing paths returned null without setting lastWorkerError:

- `candidateBlocked` — all worker model candidates blocked
- `no-auth` — no auth credentials for the provider
- `cross-provider` — no upstream route (unknown provider, unconfigured)
- `protocol-mismatch` — Anthropic key but openai target (and vice versa)
- Outer network-error catch — timeout/DNS/connection refused

**Fix:** Set `lastWorkerError` on every return-null path. Each message tells the user WHAT went wrong and what to check:

```
all anthropic worker model candidates blocked (likely auth, model-not-supported, or worker-incapable)
no auth credentials available for anthropic (set LORE_WORKER_API_KEY, ANTHROPIC_API_KEY, or similar)
no upstream route for anthropic — provider is unknown or unconfigured (check LORE_*_URL or models.dev cache)
protocol mismatch: openai target but credential is an Anthropic key (set OPENAI_API_KEY or use an anthropic/* model)
network error: no response from anthropic (timeout, DNS, or connection refused)
```

After this PR, the chain's `First error:` line will be specific even for the rare paths that PR #1542 missed.

## Test results

- 117/117 import-* tests pass
- typecheck/format:check clean

### DIFF
--- a/packages/gateway/src/cli/import.ts
+++ b/packages/gateway/src/cli/import.ts
@@ -19,7 +19,7 @@ import {
 } from "@loreai/core";
 type DetectionResult =
   import("@loreai/core").conversationImport.DetectionResult;
-import { createGatewayLLMClient } from "../llm-adapter";
+import { createGatewayLLMClient, getLastWorkerError } from "../llm-adapter";
 import {
   resolveAuth,
   workerKeyScheme,
@@ -1453,7 +1453,10 @@ export async function commandImport(
         label: string;
         providerID: string;
         reason: "auth-rejected" | "no-response";
-        /** First per-chunk error from the loop, when no-response. undefined when auth-rejected. */
+        /** First per-chunk error from the loop or the LLM client, when
+         * no-response. undefined when auth-rejected. Source: the
+         * per-chunk try/catch (throw case) OR the LLM adapter's
+         * getLastWorkerError() (return-null case — most common). */
         lastError?: string;
       }> = [];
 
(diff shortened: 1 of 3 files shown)


BYK:
Closing as duplicate — the same fixes (HTTP/2 statusText + don't overwrite in catch) landed in PR #1544 which is already merged.

PR #1544 was created by rebasing #1543 onto main, then addressing both Seer inline comments inline. The two changes:
- lastWorkerError uses `text` (response body) instead of `response.statusText` (HTTP/2 empty) at llm-adapter.ts:2944
- lastWorkerError preserved via `??=` instead of being overwritten by the network-error placeholder at llm-adapter.ts:2980

PR #1560, opened 2026-08-03
feat(cli): migrate lore sync/team/admin/import to typed Stricli (Phase 3D.3)

BYK:
## Summary

Phase 3D.3 from the Stricli CLI migration plan. The remaining 3D mutation commands — `sync`, `team`, `admin`, and `import` — now route through the typed tree via the shared `runLegacyAndCollect` bridge.

### commands/mutations.ts — typed adapters

- `syncCommand`, `teamCommand`, `adminCommand`, `importCommand` each wrap their legacy counterpart via the shared `runLegacyAndCollect` bridge.
- Empty flag schemas for now — flag parsing still delegated to legacy handler (Phase 3D.3b follow-up).
- Output shape: human renders legacy stdout verbatim; JSON returns `{ output: <captured stdout> }`.

### Routing

- `lib/argv.ts`: `STRICLI_ROUTES` gains `"sync"`, `"team"`, `"admin"`, `"import"`.
- `app.ts`: all four moved out of `LEGACY_ROUTES` into the Stricli tree.
- `main.ts:413-433`: status comment block lists `sync, team, admin, import` as migrated.

## Tests

`cli-mutations-contract.test.ts` — 10 cases:

1. `sync` registered in `STRICLI_ROUTES` (routing wiring).
2. `sync` NOT in `LEGACY_ROUTES` (per-slice removal rule).
3. `team` registered in `STRICLI_ROUTES` (routing wiring).
4. `team` NOT in `LEGACY_ROUTES` (per-slice removal rule).
5. `admin` registered in `STRICLI_ROUTES` (routing wiring).
6. `admin` NOT in `LEGACY_ROUTES` (per-slice removal rule).
7. `import` registered in `STRICLI_ROUTES` (routing wiring).
8. `import` NOT in `LEGACY_ROUTES` (per-slice removal rule).
9. `sync` forwards to `commandSync` and surfaces its stdout (hermetic mock).
10. `import` forwards to `commandImport` and surfaces its stdout (same pattern).

The legacy handlers are mocked via `vi.doMock` so the tests don't touch Supabase or any IO.

## Verification

- 162 CLI tests pass across 17 files (added 10 new)
- TypeScript clean
- Lint clean (zero errors)
- Format clean
- Bundle succeeds for both CJS (Node) and ESM (Bun) targets

## Deferred (follow-up PRs)

[...]

PR #1561, opened 2026-08-03
feat(cli): migrate lore sync with explicit positional schema (Phase 3D.3b)

BYK:
## Summary

Follow-up to PR #1560 (Phase 3D.3) which was reverted due to a CRITICAL regression — empty schemas rejected every positional/flag invocation. This PR migrates `lore sync` alone with the proper positional schema. `team`, `admin`, and `import` will follow in subsequent Phase 3D.3 PRs.

## Why this PR

The independent review on #1560 found that the typed wrappers declared `parameters: { flags: {} }` with no positional schema. Stricli's strict parser rejected every documented invocation:

- `lore sync enable` → "Too many arguments"
- `lore team list` → "Too many arguments"
- `lore admin grant foo@bar pro` → "Too many arguments"
- `lore import --dry-run` → "No flag registered for --dry-run"

Bare defaults still worked, which is why the existing 10 tests didn't catch it. **The lesson**: a typed command that drops a documented legacy invocation is worse than no migration at all.

### commands/sync.ts — typed adapter

- Wraps `commandSync` from `../sync-cmd` via the shared `runLegacyAndCollect` bridge.
- Schema: `positional: { kind: "tuple", parameters: [{ parse: String, brief: "Subcommand (status | enable | disable | now)", optional: true }] }`.
- `commandSync` reads `positionals[0]` from the legacy args dict — Stricli spreads `ARGS = [string?]` into the handler signature, so the handler receives the optional subcommand string directly and wraps it in `[sub]` for the legacy call.
- No flags (auto-injected `--json` only).

### Routing

- `lib/argv.ts`: `STRICLI_ROUTES` gains `"sync"`.
- `app.ts`: `sync` moved out of `LEGACY_ROUTES` into the Stricli tree.
- `main.ts:413-439`: status comment block lists `sync` as migrated (Phase 3D.3b).

## Tests

`cli-sync-contract.test.ts` — 6 cases:

1. `sync` registered in `STRICLI_ROUTES` (routing wiring).
2. `sync` NOT in `LEGACY_ROUTES` (per-slice removal rule).
3. positional schema declared — help text shows enable/disable/status/now.
4. `lore sync enable` forwards `["enable"]` to commandSync (**would have caught C-1**).
5. `lore sync` (no positional) still reaches handler with `[]`.
6. `lore sync status --json` produces `{ output: "..." }` envelope.

The legacy handler is mocked via `vi.doMock` so the test doesn't touch Supabase.

## Verification

- 158 CLI tests pass across 17 files (added 6 new)
- TypeScript clean
- Lint clean (zero errors)
- Format clean
- Bundle succeeds for both CJS (Node) and ESM (Bun) targets

## Next steps (Phase 3D.3c/d/e)

- `lore team` — 1+ positionals + many flags (invite, role, email, offline, project, etc.)
- `lore admin` — 3 positionals (sub, target, tier)
- `lore import` — 11+ flags (dry-run, yes, agent, source, file, global, mem0-*, no-worktrees, project)

Each gets its own PR with the full flag/positional schema declared upfront.


BYK:
Superseded by PR #1562 (fresh branch `feat/cli-mutations-fresh` off main, including the F-1 cli.ts scanner-error exit-code remap from PR #1569). The original PR #1561's branch couldn't be cleanly rebased onto main because intermediate commits had been force-pushed out. Closing this in favor of the new PR.

PR #1590, opened 2026-08-06
fix(lint): surface judge null-text on stderr + enable LORE_DEBUG in CI

BYK:
## Problem

PR #1587 CI run 31099868171 reported "20/20 unparseable" on github-copilot/gpt-5.6-luna with **no actionable stderr**. Every judge call completed (no exception) but the parser got null text back, and @loreai/core's `log.info` is debug-gated — so an operator looking at "20/20 unparseable" could only guess at the routing/auth/encrypted-content cause.

PR #1555 made the bash `cat /tmp/lore-ic.err` unconditionally, so stderr is captured. But the *content* of stderr was still gated behind LORE_DEBUG=1.

### 1. `.github/actions/lint/action.yml` — `LORE_DEBUG: "1"`

Added to the `Resolve range + run lint` env block. The bash already captures stderr unconditionally (per PR #1555), but the logger is debug-gated — without LORE_DEBUG=1 the only stderr visible on success was the per-call judging progress lines. With it, every `log.info` / `log.warn` from @loreai/core surfaces too, so the next "20/20 unparseable" run shows the model, the routed protocol, and any worker-pipeline error message inline.

### 2. `packages/core/src/invariant-check.ts` — split unparseable into two cases

The previous code lumped both "worker returned null text" and "worker returned text but parser couldn't extract JSON verdict" into a single `unparseable++` branch with no logging. Split:

- `responseText === null` (worker returned no text — routing miss / auth rejection / encrypted-only) → `log.notice` (always visible on stderr, sink warn severity, **never** `log.error` per the [lore] 'expected user-actionable errors are warn not error' rule). Names the model + provider + likely cause.
- `responseText` is a string but unparseable JSON → `log.info` (debug-gated — model behavior, not operator-actionable). Logs the truncated text for offline triage once LORE_DEBUG=1 is on.

The `unparseable++` counter is preserved in both branches so a systemically-broken judge stays visible rather than silently indistinguishable from a genuine clean run.

## Regression test

`invariant-check.test.ts > "emits a notice log when the judge returns null text (so 20/20 unparseable in CI is actionable)"`. **TDD-verified per lore [019fb95e]:**
- Stash fix → test FAILS at `expect(noticeSpy).toHaveBeenCalledTimes(1)` (got 0 times)
- Restore → test PASSES (1 time, contains "github-copilot/gpt-5.6-luna" + "null text")

## Pre-merge gate

- format ✓ (717 files)
- lint ✓ (only pre-existing `typescript(unbound-method)` warnings remain)
- typecheck ✓ (all 5 packages)
- tests: **7198 passed, 227 skipped, 2 failed** in 398 test files (813s)
  - 2 failures in `packages/gateway/test/import-auth-chain.test.ts` cases (a) and (b)
  - **Pre-existing on origin/main** — verified by checking out main and running the same file: same 2 failures, 43 passed / 45 total. Not introduced by this PR.
  - Per lore `[019fab42]` "Separate PR for unrelated CI/workflow fixes; rigorous PR hygiene", these belong in a follow-up PR, not this one.

## Files

```
.github/actions/lint/action.yml            |  10 +++
packages/core/src/invariant-check.ts       |  33 +++++++-
packages/core/test/invariant-check.test.ts |  45 ++++++++++
3 files changed, 84 insertions(+), 4 deletions(-)
```

### DIFF
--- a/.github/actions/lint/action.yml
+++ b/.github/actions/lint/action.yml
@@ -113,6 +113,16 @@ runs:
         LORE_DB_PATH: ${{ steps.key.outputs.db }}
         CACHE_HIT: ${{ steps.cache.outputs.cache-hit }}
         LORE_IC_GATE: ${{ inputs.gate }}
+        # Enable log.info/log.warn to surface on stderr. The action's bash already
+        # captures stderr unconditionally (per PR #1555), but the @loreai/core
+        # logger is debug-gated — without LORE_DEBUG=1 the only stderr we ever see
+        # on a successful lint run is the per-call judging progress lines. When
+        # a judge call returns null text (routing miss, auth failure, encrypted
+        # content, etc.) the unparseable++ count surfaces in the report but the
+        # underlying cause stays silent. Surfacing it here lets the next CI run
+        # pinpoint which path failed, instead of treating "20/20 unparseable" as
+        # the symptom we guess about in PR comments.
+        LORE_DEBUG: "1"
       run: |
         set -euo pipefail
         mkdir -p "$(dirname "$LORE_DB_PATH")"
--- a/packages/core/src/invariant-check.ts
+++ b/packages/core/src/invariant-check.ts
@@ -1075,12 +1075,37 @@ export async function checkInvariants(input: {
       continue;
     }
     judgeCalls++;
+    if (responseText === null) {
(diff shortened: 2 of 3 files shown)


BYK:
Re-run CI

PR #1604, opened 2026-08-08
fix(gateway): close vector-pool readers and main SQLite writer on shutdown (#1599)

### LINKED ISSUE #1599: gateway: close vector readers and SQLite during graceful shutdown
## Problem

The gateway shutdown closure stops HTTP traffic and embedding work, but it never closes the read-worker SQLite connections or the main SQLite writer before the signal path force-exits.

Current shutdown sequence in `packages/gateway/src/cli/start.ts:442-472`:

```ts
boundServer.stop();
removePortFile(actualPort);
removePidFile();
await resetPipelineState({ fast: true });
await embedding.settleDocumentEmbeds(EMBED_DRAIN_DEADLINE_MS);
await embedding.resetProvider();
```

The core already has the correct writer close primitive at `packages/core/src/db.ts:3732-3751`:

```ts
instance.exec("PRAGMA busy_timeout = 0");
instance.exec("PRAGMA wal_checkpoint(TRUNCATE)");
instance.close();
instance = undefined;
```

It reads `instance` directly, so calling `close()` never forces a database open.

The vector worker pool also owns independent query-only WAL readers. `packages/core/src/db/reader.ts:30-44` says the caller owns each connection lifetime. The worker closes its reader on a shutdown message at `packages/core/src/vector-worker.ts:144-153`, but `shutdownVectorPool()` in `packages/core/src/vector-pool.ts:782-801` posts shutdown and immediately calls `worker.terminate()` without waiting for reader closure.

Today the signal handler always reaches `forcedExit()` after the bounded closure. SQLite WAL recovery protects correctness, but every normal service stop skips the existing checkpoint-and-close path and may leave readers alive while the writer attempts to close.

## Required behavior

1. Stop accepting HTTP traffic and stop scheduling background work.
2. Finish the existing bounded in-flight embedding drain and embedding-worker reset.
3. Shut down all vector/read workers and wait for their SQLite readers to close.
4. Call the existing core `close()` function after readers are gone.
5. Keep the existing forced-exit protection for a timeout, a second signal, or a native ONNX call that cannot stop.

## Invariant

🔴 `packages/gateway/src/cli/shutdown.ts:9-11`: "shutdown can never block the process longer than a hard deadline" and "a second SIGINT/SIGTERM forces an immediate exit."

The default remains 4,000 ms (`packages/gateway/src/cli/shutdown.ts:15-40`), invalid configuration must still fall back safely, and the timeout can never be disabled. Never replace the timeout path with `process.exit()`: `packages/gateway/src/cli/shutdown.ts:113-135` requires `forcedExit()` because a native ONNX worker may still be mid-inference.

## Acceptance criteria

- [ ] A bounded async vector-pool shutdown waits for every worker exit/reader close, always resolves, and never rejects.
- [ ] The gateway shutdown order is asserted: pipeline stop -> embedding drain/reset -> vector readers closed -> writer checkpoint/close.
- [ ] Core `close()` is called exactly once on the normal path and never opens an unopened database.
- [ ] Busy reader/checkpoint failures remain best-effort and never exceed the global shutdown deadline.
- [ ] Timeout and second-signal paths still use `forcedExit()` and return promptly.
- [ ] Reopening the database after SIGTERM passes `PRAGMA integrity_check` and preserves accepted writes.
- [ ] Tests cover no vector workers, healthy workers, a worker that never exits, an unopened DB, and an already-closed DB.

## Out of scope

- Changing the 4-second hard deadline.
- Draining queued LLM background work that the current fast shutdown deliberately replays next session.
- Changing SQLite WAL mode or making EBS snapshot behavior part of gateway shutdown.

BYK:
## Summary

Closes #1599.

The gateway shutdown sequence stopped HTTP traffic and reset the pipeline,
but never closed the vector-pool workers' SQLite readers or the main-thread
SQLite writer before the signal-handler force-exit. WAL recovery protects
correctness, but every normal service stop skipped the existing
checkpoint-and-close path and could leave readers alive while the writer
attempted to close — busy-waiting up to 5s on `PRAGMA wal_checkpoint(TRUNCATE)`.

## What changed

- **core/vector-pool: `shutdownVectorPoolAsync`** — bounded, cooperative
  shutdown that posts the shutdown message, waits up to 1s for each worker's
  exit event, and force-terminates on timeout. The synchronous
  `shutdownVectorPool()` remains for the test seam (FakeWorker fires exit
  from `terminate()`).
- **core/index: export `shutdownVectorPoolAsync`** from `@loreai/core`.
- **gateway/start: `shutdownVectorPoolAsync()` + `close()`** wired into
  the shutdown chain AFTER `embedding.resetProvider()` so the writer's
  checkpoint+close runs only after every reader has detached.
- **tests**: 7 new vector-pool tests (no workers, healthy workers, never-exits
  workers, idempotent calls, postMessage-throw tolerance, invalid timeout
  override).
- **tests**: 2 new `close()` idempotency tests (unopened DB, already-closed DB).
- **tests**: 3 new gateway shutdown order tests verifying the full chain
  (pipeline stop → embed drain → embed reset → vector pool read → writer close).

## Acceptance criteria

- [x] Bounded async vector-pool shutdown waits for every worker exit/reader
  close, always resolves, and never rejects.
- [x] Gateway shutdown order: pipeline stop → embedding drain/reset → vector
  readers closed → writer checkpoint/close.
- [x] Core `close()` is called exactly once on the normal path and never
  opens an unopened database.
- [x] Busy reader/checkpoint failures remain best-effort.
- [x] Timeout and second-signal paths still use `forcedExit()` and return
  promptly.
- [x] Tests cover no vector workers, healthy workers, a worker that never
  exits, an unopened DB, and an already-closed DB.

## Verification

- `pnpm typecheck` — passes
- `pnpm lint` — no new warnings
- `pnpm format:check` — passes
- `pnpm vitest run packages/core/test/vector-pool.test.ts packages/core/test/db.test.ts` — 188 passed
- `pnpm vitest run packages/gateway/test/shutdown.test.ts packages/gateway/test/shutdown-embed-drain.test.ts packages/gateway/test/exit.test.ts packages/gateway/test/shutdown-vector-pool-close.test.ts packages/gateway/test/start-gateway-quiet.test.ts` — 35 passed

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -97,6 +97,7 @@ export {
   type VecWorkerCheck,
   checkReadOffload,
   type ReadOffloadCheck,
+  shutdownVectorPoolAsync,
 } from "./vector-pool";
 export { load, config, type LoreConfig } from "./config";
 export {
--- a/packages/core/src/vector-pool.ts
+++ b/packages/core/src/vector-pool.ts
@@ -781,7 +781,16 @@ export async function checkReadOffload(
 
 /** Tear down the pool (test teardown + process reset). Idempotent. The
  *  shuttingDown guard keeps the terminate()-induced worker exits below from
- *  being counted as structural failures. */
+ *  being counted as structural failures.
+ *
+ *  Synchronous fire-and-forget: posts the shutdown message and immediately
+ *  force-terminates each worker. The reader close inside the worker thread
+ *  (vector-worker.ts:144-153) is NEVER given a chance to run — fine for the
+ *  test seam (FakeWorker fires `exit` synchronously from `terminate()`) but
+ *  NOT for the real gateway shutdown, which needs the reader to close BEFORE
+ *  the main thread's `close()` runs (`PRAGMA wal_checkpoint(TRUNCATE)` while
+ *  a reader holds the WAL would busy-wait up to 5s). Use
+ *  {@link shutdownVectorPoolAsync} for that. */
 export function shutdownVectorPool(): void {
   shuttingDown = true;
   for (const pw of workers) {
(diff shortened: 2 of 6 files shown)


BYK:
Superceded

PR #1697, opened 2026-09-01
fix(gateway): follow same-owner symlinked setup configs

rgutzen:
## Summary
- `lore setup` / `lore doctor` fail with `Unsafe setup path ...: symbolic links are not allowed` whenever a harness config (`~/.claude/settings.json`, `opencode.json`, hermes config, ...) is itself a symlink — even when the symlink and its target are both owned by the current user.
- This makes setup unusable under GNU Stow, chezmoi, and similar personal dotfile managers, which commonly replace a config file with a symlink into a separately version-controlled location.
- Resolves the symlink to its real path whenever the whole chain resolves to a regular file owned by the same uid as the process, and performs every trusted-file check/read/atomic-write against that real path instead. A broken link, a non-regular-file target, or a target owned by someone else still fails closed with the same error as before.
- Nothing ever renames or hard-links over the symlink itself (which would silently sever it from its target) — the link stays intact and the real file underneath it is what gets read and rewritten.

## Changes
- `packages/gateway/src/cli/json-config.ts`: adds `trustedRealPath()` and applies it at every trusted-file entry point (`trustedFileExists`, `readTrustedTextFile`, `atomicWriteTrustedFile`, `removeTrustedFile`, `restoreTrustedFileSnapshot`, `withTrustedFileTransaction`'s `assertCurrentIdentity`, `stageTrustedFileMutation`'s path list).
- `packages/gateway/test/setup-io.test.ts`: turns the old "always reject symlinks" test into "follow a same-owner symlink and preserve it", and adds coverage for a symlink target that's not a regular file and a symlink that doesn't resolve at all.

## Test plan
- [x] `pnpm --filter @loreai/gateway run typecheck`
- [x] `vitest run packages/gateway/test/setup-io.test.ts packages/gateway/test/setup.test.ts packages/gateway/test/uninstall.test.ts` (144 passed)
- [x] Manually verified `lore doctor` against a real GNU Stow-managed `~/.claude/settings.json` symlink — previously failed with `Unsafe setup path ...: symbolic links are not allowed`, now reads through it correctly.

### DIFF
--- a/packages/gateway/src/cli/json-config.ts
+++ b/packages/gateway/src/cli/json-config.ts
@@ -11,6 +11,7 @@ import {
   mkdirSync,
   openSync,
   readFileSync,
+  realpathSync,
   renameSync,
   unlinkSync,
   writeFileSync,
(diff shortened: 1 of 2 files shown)


BYK (approved):
Heya, thanks a lot for this!

Issue #807, opened 2026-06-18
cch: defensively handle billing headers that arrive without a cch segment


BYK:
## Background

Claude Code 2.1.181 only emits the `cch` billing field when it believes it is talking to the first-party API (host == `api.anthropic.com`). PR #806 fixes this for all `lore`-managed launches by setting `_CLAUDE_CODE_ASSUME_FIRST_PARTY_BASE_URL=1` (both `lore run` via `agents.ts` and `lore setup` via `setup.ts`).

## Gap

The gateway still has **no defensive fallback** for a billing header that arrives **without** a `cch` segment (e.g. a manually launched client that bypasses `lore run`/`lore setup`, or a future env-var rename by Anthropic):

- `BILLING_HEADER_RE` in `packages/gateway/src/cch.ts` requires `cch=[0-9a-fA-F]+;`. A `cch`-less header fails it.
- `hasBillingHeader()` → false → the resign gate (`pipeline.ts:~2966`) is skipped → the body is forwarded **unsigned and unchanged**, silently.
- `captureBillingPrefix()` (`cch.ts`) is the only setter of `sessionNeedsBilling`, and it's gated on the same regex. So a `cch`-less header means the session is never marked as needing OAuth billing → `buildBillingBlock` / `buildOAuthWorkerHeaders` / `isClaudeCodeOAuthSession` all no-op → worker calls (distillation/curation) lose their billing header and OAuth fingerprint, risking 401s.

This was previously unreachable (pre-2.1.181 always emitted `cch`); 2.1.181 makes a `cch`-less billing header a now-reachable state for any non-`lore`-managed launcher.

## Proposed fix

- Recognize a billing header that lacks `cch` (relax `BILLING_HEADER_RE` / add a `cch`-less branch) so `captureBillingPrefix` still marks the OAuth session.
- When signing, inject a `cch=00000;` segment if absent, then sign — so a `cch`-less body is still re-signed correctly.
- Add regression tests for: (a) a billing header with no `cch` is detected and signed; (b) the OAuth session is marked; (c) existing `cch`-bearing path unchanged.

## References

- PR #806 (the env-var fix), `quality/CCH.md` (first-party gate section).
- Surfaced by an independent code review of #806.

PR #1700, opened 2026-09-02
fix(gateway): stop misclassifying forced-attribution classifier calls as coding turns

rgutzen:
## Problem

Running Claude Code through the Lore gateway breaks auto mode. Sessions report, verbatim:

- `[bug] Bash tool unusable for ~1 hour: safety classifier (claude-sonnet-5)`
- `[bug] Bash safety-classifier outage blocked all non-trivial commands for a full session`

Claude Code's auto-mode permission classifier (one call per tool action, model `claude-sonnet-5`) fails repeatedly, and after 3 consecutive bad verdicts Claude Code drops auto mode to fail-closed — every Bash action gets blocked or prompted.

## Root cause

PR #1161 introduced the side-channel filter in `packages/gateway/src/side-channel.ts`, which forwards Claude Code auxiliary calls (auto-mode classifier, title/topic generation, subagent naming) upstream untouched. It classifies a request as a *side-channel* when it is a Claude Code client **and** `hasClaudeCodeCodingPrompt()` returns `false`. That predicate used the anchored OAuth billing header as a **sufficient** signal:

```
hasClaudeCodeCodingPrompt(system)
  = hasBillingHeader(system)          // signal 1 (SUFFICIENT)
  || "Working directory:" marker      // signal 2
  || authoritative cwd/CLAUDE.md path // signal 3
```

At the time, classifier calls were built with `skipSystemPromptPrefix: true` and carried **no** billing header, so signal 1 never fired for them and they were correctly passed through.

**Claude Code 2.1.258 changed this.** The classifier request is now built with `forceAttributionHeader: true`, so it carries the anchored billing header at `system[0]` even though it is still a `skipSystemPromptPrefix` side-channel call (it never embeds the coding prompt).

With signal 1 firing, `isClaudeCodeSideChannel()` returns `false` for the classifier, so it falls into the full context pipeline — LTM injection, distilled-prefix injection, and structural-compaction mis-routing — which corrupts the verdict and triggers the fail-closed fallback.

## Solution

Demote the billing header from a *sufficient* signal to advisory in `hasClaudeCodeCodingPrompt()`. The discriminator now relies on:

1. the `Working directory:` marker line (present in every real coding turn, including subagents, on any OS), and
2. authoritative workspace inference (a `cwd` field or a CLAUDE/AGENTS/.lore.md path).

Neither of these is carried by a `skipSystemPromptPrefix` classifier call, so the classifier (and any other `forceAttributionHeader` side-channel) is again forwarded upstream untouched. A real coding turn always carries the marker, so it is never mis-classified — with or without the billing header.

The change is scoped to the side-channel discriminator only. `hasBillingHeader()` is untouched and continues to drive the re-sign gate, `captureBillingPrefix()`, and the OAuth worker-header logic.

## Changes

- `packages/gateway/src/side-channel.ts` — drop `hasBillingHeader(system)` from `hasClaudeCodeCodingPrompt()`; update docs.
- `packages/gateway/test/side-channel.test.ts` — regression tests: billing header alone no longer marks a coding prompt; a forced-attribution classifier still bypasses the pipeline; a real coding turn (billing header + `Working directory:`) is still not a side-channel.
- `packages/gateway/test/session-rotation-merge.test.ts` — update CC coding-turn fixtures to carry the `Working directory:` marker (they previously relied on the billing header alone).

## Related

- PR #1161 — the original side-channel fix; its "no billing header" assumption is what regressed in 2.1.258.
- Issue #807 — `cch: defensively handle billing headers that arrive without a cch segment` (adjacent: it made billing-header detection *more* permissive, compounding this mis-detection).

### DIFF
--- a/packages/gateway/src/side-channel.ts
+++ b/packages/gateway/src/side-channel.ts
@@ -5,10 +5,12 @@
  * turns: the auto-mode permission classifier (one call per tool action),
  * conversation title/topic generation, and subagent naming/summary. These are
  * built with `skipSystemPromptPrefix: true`, so they carry NEITHER the coding
- * system prompt (no "Working directory:" line, no CLAUDE.md content) NOR the
- * anchored OAuth billing header — yet they still carry the SAME
- * `x-claude-code-session-id` header as the live coding conversation (Claude
- * Code attaches it to every request).
+ * system prompt (no "Working directory:" line, no CLAUDE.md content) NOR —
+ * since Claude Code 2.1.258, where the classifier request is built with
+ * `forceAttributionHeader: true` — is the anchored OAuth billing header a
+ * reliable discriminator: the classifier now DOES carry it at `system[0]`.
+ * All of these calls still carry the SAME `x-claude-code-session-id` header as
+ * the live coding conversation (Claude Code attaches it to every request).
  *
  * Running these through Lore's context pipeline is harmful:
  *   - LTM system blocks + the distilled conversation prefix get injected, and
@@ -25,7 +27,6 @@
  * any Lore processing (`handlePassthrough`), never touching session state or
  * memory.
  */
-import { hasBillingHeader } from "./cch";
 import { inferProjectPathDetailed } from "./config";
 import { isClaudeCodeClient } from "./session";
 import type { GatewayRequest } from "./translate/types";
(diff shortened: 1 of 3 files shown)


BYK (approved):
Thanks so much!

PR #1705, opened 2026-09-05
fix(pi): build for the Bun runtime —         node:sqlite does not exist there

rgutzen:
## Problem                                              
                                                                                                   
   Pi 0.84.x is a Bun-compiled binary (embedded Bun, jiti loader for .ts extensions). The pi       
 bundle was built with `conditions:["node"]`, which selected @loreai/core's node target        
 (`node:sqlite`) and failed at load:                                                             
                                                                                                   
   ```                                                                                          
   ResolveMessage: No such built-in module: node:sqlite                                            
   ```                                                                                          
                                                                                                   
   This broke Pi runtime initialization when loading extension code.                               
                                                                                                   
   ## Solution                                                                                     
                                                                                                   
   Build the pi bundle with `conditions:["bun"]` to select core's `bun:sqlite` driver        
 instead. Add `bun:*` to the externals list so Bun's built-in modules resolve correctly.         
                                                                                                   
   ## Verification                                                                                 
                                                                                                   
   - ✅ All 37 pi tests pass                                                                       
   - ✅ Real-runtime e2e verified against Bun binary                                               
   - ✅ Error now reaches the auth gate (expected behavior), not the module loader                 
                                                                                                   
   ## Changes                                                                                      
                                                                                                   
   - Update build script conditions from `["node"]` to `["bun"]`                           
   - Add `bun:*` pattern to externals                                                            
   - Update comment to reflect that Pi runs on Bun (0.84.x+), not Node

### DIFF
--- a/packages/pi/script/build.ts
+++ b/packages/pi/script/build.ts
@@ -1,8 +1,12 @@
 /**
- * Build @loreai/pi into a publishable ESM bundle for Node.
+ * Build @loreai/pi into a publishable ESM bundle for the Bun-based Pi runtime.
  *
- * Pi runs on Node exclusively (unlike OpenCode which is Bun), so we only
- * build one target with `node:sqlite` imports preserved via `conditions: ["node"]`.
+ * Pi 0.84.x is a Bun-compiled binary (embedded Bun, jiti loader for .ts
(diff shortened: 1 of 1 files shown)


BYK (approved):
I'm hoping this is compatible both ways?

Thanks so much for all the PRs recently btw!

Issue #1716, opened 2026-09-07
fix(gateway): recover gracefully when the recall budget is exhausted


BYK:
## Problem

A real resumed OpenCode/Codex session stopped with:

```text
Recall depth limit reached (10).
```

The ten-round recall cap is an intentional resource bound. The failure is the terminal handling: Lore replaces the next recall request with a synthetic assistant-text marker and ends the response without giving the model a chance to complete the user's task using the information already retrieved. In the Responses streaming path, the existing regression explicitly expects a normal `response.completed` containing this marker.

This is independent of #1710's full-history resume cost. Faster preparation alone does not address it.

## Evidence — 7 September 2026 (UTC)

The user reported this terminal message after a resumed request containing 5,729 source messages:

- 18:07:33.170 — turn logged.
- 18:09:44.705 — `semantic-preparation [object Object]` (separately tracked in #1715).
- 18:09:44.706 — first subsequent upstream log.
- Additional upstream calls continued through 18:12:03.086.
- 18:12:22.403 — turn warmup-probe bookkeeping logged.
- The user then observed the recall-depth marker and the session stopped.

Background curation and consolidation also completed during this interval. The upstream log lines do not identify request purpose, so they cannot all be attributed to recall. These logs do not establish whether the ten recalls were useful, repeated, empty, or unsuccessful.

## Confirmed implementation

- `packages/gateway/src/recall.ts`: `MAX_RECALL_DEPTH = 10`.
- `packages/gateway/src/pipeline.ts`, `streamResponsesRecallAware()`: once the recall budget is exhausted, a further recall is replaced by the marker, no next recall continuation is scheduled, and the stream is finalized.
- The buffered recall path similarly strips outstanding recall tool calls and substitutes the marker.
- `packages/gateway/test/openai-responses-recall-aware-stream.test.ts` has a test named `replaces an exhausted continuation recall with a marker` that asserts this behavior with a depth limit of one.

Review all recall paths, including non-Responses streaming, so the recovery policy is consistent.

## Recovery policy — updated after user feedback

Keep the bounded recall budget, and **keep the tool definitions and tool controls unchanged**: retracting recall mid-chain can invalidate the cached tool prefix.

1. Append static budget guidance to the result of the last permitted recall, asking the model to finish the task from available results or return an ordinary tool call.
2. Use that follow-up as the one final continuation; no extra recovery request or final transport retry is allowed.
[...]

Issue #1710, opened 2026-09-07
Make long-session resume O(delta): persist a source transcript frontier and active model window


BYK:
Parent investigation: #1708

## Problem

A real resumed OpenCode session showed:

```text
13:37:39.461  turn received: messages=5580
13:37:39.831  max_tokens resolved
13:38:53.865  ltm-budget
13:38:55.690  gradient complete: 180409 tokens
13:38:55.710  first upstream request
```

Lore therefore spent about **74 seconds after request parsing/session identification but before LTM selection**, on a known Tier-1 session. The gradient itself took only about 1.8 seconds. OpenCode reconstructing/sending the transcript is real overhead, but it cannot explain this measured Lore-side interval.

The pathological dimension appears to be structural history size (`5580` messages), not only tokens (`180409`).

## Current hot path

On every normal turn, including a resumed known session, Lore currently rebuilds the full semantic transcript:

1. `gatewayMessagesToLore(req.messages, sessionID)` over every source message.
   - SHA-256 over complete message content for the current deterministic ID.
   - A second complete-content hash for the legacy ID.
   - Another hash per part.
   - Part/object allocation for the entire transcript.
   - BPE work for provenance-bearing messages.
2. `responsesProvenanceByMessageId(...)` over the full history.
3. `resolveToolResults(...)` makes three passes over all messages/parts.
4. Tool-result-only messages invoke `temporal.storedMessageId(...)`; that performs a synchronous SQLite compatibility lookup. A tool-heavy coding transcript can therefore issue a large N+1 query set on the pre-upstream main-thread path.
5. Post-response temporal bookkeeping can rebuild/resolve the full transcript again even though it only persists the latest user message and new assistant response.

`gatewayMessagesToLore` already has a `startIndex` argument, but the pipeline currently calls it with the complete history.

## Design goal

For an append-only resumed session, semantic preparation should be:

```text
O(new/changed source suffix + bounded active raw window)
```

—not `O(all historical messages)`.

Request-body decoding may remain `O(wire bytes)` until #TBD-streaming lands, but hashing, BPE, object construction, tool pairing, DB identity resolution and gradient input construction must become suffix/window bounded.

## State model

Keep two distinct durable concepts. Do not overload gradient's `lastKnownMessageCount`, which is the **compressed model-window count**, not the source transcript frontier.

### 1. Source transcript frontier

Answers: “Which client-owned transcript prefix has Lore already validated and ingested?”

Suggested persisted fields:

```ts
[...]

PR #1717, opened 2026-09-07
fix(perf): reuse exact provenance token counts across restarts

### LINKED ISSUE #1715: fix(observability): preserve semantic-preparation timings in persistent logs
## Problem

The semantic-preparation instrumentation added in #1714 loses its counters and stage timings in the persistent log and registered log sink:

```text
2026-09-07T18:09:44.705Z [INFO ] semantic-preparation [object Object]
```

This prevents diagnosis from user-supplied logs of the remaining long-session delay tracked in #1708 and #1710.

## Cause

`PreparationTiming.upstreamStart()` in `packages/gateway/src/semantic-preparation.ts` calls:

```ts
log.info("semantic-preparation", { ...this.counts, stages: this.stages });
```

However, `formatArgs()` in `packages/core/src/log.ts` converts non-string/non-Error arguments with `String(a)`. The resulting `[object Object]` is passed to the registered sink and persistent file. The separately emitted Sentry metrics are not affected by this formatting defect; this issue concerns the readable log output.

## Production evidence — 7 September 2026 (UTC)

A resumed OpenCode/Codex request with 5,729 source messages produced:

| Event | Timestamp |
| --- | --- |
| Turn received | 18:07:33.170 |
| Idle-cache refresh logged | 18:07:33.441 |
| LTM budget logged | 18:08:44.849 |
| Distillation prewarm complete (27 rows) | 18:09:43.083 |
| Semantic-preparation log | 18:09:44.705 |
| First subsequent upstream log | 18:09:44.706 |

Observed intervals: 71.408 seconds from idle refresh to LTM budget; 58.234 seconds from LTM budget to prewarm completion; 131.536 seconds from turn receipt to the first subsequent upstream log.

These are coarse log intervals, not precise attribution to individual operations. A concurrent embedding rebuild (856 knowledge entries and 11,827 distillations pending after fingerprint invalidation) also confounds comparison with the earlier run. The missing stage measurements prevent determining whether semantic conversion, stored-ID resolution, or scheduling delays dominate.

## Reproduction

1. Execute `PreparationTiming.upstreamStart()` with populated counters and stage measurements using the real logger.
2. Capture the registered log sink and/or persistent log file.
3. Observe `semantic-preparation [object Object]` instead of the numeric payload. A large transcript is not required to reproduce the formatting defect.

## Expected fix and regression coverage

- Serialize this bounded numeric payload explicitly, or provide an equally readable structured representation, so the actual sink/file output preserves counts and stage wall/CPU timings.
- Prefer a narrowly scoped call-site fix unless a general logger change is deliberately justified; preserve existing redaction and Error formatting.
- Add a regression through the real formatting/sink path, asserting recoverable numeric counters and stage timings and absence of `[object Object]`. Mocking `log.info` alone would miss this defect.
- Preserve Sentry emission and best-effort instrumentation behavior.
- Do not introduce transcript text, credentials, paths, or message/session identifiers into the timing payload.

Related: #1714, #1712, #1710, #1708.

BYK:
Long resumed Codex histories repeatedly tokenize encrypted provenance before LTM, matching the phase where the supplied logs stall for 83 seconds on successive turns. The earlier fixture's repetitive ciphertext substantially understated this cost.

This change stores exact hidden-input token counts keyed by the complete serialized visible/provenance pair and a tokenizer version. Warm requests and database restarts reuse those counts while retaining the authoritative full message graph, original wire provenance, tool pairing, source IDs, and fresh timestamps.

The schema-v86 cache is local-only, tenant/project/session scoped, bounded, and disposable. No-store bypasses durable reads/writes. Corrupt entries recompute. Clear/delete/recreate, rollback, and merge preserve deletion semantics. Publication cannot wait on background writers or restore metadata after intervening database writes.

It also fixes the real log sink printing `semantic-preparation [object Object]`: numeric stage timings and cache diagnostics are explicitly serialized.

### Evidence

Committed benchmark: [report](https://github.com/BYK/loreai/blob/04290883fa8992fb6bd365ca43990ca88b894830/quality/benchmarks/semantic-token-cache-1710.md), [raw observations](https://github.com/BYK/loreai/blob/04290883fa8992fb6bd365ca43990ca88b894830/quality/benchmarks/semantic-token-cache-1710.json).

Final synthetic 5,580-message high-entropy fixture:

- Without durable reuse: **102.4 seconds**, including 101.7 seconds of BPE.
- First population: **110.8 seconds**.
- Same request warm: **150 ms**, zero BPE misses.
- Database close/reopen: **148 ms**, zero BPE misses.
- Append two source messages: **196 ms**, exactly one new provenance count.
- Equivalent requests produce identical serialized wire hashes; retained payload is about 207 KB.

These measure semantic preparation, excluding HTTP parsing, LTM, gradient processing, and inference. First use, eviction, invalidation, or no-store can still pay the full tokenization cost.

### Validation

Regression tests were proven failing before the logging/reuse fixes and the review-found lifecycle/contention repairs. Focused tests cover restart, historical edits/rewinds, byte-complete keys, timestamps, no-store, tenant isolation, corruption, bounds, migration/recovery, deletion/recreation from both connections, and locked writers.

Independent correctness and security reviews are complete with no unresolved findings. All ten targeted mutants are killed. Typecheck, lint, formatting and bundle pass; all ten sync-property runs pass. Security focused validation passed 129 tests; the final test-only addition passes all 28 core cache tests.

The full suite on identical production source finished with **9,554 passed, 7 failed, 229 skipped** (exit 1). All seven match documented base/environment failures: raw-window e2e (1), compact-endpoint budget assertions (2), installer process/ps checks (3), optional GPU assets (1). There are no newly introduced failures. The final delta adds one regression parameter; lint/format and its focused tests were revalidated.

The literal pnpm-test pretest hits the known tsx IPC restriction here. Validation used the equivalent explicit Node/tsx bundle command and complete Vitest suite. Seer passed on final head `04290883fa8992fb6bd365ca43990ca88b894830` with no issues (reference 16546882). Hosted semantic lint, documentation checks and Postgres sync integration also pass. The hosted test job and its downstream build/smoke gates remain pending; no failed gate is bypassed.

Fixes #1715.
Related to #1710, which remains open: this removes the measured BPE bottleneck but does not yet implement the source frontier or bounded active model window. #1716 recall-budget recovery remains a separate follow-up.

### DIFF
--- a/packages/core/src/data.ts
+++ b/packages/core/src/data.ts
@@ -654,6 +654,10 @@ export function clearProject(projectPath: string): ClearResult {
   // Delete in dependency order
   database.exec("BEGIN IMMEDIATE");
   try {
+    // Preparation may publish derived counts before any temporal row exists.
+    database
+      .query("DELETE FROM semantic_token_cache WHERE project_id = ?")
+      .run(pid);
     database
       .query("DELETE FROM session_prompt_deltas WHERE project_id = ?")
       .run(pid);
--- a/packages/core/src/db.ts
+++ b/packages/core/src/db.ts
@@ -2058,6 +2058,18 @@ export const MIGRATIONS: readonly string[] = Object.freeze([
   CREATE INDEX IF NOT EXISTS idx_temporal_embedding_queue_enqueued
     ON temporal_embedding_queue(enqueued_at, message_id);
   `,
+  `
+  -- Version 86: bounded local derived token counts; no transcript or vectors.
+  CREATE TABLE IF NOT EXISTS semantic_token_cache (
+    project_id TEXT NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
+    session_id TEXT NOT NULL REFERENCES session_state(session_id) ON DELETE CASCADE,
+    payload TEXT NOT NULL CHECK(length(payload) <= 1000000),
+    updated_at INTEGER NOT NULL,
+    PRIMARY KEY (project_id, session_id)
+  );
+  CREATE INDEX IF NOT EXISTS idx_semantic_token_cache_updated
+    ON semantic_token_cache(updated_at);
+  `,
 ]);
 
 // Index of the migration whose work is performed by a column-presence-aware JS
@@ -2927,6 +2939,10 @@ export function dbPath(): string {
 }
 
 let instance: Database | undefined;
+/** Read-only identity check for deferred derived-cache writes; never opens a DB. */
+export function isCurrentDatabase(connection: Database): boolean {
+  return instance === connection;
+}
 /** Actual file backing `instance`; absent for private in-memory databases. */
 let instanceFilePath: string | undefined;
 
@@ -4113,6 +4129,17 @@ function recoverMissingObjects(database: Database) {
     CREATE INDEX IF NOT EXISTS idx_temporal_embedding_queue_enqueued
       ON temporal_embedding_queue(enqueued_at, message_id);
   `);
+  database.exec(`
+    CREATE TABLE IF NOT EXISTS semantic_token_cache (
+      project_id TEXT NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
+      session_id TEXT NOT NULL REFERENCES session_state(session_id) ON DELETE CASCADE,
+      payload TEXT NOT NULL CHECK(length(payload) <= 1000000),
+      updated_at INTEGER NOT NULL,
+      PRIMARY KEY (project_id, session_id)
+    );
+    CREATE INDEX IF NOT EXISTS idx_semantic_token_cache_updated
+      ON semantic_token_cache(updated_at);
+  `);
   // Version 54: knowledge_session_injections.verdict (outcome impact, #497).
   // The verdict-keyed index MUST be created here, AFTER the column is ensured —
   // never in the big exec above, which runs before this ALTER and would throw
@@ -4234,6 +4261,7 @@ export const PROJECT_MERGE_TABLES = Object.freeze([
   "project_id_aliases",
   "project_path_aliases",
   "session_prompt_deltas",
+  "semantic_token_cache", // Disposable; source-project deletion cascades it away.
   "session_rollup",
   "temporal_messages",
   "temporal_vec",
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -343,6 +343,7 @@ export {
 } from "./vec-latency";
 export { distillLimiter, curatorLimiter } from "./session-limiter";
 export { estimateTokens, encodingForModel } from "./tokenize";
+export { SemanticTokenCache } from "./semantic-token-cache";
 export {
   installFetchInterceptor,
   shouldIntercept,
--- a/packages/core/src/semantic-token-cache.ts
+++ b/packages/core/src/semantic-token-cache.ts
@@ -1,0 +1,178 @@
+import { createHash } from "node:crypto";
+import { db, isCurrentDatabase, projectId, withSavepoint } from "./db";
+import { currentTenantId } from "./tenant";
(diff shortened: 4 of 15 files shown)


BYK (commented):
Independent security review — no unresolved findings on final head `04290883fa8992fb6bd365ca43990ca88b894830` (tree `c43708ddf7d7cb3f27ac4accc30ace6b7a595acf`). Conducted by a separate review agent; this comment records its evidence.

- PASS — pre-response project clear removes derived hashes; the initially reproduced retention defect is fixed.
- PASS — atomic local-write/external-commit guards reject publication after delete/recreate from either connection. Connection and tenant checks prevent delayed writes into a replacement owner.
- PASS — tenant-resolved project/session lookup, bound SQL values, complete framed input hashes, version/checksum/shape validation, bounded payload/entries/session rows.
- PASS — no-store and amnesia bypass durable cache I/O; table is excluded from every sync tier. Only hashes and counts persist.
- PASS — zero writer busy timeout is restored; rejected publication does not evict other cache rows.
- PASS — numeric diagnostics contain no source text, ciphertext, credentials, paths or session IDs.
- PASS — upgrade/recovery, rollback, clear/delete/recreate, and project merge regressions.

Independent runs: 129 focused tests passed, one benchmark skipped; all ten sync-property runs passed (five tests each). Final test-only visible-tail addition was separately reviewed and its seven field-difference cases passed. Reviewed and published tree hashes match; working trees were clean and no production source was changed by reviewers.

Security verdict: MERGE subject to the separate correctness/full-suite and hosted gates; this is not a claim that outstanding CI has passed.

BYK (commented):
Independent adversarial/correctness review complete on final head `04290883fa8992fb6bd365ca43990ca88b894830`, tree `c43708ddf7d7cb3f27ac4accc30ace6b7a595acf`. A separate reviewer ran the suite and mutants in isolated worktrees; this comment records its evidence.

PASS — exact BPE arithmetic and zero counts, full visible/provenance keys, algorithm versioning; envelope integrity and resource bounds; no-store/tenant scope; migration/recovery and project merge; clear/delete/recreate lifecycle; atomic revision checks; nonblocking writer publication and timeout restoration; authoritative source/provenance/tool pairing/IDs/fresh timestamps; numeric real-sink telemetry.

All initial findings are fixed with regressions: pre-first-response clear retained hashes; deletion/recreation permitted stale cache publication; disposable writes could wait on a competing writer; visible-key truncation escaped the original tests.

All ten deliberate mutants were killed at the intended assertions: omitted provenance, truncated visible key, removed checksum, removed version check, bypassed no-store, removed local-write guard, removed external-commit guard, blocking cache publication, removed project-clear cleanup, and missing log serialization. The final visible-tail case was independently RED under the mutant, then GREEN after restoring production source; all 28 cache tests passed.

Full suite on identical production source: **9,554 passed, 7 failed, 229 skipped**, 457 files, 624.99 seconds, exit 1. The final delta adds one deterministic test parameter only. All seven failures match existing base/environment evidence: raw-window e2e (1), compact-endpoint budget assertions (2), installer process/ps checks (3), and optional GPU vendor assets (1). No newly introduced failure was found.

Typecheck/lint/format each exit 0; final lint/format repeated after the test-only delta. Bundle passes. The literal pnpm-test pretest encountered the documented tsx IPC EPERM; validation used `node --import tsx packages/gateway/script/bundle.ts` followed by the full Vitest run with maxWorkers=2. The security pass separately supplied ten successful sync-property runs.

Both reviewer worktrees and the parent checkout are clean. Final tracked-content SHA-256: `3f92169883996bc3e1536660aaa3d64dea3212ae17d83616d03e315e4e44db29`.

CONCERN/scope: this removes measured repeated BPE while preserving full source processing. Keep #1710 open for the complete frontier/window goal.

Verdict: **MERGE subject to final hosted checks. No unresolved correctness findings.** Outstanding hosted checks are not bypassed or claimed green.

PR #1727, opened 2026-09-08
fix(core): move startup vector orphan discovery to idle read workers

### LINKED ISSUE #1681: perf(core): move vec0 orphan discovery off startup and the event loop
## Problem

Embedding startup always runs full `gcVec0DanglingRows(db())` in vec0 mode, even after migration and rechunk state have converged. Temporal orphan discovery uses a whole-table anti-join on the writer connection and blocks startup.

## Evidence

- An equivalent temporal dangling-row anti-join took 2.87 seconds: 2.24 seconds user CPU and 0.42 seconds system CPU.
- Durable temporal migration state was already `done=1` with no remaining rows, so the scan was unrelated to useful backfill work.

## Required fix

Discover orphans in bounded pages on a read worker during idle maintenance. Return exact chunk IDs and delete them on the writer connection. Never rerun a timed-out scan synchronously.

## Acceptance criteria

- Local embedding prewarm and startup never await a full vec0 scan.
- Orphan discovery is bounded and off-thread.
- Cleanup deletes exact chunk IDs only.
- Timeout or worker failure degrades safely and never blocks startup.
- Backed rows survive while dangling rows are eventually removed.

BYK:
Startup still scanned every vec0 row on the SQLite writer after embedding backfill, even when migration and re-chunking had already converged. Large databases could block the gateway's event loop during that scan.

This moves startup orphan discovery to idle read-worker maintenance. Each page examines at most 128 vector keys using an indexed cursor. The writer deletes exact keys only after rechecking their current backing rows, skips lock contention, and ignores late results after shutdown or database replacement. Maintenance pauses for active sessions and retries worker failures without a synchronous fallback. Existing explicit bulk-delete behavior is unchanged.

Fixes #1681

## Validation

- 129 focused tests pass, including 20 cleanup regressions, the native bundled read-worker path, indexed query plans, restore/replacement races, shutdown fencing, worker failure/backoff, actual SQLite contention and gateway idle ownership.
- Startup and busy-timeout regressions were proven failing before their fixes; independent mutation checks verify liveness, rollback, ownership, foreground priority, paging and knowledge-current guards.
- Typecheck, lint, formatting and gateway bundle pass. Both reviewers independently passed ten property-test runs.
- Two independent full-suite runs: 9,969 passed, 7 documented base/environment failures, 234 skipped. Failures are installer process checks (3), compact endpoint behavior (2), cache raw-window boundary (1), and optional Linux GPU assets (1). A separate stale-bundle setup failure was resolved with an isolated rebuild; all 7 bundle-export retests pass. Final test-only additions were rerun separately.
- Independent correctness and security reviews: MERGE subject to hosted gates; no unresolved findings. Exact reviewed tree: `cb36632e1f1bf54955f5cd3c60efd8017a238b15`.
- Seer passed on final head `1f5077215079fc32d48a08ac6c5e820d18d20e5d` with no issues. Hosted test completion remains pending.

### DIFF
--- a/packages/core/src/db/vec-store.ts
+++ b/packages/core/src/db/vec-store.ts
@@ -690,8 +690,9 @@ const ALL_EMBEDDING_TABLES: readonly EmbeddingTable[] = [
  * rows. These rows are already HARMLESS for correctness (recall hydration drops
  * a hit whose base row is missing, and a deleted project's rows live in their
  * own partition), so this is a bloat / recall-quality backstop, not a fix. Runs
- * at startup in vec0 mode (all tables) and after a bulk base-row delete (scoped
- * to the tables that delete touched — see `data.ts`). One bounded anti-join pass
+ * only after an explicit bulk base-row delete (scoped to the tables that delete
+ * touched — see `data.ts`). Startup uses idle read-worker maintenance instead.
+ * One full anti-join pass
  * per requested table; each is a single scan (cheaper than a per-id delete on
  * the un-indexed `temporal_vec.message_id` aux column for large id sets).
  */
--- a/packages/core/src/embedding.ts
+++ b/packages/core/src/embedding.ts
@@ -33,7 +33,6 @@ import {
   type EmbeddingTable,
   embeddingColumnExists,
   ensureVec0Store,
-  gcVec0DanglingRows,
   hasEmbeddingSql,
   missingEmbeddingSql,
   readStorageMode,
@@ -3458,9 +3457,8 @@ export async function runStartupBackfill(
     };
   }
 
-  // Startup backstop: reclaim vec0 rows orphaned by bulk base-row deletes
-  // (project/session/prune) since the last run. Harmless if there are none.
-  if (mode === "vec0") gcVec0DanglingRows(db());
+  // Orphan discovery belongs to the host's idle read-worker maintenance.
+  // Never scan the full vector corpus on the startup writer (#1681).
 
   // Coverage stats — always log to stderr so the problem is visible.
   const kTotal = (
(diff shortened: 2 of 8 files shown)


BYK (commented):
Independent correctness and security reviews are complete for final head `1f5077215079fc32d48a08ac6c5e820d18d20e5d`, tree `cb36632e1f1bf54955f5cd3c60efd8017a238b15`.

Both verdicts: **MERGE subject to required hosted gates**. No blocking source or security findings remain. The three initial lint findings were redundant assertions and are fixed.

- Correctness: bounded indexed paging, exact-key writer deletion, restored/replaced sources, rollback/retry, busy-writer handling, shutdown/database ownership and foreground priority pass.
- Security: fixed SQL identifiers and bound IDs, cross-tenant live-vector preservation, knowledge-current/tombstone behavior, hostile/large IDs and privacy-safe failure handling pass.
- Mutation checks catch omitted liveness, rollback, stop fencing, foreground gating, page limits, knowledge-current selection and zero busy-timeout guards. Every mutant was reverted; both review worktrees are clean.
- Final focused verification: **129 passed**, including **20 cleanup regressions** and the native bundled read worker. Typecheck, lint, format and bundle pass. Each reviewer independently passed ten consecutive property runs.
- Each full-suite run finished with **9,969 passed, 7 known base/environment failures, 234 skipped** on runtime-equivalent source. Those failures are installer process checks (3), compact endpoint behavior (2), cache raw-window boundary (1), and optional Linux GPU assets (1). A separate stale-bundle setup failure was resolved by an isolated build using the supported Node/tsx invocation; **all 7 bundle-export retests pass**. Final test-only additions were inspected and rerun separately.

Seer also passed on this exact head with no issues (reference 16567953). Local full-suite failures are documented, not treated as permission to bypass hosted gates.

PR #1756, opened 2026-09-10
fix(gateway): make foreground SSE inactivity + request deadlines configurable

rgutzen:
## Problem

The foreground relay aborted a stream whenever the upstream emitted nothing for **120 s**:

```ts
const FOREGROUND_SSE_INACTIVITY_MS = 120_000;  // hardcoded, no env override
```

All four foreground parse sites read it (Anthropic relay, Anthropic recall continuation, Responses relay, Responses recall continuation). Opus-class extended thinking over large cached prompts can go silent for minutes, so the gateway killed the relay while the client's own watchdogs (180 s / 300 s) never fired. The client saw a truncated `200` with zero content and reported a transport loss rather than a stall.

Two ceilings existed with **no declared relationship**: the 120 s inactivity deadline and a 300 s foreground request timeout. A naive bump of the inactivity deadline would have been a no-op above 300 s — the coarser timer fires first. The worker path had the identical pair (120 s / 300 s).

## Fix

Add `packages/gateway/src/sse-inactivity.ts` centralizing the deadlines:

| Env var | Default |
|---|---|
| `LORE_FOREGROUND_SSE_INACTIVITY_MS` | 600 s (was hardcoded 120 s) |
| `LORE_FOREGROUND_REQUEST_TIMEOUT_MS` | 900 s (was hardcoded 300 s) |
| `LORE_WORKER_RESPONSE_INACTIVITY_MS` | 600 s (was hardcoded 120 s) |
| `LORE_WORKER_REQUEST_TIMEOUT_MS` | 900 s (was hardcoded 300 s) |

Each request ceiling is clamped to `inactivity + 60 s headroom`, so raising the inactivity deadline can never leave a tighter request ceiling silently in force — the trap that would have made a 120→600 s change inert.

600 s sits above every client watchdog the gateway serves, so the client stays the authority on a stalled stream and the gateway intervenes only on a genuinely dead connection.

## Tests

Five suites advanced a literal `300_000` to trip a deadline; each now derives from the exported constant, so a test can no longer encode a stale default. Two tests encoded the old value in their scenario shape rather than just their timer advance (a 600 s throttle that only lost the race while the deadline was 300 s; a 299 s attempt that only abutted a 300 s deadline) and are now expressed relative to the deadline.

New `sse-inactivity.test.ts` covers parsing, the defaults, and the invariant `request ceiling > inactivity`.

## Note

This is a latent-defect fix, independent of the parallel-recall abort fixed in #1755. It does not address that issue and is not required by it.

### DIFF
--- a/packages/gateway/src/llm-adapter.ts
+++ b/packages/gateway/src/llm-adapter.ts
@@ -69,7 +69,10 @@ import {
   SSEStreamLimitError,
   SSEStreamTransportError,
 } from "./stream/anthropic";
-import { WORKER_RESPONSE_INACTIVITY_MS } from "./sse-inactivity";
+import {
+  WORKER_REQUEST_TIMEOUT_MS,
+  WORKER_RESPONSE_INACTIVITY_MS,
+} from "./sse-inactivity";
 import { isBedrockMantleHost, toMantleModelId } from "./translate/bedrock";
 import {
   ANTHROPIC_CONTENT_BLOCK_TYPES,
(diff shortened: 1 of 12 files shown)


BYK:
Superseded by #1876. The replacement is rebased onto current `main`, lives in `BYK/loreai`, and includes the timer-boundary fix and tests. Closing this PR so review and CI continue on #1876.

